PR #33, opened 2024-07-21
Fix a doc for CubeMode

hatoo:
The doc for `#[cude]` proc macro seems wrong.

According to https://github.com/tracel-ai/cubecl/blob/89af40f86bc634e228d3c9fe044ebf3f680c5e08/crates/cubecl-macros/src/lib.rs#L87 and some experiment on my side, `#[cube(debug)]` should be correct.


nathanielsimard (approved):

nathanielsimard:
You're right, merged, thanks for the PR

PR #35, opened 2024-07-23
Upgrade wgpu to 22.0.0

ArthurBrussee:
**Changes**

Upgrade wgpu to 22.0.0. There's two breaking things:

Have to specifiy a MemoryHint, choosing between performance and lower memory usage. GIven wgpu has it's own block allocator now, MemoryUsage seems better. I did a quick test on the train sample and see no clear performance difference between either.

Need to specify a pipeline cache. This allows to cache compiled shader outputs. As per the comment in the wgpu code this is usually done automatically though not on Android seemingly. Could be nice someday.


nathanielsimard (approved):
Thanks

PR #43, opened 2024-07-29
Allow vectorized_empty on Numeric trait

louisfd:


crates/cubecl-core/src/frontend/element/numeric.rs
```diff
@@ -99,6 +103,19 @@ pub trait Numeric:
...
         new_var.into()
     }
+
+    fn __expand_vectorized_empty(
+        context: &mut CubeContext,
+        vectorization: UInt,
+    ) -> <Self as CubeType>::ExpandType {
+        if vectorization.val == 1 {
+            Self::__expand_from_int(context, ExpandElementTyped::from_lit(0))
+        } else {
```
nathanielsimard, line 113: Is it necessary?

PR #52, opened 2024-08-08
Made compatible with no_std and no atomic ptr

BjornTheProgrammer:
This is in preparation for Burn's pull request [#2096](https://github.com/tracel-ai/burn/pull/2096).

Some tests should be added to ensure that this doesn't end up breaking in the future, I can do so, but I'm unfamiliar with the testing story here. It would just need to be built with the `thumbv6m-none-eabi` target

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -68,5 +68,7 @@ derive_more = { version = "0.99.18", features = ["display"], default-features = 
 env_logger = "0.11.3"
 strum = {version = "0.26.3", features = ["derive"]}
 
+portable-atomic-util = { version = "0.2.2", features = ["alloc"] } # alloc is for no_std
+
 [profile.dev]
 opt-level = 2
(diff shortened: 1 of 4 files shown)


antimora:
@BjornTheProgrammer FYI, it's sufficient to fix make this work with crates in this test: https://github.com/tracel-ai/burn/tree/main/crates/burn-no-std-tests

cubecl doesn't need to be no-std compatible.

BjornTheProgrammer:
You're totally right, for whatever reason, during my testing on my local machine it broke without my version. I think I've found the issue though. Sorry for the inconvenience

antimora:
@nathanielsimard can you please look into this? For some reason burn-common has dependency on cube.

nathanielsimard (commented):
`cubecl-common` should actually be `no-std` when `std` isn't enabled. The current fix seems fine to me.

BjornTheProgrammer:
> `cubecl-common` should actually be `no-std` when `std` isn't enabled. The current fix seems fine to me.

Is it fine for this to be merged then?

PR #53, opened 2024-08-08
fix wgpu compiler float precision

wcshds:
Fixes Burn issue [#2118](https://github.com/tracel-ai/burn/issues/2118)


nathanielsimard (approved):
It's strange that the "display" of Rust doesn't handle that part, but hey what can we do about it (your fix) 😅

PR #54, opened 2024-08-09
fix wgpu compiler visibility

wcshds:
```rust
impl Display for Visibility {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        match self {
            Visibility::Read => f.write_str("read_write"),
            Visibility::ReadWrite => f.write_str("read_write"),
        }
    }
}
```
should be
```rust
impl Display for Visibility {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        match self {
            Visibility::Read => f.write_str("read"),
            Visibility::ReadWrite => f.write_str("read_write"),
        }
    }
}
```

### DIFF
--- a/crates/cubecl-wgpu/src/compiler/wgsl/shader.rs
+++ b/crates/cubecl-wgpu/src/compiler/wgsl/shader.rs
@@ -228,7 +228,7 @@ impl Display for Location {
 impl Display for Visibility {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
-            Visibility::Read => f.write_str("read_write"),
+            Visibility::Read => f.write_str("read"),
             Visibility::ReadWrite => f.write_str("read_write"),
         }
     }


wcshds:
but burn-wgpu test failed...

image

wcshds:
> This is unfortunately quite hard to fix :/ See the discussion previously on the burn repo: [tracel-ai/burn#1996](https://github.com/tracel-ai/burn/issues/1996)
> 
> It will require some new approaches to memory management that prevents these conflicts.
> 
> I do think this might be important as it avoids some barriers wgpu otherwise inserts, and issues with `workgroupBarrier `.

Thanks for your reply! So, is the current code intentional? Sorry, I don't know much about wgpu and cubecl... I just encountered the following issue when using burn in wasm, so I want to know if it's possible to change the `info` variable to read-only.

image

nathanielsimard:
@wcshds yes, the current code is intentional. We still pass the visibility to the compiler, but for now, given the restrictions of WGPU, we can't heavily reuse buffers with different visibility, even if it's for different slices of the buffer.

wcshds:
The issue I encountered in wasm can be resovled by following code:
```rust
fn format_binding(
    f: &mut core::fmt::Formatter<'_>,
    name: &str,
    binding: &Binding,
    num_entry: usize,
) -> core::fmt::Result {
    let ty = match binding.size {
        Some(size) => format!("array<{}, {}>", binding.item, size),
        None => format!("array<{}>", binding.item),
    };

    #[cfg(all(not(target_family = "wasm"), feature = "std"))]
    let visibility = binding.visibility.to_string();
    #[cfg(any(target_family = "wasm", not(feature = "std")))]
    let visibility = if name == "info" {
        "read".to_string()
    } else {
        binding.visibility.to_string()
    };

    f.write_fmt(format_args!(
        "@group(0)
@binding({})
var<{}, {}> {}: {};
\n",
        num_entry, binding.location, visibility, name, ty
    ))?;

    Ok(())
}
```
The workaround is really hacky though.

nathanielsimard:
We have token another direction, we fixed the problem on wasm by using the simpler memory management.

PR #57, opened 2024-08-09
Fix/layout/broadcasted

laggui:
Bug encountered with Burn [PR #2137](https://github.com/tracel-ai/burn/pull/2137).

Some expand tests were failing because the contiguous check was incorrect. Tracked it down to `matrix_layout` not handling broadcasted dims properly.

Lmk if I mixed up some `HighlyPermuted` and `MildlyPermuted` cases when handling broadcasted dims.


nathanielsimard (approved):
LGTM

PR #62, opened 2024-08-12
Add support for atomics and atomic operations

wingertge:
This adds basic atomics and atomic operations, as well as the smallest common denominator functionality between CUDA and wgpu. 

### CUDA
CUDA atomics get compiled to regular types since cuda has untyped atomics, and `load/store` get compiled as regular pointer dereference operations since CUDA currently doesn't support `load/store` semantics (it is supported in the intermediate bytecode so it may support it at some point). All other functions compile to the cuda equivalent function calls.

### WGPU
Atomic operations are mapped 1:1 with the equivalent `wgpu` builtins. While `wgpu`'s `atomic<T>` is generic in theory,  it only supports `i32` and `u32` as generic parameters so it is not really generic in practice. I had to add a lot of special exceptions to the compiler when working with atomic values, because they can't be constructed, cast or assigned to variables. All local bindings for atomics are unique `let` bindings of a pointer to the atomic value.

### Implementation considerations
1. I chose to implement separate functions rather than regular `AddAssign/SubAssign` operations because atomic operations function on very different semantics (i.e. they all return the old value). 
2. Because CUDA only returns the old value for `atomicCAS`, `compare_and_swap` returns the old value on its own. While `wgpu` returns both a `bool` of whether the operation was successful and the old value, which is more efficient when used in a spinlock, to provide a unified API with CUDA `compare_and_swap` has to return only the old value.
3. Atomics are implemented as separate types because `wgpu` only supports integer atomics and doesn't support any regular operations on atomic types. CUDA doesn't care, but again, I'm trying to create a common API that works for both and doesn't allow invalid behaviour on either.

### Limitations
As a result of not implementing `Numeric` (because they don't support most numeric operations), atomics cannot currently be used as scalar launch parameters. I would like to change this in the future and the scaffolding for it is there, but it needs more work and would be a non-breaking change anyways. They do work when they're wrapped in an array-like structure.

### Testing
* All `wgpu` tests still run successfully - I can't test CUDA locally because I can't figure out which exact version it's looking for.
* While no new tests were written as of yet (I'm not familiar enough with the testing architecture to figure out how to test the codegen), I did test the atomic `add` and `compare_and_swap` operations in the context of a deformable convolution shader and the results are identical to `pytorch` (so they seem to work as expected).
* I'd be happy to add tests if I can get a better understanding of how to do it.

### DIFF
--- a/crates/cubecl-core/src/codegen/execution.rs
+++ b/crates/cubecl-core/src/codegen/execution.rs
@@ -326,6 +326,7 @@ struct ExecuteSettings<R: Runtime> {
     cube_count: CubeCount<R::Server>,
 }
 
+#[allow(clippy::too_many_arguments)]
 fn execute_settings<
     'a,
     R: Runtime,
--- a/crates/cubecl-core/src/codegen/execution.rs
+++ b/crates/cubecl-core/src/codegen/execution.rs
@@ -74,9 +74,11 @@ where
     /// Execute a dynamic kernel.
     #[allow(unused)]
     pub fn execute(self, launch: CubeCountSettings<R::Server>) {
-        execute_dynamic::<R, K, f32, f32, f32>(
+        execute_dynamic::<R, K, f32, f32, f32, f32, f32>(
             self.inputs,
             self.outputs,
+            None,
+            None,
             None,
             None,
             None,
@@ -109,10 +111,12 @@ where
     /// Execute a dynamic kernel.
     #[allow(unused)]
     pub fn execute(self, launch: CubeCountSettings<R::Server>) {
-        execute_dynamic::<R, K, E, f32, f32>(
+        execute_dynamic::<R, K, E, f32, f32, f32, f32>(
             self.inputs,
             self.outputs,
             Some(self.scalars.0),
+            None,
+            None,
             None,
             None,
             self.kernel,
@@ -149,11 +153,13 @@ where
         K: Kernel + 'static,
         R: Runtime,
     {
-        execute_dynamic::<R, K, E1, E2, f32>(
+        execute_dynamic::<R, K, E1, E2, f32, f32, f32>(
             self.inputs,
             self.outputs,
             Some(self.scalars.0),
             Some(self.scalars.1),
+            None,
+            None,
             None,
             self.kernel,
             launch,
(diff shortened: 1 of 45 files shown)


crates/cubecl-core/src/codegen/execution.rs
```diff
@@ -74,12 +74,14 @@ where
     /// Execute a dynamic kernel.
     #[allow(unused)]
     pub fn execute(self, launch: CubeCountSettings<R::Server>) {
-        execute_dynamic::<R, K, f32, f32, f32>(
+        execute_dynamic::<R, K, f32, f32, f32, f32, f32>(
```
nathanielsimard, line 77: This is the old way of launching kernels, we will probably delete it once everything is migrated to the newer CubeCL API. What made you change it?

crates/cubecl-core/src/frontend/element/atomic.rs
```diff
@@ -0,0 +1,375 @@
...
+    frontend::{CubeContext, CubePrimitive, CubeType, ExpandElement, UInt},
+    ir::{
+        BinaryOperator, CompareAndSwapOperator, Elem, IntKind, Item, Operator, UnaryOperator,
+        Vectorization,
+    },
+    prelude::KernelBuilder,
+    unexpanded,
+};
+
+pub trait Atomic: Sized + CubeType
```
nathanielsimard, line 15: I would add some documentation before merging.

crates/cubecl-core/src/frontend/element/cast.rs
```diff
@@ -29,3 +32,32 @@ impl<P: CubePrimitive> Cast for P {
         unexpanded!()
     }
 }
+
+pub trait BitCast: CubePrimitive {
```
nathanielsimard, line 36: Same, I would add some documentation, but yeah we needed that operation thanks!

crates/cubecl-core/src/frontend/element/cast.rs
```diff
@@ -29,3 +32,32 @@ impl<P: CubePrimitive> Cast for P {
...
+    #[allow(unused_variables)]
+    fn bitcast_from<From: CubePrimitive>(value: From) -> Self {
+        unexpanded!()
+    }
+
+    fn __expand_bitcast_from<From>(
+        context: &mut CubeContext,
+        value: From,
```
nathanielsimard, line 44: Theorically, it should be `From::ExpandType`.

crates/cubecl-core/src/frontend/element/cube_elem.rs
```diff
@@ -50,3 +50,21 @@ impl From<UInt> for ExpandElement {
...
+    fn from(_value: AtomicI64) -> Self {
+        todo!()
+    }
+}
+
+impl From<AtomicUInt> for ExpandElement {
+    fn from(_value: AtomicUInt) -> Self {
+        todo!()
+    }
+}
```
nathanielsimard, line 70: If not necessary I would remove those. Probably linked to the generic `From` parameter I guess.

crates/cubecl-core/src/ir/processing.rs
```diff
@@ -184,6 +184,37 @@ impl ScopeProcessing {
...
+                    sanitize_constant_scalar_ref_var(&mut op.rhs, &op.out);
+                }
+                Operator::AtomicAnd(op) => {
+                    sanitize_constant_scalar_ref_var(&mut op.rhs, &op.out);
+                }
+                Operator::AtomicOr(op) => {
+                    sanitize_constant_scalar_ref_var(&mut op.rhs, &op.out);
+                }
+                Operator::AtomicXor(op) => {
+                    sanitize_constant_scalar_ref_var(&mut op.rhs, &op.out);
```
nathanielsimard, line 216: Is there a reason not to sanitize lhs as well?

nathanielsimard (changes requested):

crates/cubecl-core/src/codegen/execution.rs
wingertge: I wasn't aware it was outdated - it's still used by at least a few `burn` kernels. I changed it because it includes a match on the element type, and I figured I should at least build the scaffolding to allow atomic scalar params. Tehy need to be treated separately because `wgpu` treats them  as entirely separate types. I can undo it if it's legacy code anways.

crates/cubecl-core/src/frontend/element/atomic.rs
wingertge: Good idea - I was changing things so much I forgot to go back and add docs. I'll go add it to all the new public types/traits.

crates/cubecl-core/src/frontend/element/cast.rs
wingertge: I kept it the same as the `Cast` type. It also uses the plain generic type.

crates/cubecl-core/src/ir/processing.rs
wingertge: As I understood it the santization is to ensure matching types, but `lhs` will explicitly not match (`AtomicUInt` vs `AtomicI32`). It can also be either of those. That being said, I can probably use the `Primitive` associated type to do the santization. I'll give it a try.

wingertge: Nevermind, I no longer have the type information at that point. So unless I match on every possible atomic variant I can't really sanitize that part.

wingertge:
Seems like CI is a bit broken and can't find a graphics adapter for the `wgpu` tests.

crates/cubecl-core/src/frontend/element/atomic.rs
```diff
@@ -12,28 +12,40 @@ use crate::{
...
     ExpandElement: From<<Self::Primitive as CubeType>::ExpandType>,
     ExpandElement: From<<Self as CubeType>::ExpandType>,
 {
+    /// The numeric primitive represented by the atomic wrapper.
     type Primitive: Numeric;
 
+    /// Load the value of the atomic. This operation is only atomic in `wgpu`, CUDA does not support
+    /// atomic loads.
     #[allow(unused_variables)]
     fn load(pointer: &Self) -> Self::Primitive {
```
nathanielsimard, line 27: What is the current fix for this? Can we use atomic add with 0 to simulate an atomic load?

nathanielsimard:
> Seems like CI is a bit broken and can't find a graphics adapter for the `wgpu` tests.

Yeah gonna work on this soon, as long as the ci is working fine locally we can merge for now :)

crates/cubecl-core/src/frontend/element/atomic.rs
wingertge: I think eventually we could emit inline PTX, or emit a thread fence if it's really necessary. But most of the time atomic loads aren't actually necessary and thread fences are a pretty significant performance degredation so I would go for just the note in the docs for the time being. We can still change the exact semantics later I think. The main reason the method exists is because `wgpu` *only* allows atomic loads for atomics variables and doesn't allow direct access.

wingertge: Now that I've thought about it more and looked at the code, I think inline PTX is actually the way to go. Using inline PTX sucks when you're writing normal code, but I think it actually works fine if we're just compiling to it. I'll try to implement it.

wingertge: I have once again been mislead by the internet. PTX does not in fact support atomic loads, just non-tearing loads/stores. I think that idea about `atomicAdd` is a good one, with `atomicExch` for store since they're identical, the only difference being that `atomicExch` returns the value. Apparently the CUDA compiler will just optimize that out anyways and replace it with the appropriate instruction.

nathanielsimard (approved):

PR #65, opened 2024-08-12
Revert float precision changes to wgsl compiler

laggui:
Revert the precision changes applied in #53.

While the initial PR was merged to fix Burn's [issue with the web image classification example](https://github.com/tracel-ai/burn/issues/2118), this has affected a bunch of wgpu tests.

An alternative has been proposed by using 13 decimal points, which seems to resolve the previously introduced errors in the wgpu tests, but it feels a bit arbitrary.

The original issue will remain open and requires further investigation.


laggui:
Damn it, we need to change the default PR settings to squash and merge @nathanielsimard 😅

PR #66, opened 2024-08-12
Support destructuring tuples in cube macro

cBournhonesque:
I'm trying to convert the reduce kernels to use the `#[cube]` macro: https://github.com/tracel-ai/burn/pull/2117

One problem I'm running into is that the accumulator for some of the kernels is a tuple, and tuple destructuring doesn't seem to be supported currently.

This PR adds tuple destructuring, with a unit test


nathanielsimard (approved):
LGTM 🔥

PR #67, opened 2024-08-12
support assigning to a mutable reference in cube macro

cBournhonesque:


crates/cubecl-macros/src/codegen_function/variable.rs
```diff
@@ -167,6 +167,18 @@ pub(crate) fn codegen_assign(
...
+            let rhs = codegen_expr(&assign.right, loop_level, variable_tracker);
+
+            quote::quote! {
+                {
+                    let _assign_lhs = #lhs;
+                    let _assign_rhs = #rhs;
+                    cubecl::frontend::assign::expand(context, _assign_rhs, _assign_lhs)
+                }
+            }
+        }
```
nathanielsimard, line 181: Could we use the | operator for this?

```rust
 syn::Expr::Unary(_) | syn::Expr::Field(_) | sync::Expr::Path(_) => {
      let lhs = codegen_expr(&assign.left, loop_level, variable_tracker);
      let rhs = codegen_expr(&assign.right, loop_level, variable_tracker);

      quote::quote! {
          {
              let _assign_lhs = #lhs;
              let _assign_rhs = #rhs;
              cubecl::frontend::assign::expand(context, _assign_rhs, _assign_lhs)
          }
      }
  }
```

nathanielsimard (commented):
Maybe we can reduce code duplication a bit?

crates/cubecl-macros/src/codegen_function/variable.rs
cBournhonesque: good call

cBournhonesque:
The format checks fail in ci but not locally

PR #69, opened 2024-08-13
Automatically enable autotune_persistent_cache only on desktop platforms

ArthurBrussee:
Currently, the persistent cache fails on platforms without easy access to a filesystem like android and the web. Configuring this as a negative feature becomes quite annoying.

Instead, as per [this discord discussion](https://discord.com/channels/1038839012602941528/1265322519569764403), this just makes it so it's only enabled if on a desktop platform. To save some gnarly cfgs everywhere this using cfg_aliases which might be useful more generally in cube/burn, but lmk if it's preferred to not have this.

There's also a leftover flag in cubecl-cuda for autotune I think. It seems Burn actually handles all the autotune flags, cube doesn't know about them.

### DIFF
--- a/crates/cubecl-macros/src/analyzer.rs
+++ b/crates/cubecl-macros/src/analyzer.rs
@@ -80,16 +80,22 @@ impl VariableAnalyzer {
                             for pat in pat_tuple.elems.iter() {
                                 let (id, is_comptime) = find_local_declaration_ident(pat);
                                 if let Some(id) = id {
-                                    self.variable_tracker
-                                        .analyze_declare(id.to_string(), depth, is_comptime);
+                                    self.variable_tracker.analyze_declare(
+                                        id.to_string(),
+                                        depth,
+                                        is_comptime,
+                                    );
                                 }
                             }
                         }
                         _ => {
                             let (id, is_comptime) = find_local_declaration_ident(&local.pat);
                             if let Some(id) = id {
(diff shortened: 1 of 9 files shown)


crates/cubecl-macros/src/analyzer.rs
```diff
@@ -80,16 +80,22 @@ impl VariableAnalyzer {
                             for pat in pat_tuple.elems.iter() {
                                 let (id, is_comptime) = find_local_declaration_ident(pat);
                                 if let Some(id) = id {
-                                    self.variable_tracker
-                                        .analyze_declare(id.to_string(), depth, is_comptime);
+                                    self.variable_tracker.analyze_declare(
```
ArthurBrussee, line 83: This is just to keep CI happy, seems some bad formatting was comitted.

crates/cubecl-runtime/build.rs
```diff
@@ -0,0 +1,8 @@
+use cfg_aliases::cfg_aliases;
+
+fn main() {
+    // Setup cfg aliases
+    cfg_aliases! {
+        autotune_persistent_cache: { all(feature = "std", any(target_os = "windows", target_os = "linux", target_os = "macos")) },
+    }
```
nathanielsimard, line 7: Ho I didn't know we could do that, awesome!

nathanielsimard (approved):
LGTM

PR #71, opened 2024-08-15
Stepped range

wingertge:
Adds a step variable to the for loop branch and a `range_stepped` function to generate a stepped range loop.


wingertge:
Needs more work, apparently the compiler was just not giving me the errors it was supposed to.

nathanielsimard (approved):

PR #72, opened 2024-08-17
Impl SyncStorage

AntiAnimeGeneral:
storageBarrier()  be used in conjunction with atomic operations to ensure visibility

### DIFF
--- a/crates/cubecl-wgpu/src/compiler/wgsl/compiler.rs
+++ b/crates/cubecl-wgpu/src/compiler/wgsl/compiler.rs
@@ -397,6 +397,9 @@ impl WgslCompiler {
         match synchronization {
             cube::Synchronization::SyncUnits => {
                 instructions.push(wgsl::Instruction::WorkgroupBarrier)
+            },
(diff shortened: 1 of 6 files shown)


crates/cubecl-wgpu/src/compiler/wgsl/compiler.rs
```diff
@@ -26,15 +26,21 @@ pub struct WgslCompiler {
 }
 
 impl core::fmt::Debug for WgslCompiler {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+    fn fmt(
+        &self,
+        f: &mut std::fmt::Formatter<'_>,
+    ) -> std::fmt::Result {
```
nathanielsimard, line 32: Is there a reason why the fommatting is different? What Rust version / rust-fmt are you using?

nathanielsimard (commented):
Thanks for the addition, only a comment about the formatting, but maybe it's due to a rustfmt update.

PR #74, opened 2024-08-17
Allow graceful failover when one tunable panics

wingertge:
Catches panics that occur during the benchmarking phase of autotuning and gracefully falls back to the next algorithm in line.

This allows selecting algorithms based on things like memory constraints (without adding `Result` types all through the memory allocation chain) or even things like potentially unsupported GPU features. Note that this only catches panics during benchmarking, not during regular execution. If an algorithm fails at that point it will still give error feedback to the user.
If all tunables panic, propagates the first panic.

One thing worth mentioning is that while it will gracefully continue to the next algorithm, it will not necessarily continue quietly. The error message still gets printed to the console, which could confuse users. Unfortunately I don't think that can be prevented due to how panic is implemented, the print gets output before unwind gets to the benchmark code where the unwind is caught. This can be prevented with a custom panic hook, but that would affect *all* panics everywhere, which is a bit too intrusive.

### Testing
All existing tuners work as expected with all tests passing, and the fallback works for `im2col` convolution when called with parameters that cause an out of memory error.

### DIFF
--- a/crates/cubecl-runtime/src/tune/tuner.rs
+++ b/crates/cubecl-runtime/src/tune/tuner.rs
@@ -122,7 +122,10 @@ impl<K: AutotuneKey> Tuner<K> {
             catch_unwind(AssertUnwindSafe(|| {
                 TuneBenchmark::new(operation, client.clone()).run()
             }))
-            .map_err(ManuallyDrop::new)
+            .map_err(|e| {
+                println!("Caught error while benchmarking, falling back to next operation.");
+                ManuallyDrop::new(e)
+            })
         }
         #[cfg(not(feature = "std"))]
         Ok(TuneBenchmark::new(operation, client.clone()).run())
--- a/crates/cubecl-runtime/src/tune/tuner.rs
+++ b/crates/cubecl-runtime/src/tune/tuner.rs
@@ -3,6 +3,8 @@ use web_time::Duration;
 
 #[cfg(not(target_family = "wasm"))]
 use core::time::Duration;
+use core::{any::Any, mem::ManuallyDrop, panic::AssertUnwindSafe};
+use std::panic::{catch_unwind, resume_unwind};
 
 use alloc::boxed::Box;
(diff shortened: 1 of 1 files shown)


crates/cubecl-runtime/src/tune/tuner.rs
```diff
@@ -16,6 +19,10 @@ use crate::tune::{AutotuneOperation, AutotuneOperationSet, TuneBenchmark, TuneCa
 
 use super::AutotuneKey;
 
+/// An error that occured during benchmarking. If other benches succeeded, ignore this bench and
+/// continue gracefully. If all benches fail, panic.
+type BenchError = ManuallyDrop<Box<dyn Any + Send>>;
```
nathanielsimard, line 24: Is there a reason to use that type for the error? Can we create a custom error type that we map instead?

nathanielsimard (commented):
Thanks for the contribution, and yes I think it makes a ton of sense to have this into autotune! I only have one comment.

crates/cubecl-runtime/src/tune/tuner.rs
wingertge: Unfortunately not. The way panics work in rust means that the error type returned by `catch_unwind` is completely opaque and afaik doesn't even contain the panic message, instead containing the unwind logic and stack. That's why it has to be `ManuallyDrop`; dropping it can actually trigger the unwind and cause all sorts of havoc in the process. That's why instead of providing a custom error message, we just have to pass it to `resume_unwind`, it's the only function that knows how to do anything with this opaque blob.

nathanielsimard: I would actually add that explaination in the code as a comment, that's very clear thanks!

wingertge: Done 👍

nathanielsimard (approved):

PR #80, opened 2024-08-22
Reimplement macro to make writing kernels more ergonomic

wingertge:
This is a rewrite of the frontend macro to make writing kernels more ergonomic and "rust-like".

It replaces the `Comptime` struct with a `#[comptime]` attribute so values can be used like any other Rust constant.

Loops now use normal ranges and allow unrolling via `#[unroll]` rather than a verborse comptime parameter that is often set to a literal anyways. For conditional unrolling, the attribute supports inserting an expression: `#[unroll(should_unroll)]`.

The custom vectorized types are replaced with Rust primitives since vectorization has always been stored in the IR anyways. The vectorization field on the custom types was almost entirely redundant.

Adds support for struct destructuring in locals.

Folding of comptime values into literals has also been improved.

## Testing
All unit and integration tests (including frontend tests) are passing. Several backend codegen tests are also added.
All burn tests are passing.


wingertge:
All tests passing including frontend tests, actually ready for merge now

wingertge:
Something in burn seems to be miscompiling, going back to draft until I can figure out what's going on

crates/cubecl-core/src/runtime_tests/assign.rs
```diff
@@ -20,7 +20,7 @@ pub fn test_kernel_assign_scalar<R: Runtime>(client: ComputeClient<R::Server, R:
         CubeCount::Static(1, 1, 1),
         CubeDim::default(),
         unsafe { ArrayArg::from_raw_parts(&handle, 2, vectorization) },
-        UInt::new(vectorization as u32),
+        //UInt::new(vectorization as u32),
```
nathanielsimard, line 23: dead code

crates/cubecl-linalg/src/matmul/cmma/block_loop.rs
```diff
@@ -9,31 +9,142 @@ use super::{
...
+//     for block in 0u32..n_loops {
+//         offsets.k = block * block_size_k;
+
+//         load_to_shared_memories::<F, FC>(lhs, rhs, offsets, shared_memories, dims, config);
+
+//         sync_units();
+
+//         compute_loop::<F, FC>(shared_memories, accumulators, config);
+
+//         sync_units();
```
nathanielsimard, line 35: I guess it should be de-commented now, and the generated expand code bellow should be removed.

crates/cubecl-linalg/src/matmul/tiling2d/tile/loader.rs
```diff
@@ -186,11 +201,12 @@ pub(crate) fn load_transposed<F: Float, L: BlockLoader<F>>(
     load_info: LoadInfo<F>,
     load_indices: LoadIndices,
     check_bounds: CheckBounds,
+    #[comptime] config: CubeTiling2dConfig,
 ) {
     let coordinates = load_info.coordinates;
-    let config = load_info.config;
+    //let config = load_info.config;
```
nathanielsimard, line 207: dead code?

crates/cubecl-macros/Cargo.toml
```diff
@@ -21,7 +21,20 @@ default = []
...
-derive-new = { workspace = true }
+
+cubecl-common = { path = "../cubecl-common", version = "0.2", default-features = false }
+
+[dev-dependencies]
+compiletest_rs = { version = "0.11", features = ["tmp"] }
+cubecl-core = { path = "../cubecl-core", version = "0.2", default-features = false }
+cubecl-cuda = { path = "../cubecl-cuda", version = "0.2", default-features = false }
+cubecl-linalg = { path = "../cubecl-linalg", version = "0.2", default-features = false }
+cubecl-wgpu = { path = "../cubecl-wgpu", version = "0.2", default-features = false }
```
nathanielsimard, line 39: Why are they needed?

nathanielsimard (commented):
Great job, definitively a big improvement!

crates/cubecl-macros/Cargo.toml
wingertge: For the codegen comparison tests. Could move them but I put them here because they're testing the macro, not the backends in question. Or maybe just use the testgen macro thing.

crates/cubecl-core/src/frontend/element/atomic.rs
```diff
@@ -302,95 +310,68 @@ macro_rules! impl_atomic_int {
...
 
-impl CubePrimitive for AtomicUInt {
+impl CubePrimitive for AtomicU32 {
     fn as_elem() -> Elem {
         Elem::AtomicUInt
     }
 }
 
-impl ExpandElementBaseInit for AtomicUInt {
+impl IntoRuntime for AtomicU32 {
```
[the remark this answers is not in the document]
wingertge: `IntoRuntime` is a requirement on primitives right now to ensure returning constants from kernel functions is smooth. It's also auto-implemented for all `CubeTypes`, so you can expand comptime `CubeType`s into their expanded version for passing them as runtime args.

PR #82, opened 2024-08-24
fix  sync

AntiAnimeGeneral:
Among all backends, the memory order guarantee of WebGPU is the weakest
So Cubecl's memory order cannot be stronger than that of WebGPU,
Cuda
```
__syncthreads();
```
eq to 
Wgsl
```
workgroupBarrier();
storageBarrier();
```
According to
https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#synchronization-functions
https://github.com/gpuweb/gpuweb/discussions/3935 
https://github.com/gpuweb/gpuweb/discussions/4821
Unfortunately, there is no threadfence() function similar to Cuda  in Wgsl


nathanielsimard:
@AntiAnimeGeneral Thanks for the fix. I think we should improve documentation before merging. We should describe what the synchronization does rather than comparing to other functions in other APIs. For `sync_units` and `sync_storage` we should tell that they operation only within a cube, and one synchronize in address space (local variables) and the other synchronize across global address space (global buffers). We could add in a details sections the links you provided for futher reference, but we will add more backend, so the documentation should be for the CubeCL API without having to go in details on all potential backends.

PR #83, opened 2024-08-26
Fix wgpu memory corruption

ArthurBrussee:
After previous wgpu optimizations I missed an import detail which could lead to wrong values being read when using client.create(), which in turn could cause crashes.

This happens when in one queue.submit() call multiple write_buffer_with to the same buffer were queued up. The intention is to say `write -> read -> write -> read` on a buffer, but that would be executed as `write -> write -> read -> read`.

Where things get really mysterious is that simply fixing that bug wasn't enough. If however I also prevent buffers being re-used for the NEXT queue submission, the issue really is resolved! This is not just due to timing - if I place a poll(WAIT) in between every submit the crash still happens.  This is very fishy and needs to be investigated more, but sending this PR now so things are at least stable.

The bug was more likely to be hit with a high `tasks_max`. After this PR, even with tasks_max disabled all together and a manual flush after lots of commands runs fine.


nathanielsimard (approved):
🙏

PR #86, opened 2024-08-28
Handle empty tensors (must bind at minimum 4 bytes on wgpu)

laggui:
Some operations could return empty tensors (e.g., `argwhere` for a tensor where all elements are zero).

Trying to return an empty tensor with 0 byte alloc would fail on wgpu, as encountered in Burn issues ([2180](https://github.com/tracel-ai/burn/issues/2180), [1946](https://github.com/tracel-ai/burn/issues/1946)).

We can handle empty tensors by allocating 4 bytes (minimum required by wgpu).

We get the same result as other backends:
```rust
Tensor {
  data:
[],
  shape:  [0, 1],
  device:  BestAvailable,
  backend:  "fusion<jit<wgpu>>",
  kind:  "Int",
  dtype:  "i32",
}
```

### DIFF
--- a/crates/cubecl-wgpu/src/compute/server.rs
+++ b/crates/cubecl-wgpu/src/compute/server.rs
@@ -207,9 +207,13 @@ where
             .copied()
             .chain(self.copy_handles_used.iter().map(|x| x.0))
             .collect::<Vec<_>>();
-        let memory = self.memory_management.reserve(data.len(), &total_handles);
+        let num_bytes = data.len();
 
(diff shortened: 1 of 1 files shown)


nathanielsimard (approved):

ArthurBrussee:
Wouldn't this also be an issue for kernels using client.empty()? Eg. if after no_true_vals.argwhere() I call .float().exp().

I thought the memory allocator handled allocating buffers of some minimum size behind the scenes.

PR #92, opened 2024-09-03
Fix native vec types

nathanielsimard:
Simplify how we handle vectorized inputs and outputs by always using our custom types. Added the correct reinterprete cast for singular values.


louisfd (approved):
LGTM

PR #93, opened 2024-09-04
Migrate xtask to tracel-xtask and update CI workflow

syl20bnr:


syl20bnr:
Some tests are not passing but that should be expected.

Issue #15, opened 2024-07-18
Matmul CMMA: use less shared memories


louisfd:
At the moment there are always three shared memories: 
- one that is loaded with values converted to f16 from lhs (size B_M x B_K)
- similarly for rhs (size B_K x B_N, dtype=f16)
- one for writing to global memory, only used at the end, whose role is to allow reorganizing the output of the cmma accumulator before storing to output (size B_M x B_N, dtype=same as accumulator, often f32). 

Here are a few observations: 
- Shared memories for lhs and rhs could be merged into one, of size B_M x B_K + B_K x B_N, with rhs simply writing with an offset of B_M x B_K. It is unclear if this will make any speedup, but it's worth a try. 
- In the typical case of B_M = B_N = 64 and B_K = 32, we have B_M x B_K + B_K x B_N = B_M x B_N, so the last shared memory could in fact be a reuse of the old one, which is not used anymore. However that is only possible when accumulating in f16. 
- The final shared memory's use is only to exchange data between units of the same coop group, never with other units of the cube. Therefore we could use something that moves data within the registers of the same warp, probably with `__shfl_sync`.

Issue #13, opened 2024-07-18
Matmul CMMA: support transposed inputs


louisfd:
At the moment, Matmul CMMA assumes contiguous (row major) memory layout. It should work with transposed inputs, leveraging `cmma::MatrixLayout::ColMajor`. But this has implications on loading and writing to global memory, so it's not a trivial change.

PR #94, opened 2024-09-04
Refactor/cmma generalize

### LINKED ISSUE #12: Matmul CMMA: support other vectorizations (or none)
Right now Matmul CMMA assumes a vectorization of four, and will fail otherwise. This implies it only works on matrices whose width is a multiple of 4.

louisfd:
Generalized the cmma algorithm to allow
- Any vectorization factor (Fix #12 )
- Varying block sizes

I reduced the complexity of all the different parameters (block sizes, cube dims and other comptime info) to two degrees of liberty only: one must choose B_MN which is both block size for m and n, and B_K (block size for k), with the additional constraints that these parameters must be divisible by the tile size, and that B_MN must be divisible by B_K. This allows for only a few combinations of B_MN, B_K. 

The combination (32, 32) weirdly adds 0s if tensors are not small, so I ignored it. I don't intend to fix it in the short term because it's not likely an important combination. 

The combination (128, 16) creates a 128x128 shared memory at the write output stage. With the other shared memories this will bust most SMEMs. I will work on removing the need for a final shared memory, as mentioned in #15 . 

In the short term I will also work on #13


louisfd:
See #101 instead

PR #100, opened 2024-09-06
Implemented Vector Normalization

### LINKED ISSUE #98: Implement functions that are specific to vecN types, like normalize.
Currently it's not possible AFAIK to use functions like normalize, reflect, refract, dot, ... that work on vecN types. I have done a small test locally and have gotten this to work:

```rust
#[cube(launch_unchecked)]
fn norm_test<F: Float>(input: &Array<F>, output: &mut Array<F>) {
        output[ABSOLUTE_POS] = F::normalize(input[ABSOLUTE_POS]);
    }
}
```

The idea is that we can just implement these as we would other unary and binary functions, using a vecN type is at the end of the day implemented by using vectorization factor over float arrays, so I don't immediately see a reason why it shouldn't work like this, other than it's a different way of thinking from other shading languages like wgsl.

I will try to make a pull request for normalize in particular, but I would love to hear your feedback on this as well. Is this the way we want to do it, or do you prefer a solution that would allow us to write Array<VecN<F,3>>? I believe that would make it a lot more work, but it might end up being more ergonomic for end users. In addition, I'm not sure if this approach will work for something like .length(), since that has different input and output vectorization factors.

RianGoossens:
Closes #98, or at least moves it forward. This is my first ever contribution to an open-source project so my apologies if I'm doing something very wrong here.

This is a first attempt at allowing you to use vectorized floats as if they are vecN's, with matching functions. As a test I implemented normalize which seems to work both on the WGPU and CUDA backends.

You can now write following code:
```rust
#[cube(launch_unchecked)]
fn norm_test<F: Float>(input: &Array<F>, output: &mut Array<F>) {
        output[ABSOLUTE_POS] = F::normalize(input[ABSOLUTE_POS]);
    }
}
```

And depending on the vectorization factor of your input array it will normalize 1D, 2D, 3D or 4D vectors.

## Problems with xtask
I tried following the contributor guidelines as closely as possible, although I have a problem with the xtask command:

```bash
error: unrecognized subcommand 'pull-request-checks'
```

## Float as vecN, or vecN as a newtype? 
Is this the way forward, or do we want specialized types for vecN's?

## Temporary variables and shadowing
Is my approach for the CUDA implementation correct? I need a temporary variable and the danger of shadowing exists. I made sure the code is in its own scope, but even then the danger exists that it will shadow a preexisting variable. This seems to be the first time that this occurs in the CUDA backend so the solution I came up with is a `unique_variable_name!` macro. Alternatively we could store all variable names in the compiler to make sure we definitely make a unique temporary value, but that sounds pretty complicated for potentially little benefit.


crates/cubecl-cuda/src/compiler/instruction.rs
```diff
@@ -463,3 +465,41 @@ impl Remainder {
...
+struct Normalize;
+
+impl Normalize {
+    fn format(
+        f: &mut core::fmt::Formatter<'_>,
+        input: &Variable,
+        out: &Variable,
+    ) -> core::fmt::Result {
+        let num = input.item().vectorization;
+        let norm = unique_variable_name!("norm");
```
nathanielsimard, line 484: You can simply append "norm" to the output name:

```rust
let norm = format!("{out}_norm");
```

Putting things into another scope is also a good idea.

crates/cubecl-cuda/src/compiler/instruction.rs
```diff
@@ -463,3 +465,41 @@ impl Remainder {
...
+
+        f.write_fmt(format_args!("{{\n"))?;
+        f.write_fmt(format_args!("float {norm} = 0.0;\n"))?;
```
nathanielsimard, line 487: Don't hardcode `float`, is could also be `__half` or `__bfloat`.

```rust
let elem = out.elem();
f.write_fmt(format_args!("{elem} {norm} = {elem}(0);\n"))?;
```
```

File: crates/cubecl-cuda/src/compiler/instruction.rs [hunk 3]
```diff
@@ -463,3 +465,41 @@ impl Remainder {
...
+
+        for i in 0..num {
+            let input_i = input.index(i);
+            f.write_fmt(format_args!("{norm} += {input_i} * {input_i};\n"))?;
+        }
+
+        f.write_fmt(format_args!("{norm} = sqrt({norm});\n"))?;
```
nathanielsimard, line 494: `sqrt` might have a different name based on the element type. See how it is implemented in `unary.rs`

Maybe we could create a new variable 

```rust
enum Varialbe {
   Named(String, Item),
}
```

And reuse the other instructions to build the algo. Just a though.

nathanielsimard (changes requested):

crates/cubecl-cuda/src/compiler/instruction.rs
RianGoossens: I added a helper function for the function prefixes, that could easily be reused in unary.rs as well, I'm just not sure where to put it.

RianGoossens: I changed the helper function, it worked but it felt a bit weird. Instead you can now write ```rust
Sqrt::format_unary(f, input, out, elem)?;
```
and it works for any input and output that are ```Display```. It also automatically works for any `Unary`, not just sqrt. I made an additional `FunctionFmt` trait that can add the function name prefixes for you, I had to do this since `Unary` has the downside of requiring `Component` in it's formatting, which is necessary for `Assign`.

An alternative way of doing it could be by integrating `FunctionFmt` straight into Unary, but then that approach wouldn't work for `Assign` again.

I possibly see what you mean now with the `Variable` suggestion, that would make this work with `Assign` and other functions, by not needing to use `String` anymore for temporary variables.

nathanielsimard (approved):

PR #105, opened 2024-09-09
Implemented magnitude

RianGoossens:
This is AFAIK the first time an implementation turns values of a vectorization factor X to an output of vectorization factor 1. This will be useful for many more functions like .mean(), so I first want to make this pull request before I start making a multitude of new functions.

The way I implemented it is through a new expand function called `fixed_output_unary_expand` with following signature:
```rust
pub fn fixed_output_unary_expand<F>(
    context: &mut CubeContext,
    input: ExpandElement,
    out_item: Item,
    func: F,
) -> ExpandElement
```
It works identical to `unary_expand` but it can take in a different output item, could be a different vectorization, could also be a different type, from the input. In principal this function is strictly more general than `unary_expand`, so perhaps `unary_expand` should just turn into this function. We can use the same concept to implement new binary functions like dot which also have a different vectorization factor for the output.


crates/cubecl-wgpu/src/compiler/wgsl/instructions.rs
```diff
@@ -688,6 +692,9 @@ for (var {i}: u32 = {start}; {i} {cmp} {end}; {increment}) {{
                 "{out} = atomicCompareExchangeWeak({lhs}, {cmp}, {value}).old_value;\n"
             )),
             Instruction::Negate { input, out } => f.write_fmt(format_args!("{out} = -{input};\n")),
+            Instruction::Magnitude { input, out } => {
+                f.write_fmt(format_args!("{out} = length({input});\n"))
```
nathanielsimard, line 696: I think you forgot to call the right format here

nathanielsimard (changes requested):

crates/cubecl-wgpu/src/compiler/wgsl/instructions.rs
RianGoossens: I'm not sure if I understand correctly, but I think you mean one of three things:
- I'm using `format_args!` instead of something like `format!`... but it looks like I'm doing the same method as every other instruction
- The formatting is messed up somehow, but `cargo fmt` doesn't seem to change the file
- Maybe `length` instead of `magnitude`? I chose `magnitude` because `Instruction::Length` already existed, and `magnitude` is in my opinion a more descriptive name anyway. There is zero ambiguity about what `magnitude` is supposed to do.
Or I am missing something else, let me know and I will fix asap!

nathanielsimard: Well it seems like in the cuda backend the magnitude is implemented using the  Euclidean norm, where here it's the length? I don't think that's right.

RianGoossens: Okay, I understand the confusion. `length` is indeed implemented as the euclidean norm in wgsl, aka the length of the vector. You can verify in the wgsl spec where they define it exactly as the square root of the sum of the squares of the components.

This discussion is already a good reason for us to not call it length, but instead magnitude or norm. It's confusing and ambiguous.

RianGoossens: I have added a unit test for magnitude, showing that the outputs are indeed identical for cuda and wgpu, within some precision.

nathanielsimard (approved):

PR #112, opened 2024-09-10
Add support for numeric constants.

booti386:


nathanielsimard:
Can we add some tests to validate for each float type?

booti386:
> Can we add some tests to validate for each float type?

Yes. Also fixed my crappy impl.

booti386:
Minor changes to allow cargo fmt to parse and format the custom macro syntax.

nathanielsimard (approved):

PR #118, opened 2024-09-11
Clean up macro and optimize branch operations

wingertge:
Cleans up the macro variable tracker and applies optimizations to allow for better variable reuse in `while`, `if` and `if_else`.

`if` was just unnecessarily marked as `FnMut` in the variable tracker, which was removed.

`while` and `if_else` both had the same problem: two closures alive at once. Because the first closure (`cond`/`then_block`) can borrow variables that are also used in the second (`body`/`else_block`), variables had to always be cloned and couldn't be consumed.
The solution for `while` was to desugar it at compile time into a simple loop.
For `if_else`, I had to break the function into two parts to tell Rust that the `then_block` was consumed before `else_block`, so `else_block` can safely consume variables borrowed in `then_block`.

# Testing
All tests pass in `cubecl` and `burn`.  Frontend tests were modified to account for the variables that can now be reused.


nathanielsimard (approved):
LGTM

PR #119, opened 2024-09-11
Add `for in` syntax for immutable arrays, tensors and slices

wingertge:
Adds `for item in array` syntax for `&Array`, `&Tensor` and `Slice`. This currently only implements immutable iterators until we can figure out a way to implement mutable references into arrays.

# Testing
Adds a frontend and a runtime test to ensure it works properly.

### DIFF
--- a/crates/cubecl-core/src/frontend/element/array.rs
+++ b/crates/cubecl-core/src/frontend/element/array.rs
@@ -3,7 +3,8 @@ use std::{marker::PhantomData, num::NonZero};
 use crate::{
     compute::{KernelBuilder, KernelLauncher},
     frontend::CubeType,
-    ir::{Item, Vectorization},
+    ir::{Branch, Item, RangeLoop, Vectorization},
+    prelude::{CubeIndex, Iterable},
     unexpanded, KernelSettings, Runtime,
 };
 use crate::{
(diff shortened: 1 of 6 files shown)


crates/cubecl-core/src/frontend/element/array.rs
```diff
@@ -253,3 +254,56 @@ impl<'a, R: Runtime> ArrayHandleRef<'a, R> {
...
+            scope: child.into_scope(),
+        }));
+    }
+
+    fn expand_unroll(
+        self,
+        _context: &mut CubeContext,
+        _body: impl FnMut(&mut CubeContext, <T::Item as CubeType>::ExpandType),
+    ) {
+        unimplemented!("Can't unroll array iterator")
```
nathanielsimard, line 307: it would be cool to support unroll. I guess we would need a way to know if the len is comptime as well.

wingertge: Unfortunately there currently isn't infrastructure for that. Could make it work for shared memory and local arrays in theory, but we'd need to add the comptime size to the `ExpandElement` or something like that to read it in the iterator.

nathanielsimard (approved):

PR #125, opened 2024-09-17
Skip zero initialization of workgroup memory

ArthurBrussee:
CUDA kernels don't rely on workgroup memory being initialized to anything in particular, so no need to require this in wgpu, as it's not like cube kernels can rely on this fact anyway. On WebGPU this still happens for security reasons, but on native platforms this can be skipped.

Should save some marginal amount of performance, and maybe some kernels accidentally working on wgpu but not cuda.

### DIFF
--- a/crates/cubecl-wgpu/src/compute/server.rs
+++ b/crates/cubecl-wgpu/src/compute/server.rs
@@ -111,7 +111,10 @@ where
                     layout: None,
                     module: &module,
                     entry_point: "main",
-                    compilation_options: Default::default(),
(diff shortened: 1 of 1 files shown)


nathanielsimard:
Somehow some tests are not working anymore. It's probably a problem in the test setup, but I'm not sure @louisfd

louisfd:
@ArthurBrussee Done, you can rebase on main

ArthurBrussee:
Great! Rebased and tests are passing now

PR #127, opened 2024-09-18
Add macro to automatically generate common cases for autotune

wingertge:
Adds a small proc macro to generate the autotune boilerplate in most common cases.
See examples in doc comments for how they would be used.

### DIFF
--- a/crates/cubecl-core/src/lib.rs
+++ b/crates/cubecl-core/src/lib.rs
@@ -23,7 +23,7 @@ pub use codegen::*;
 pub use pod::*;
 pub use runtime::*;
 
-pub use cubecl_macros::{comptime, cube, CubeLaunch, CubeType};
+pub use cubecl_macros::*;
 pub use cubecl_runtime::benchmark;
 
 /// An approximation of the subcube dimension.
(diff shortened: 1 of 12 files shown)


nathanielsimard:
Just to make sure, this creates the tune operation as well as the autotune set from an autotune key, but it doesn't create the high level function actually creating a static tuner right?

wingertge:
> Just to make sure, this creates the tune operation as well as the autotune set from an autotune key, but it doesn't create the high level function actually creating a static tuner right?

Exactly. That boilerplate is extremely minor so I don't think it's worth making a macro for.

nathanielsimard:
> Exactly. That boilerplate is extremely minor so I don't think it's worth making a macro for. [Here's an example of the usage in the new `conv2d` tuner](https://github.com/wingertge/burn/blob/local-dev/crates/burn-jit/src/kernel/conv/conv2d/tune/conv2d.rs)

We could actually generate it as well, since it's the same signature as the `tune_set` function.

We could have another macro `tune` that does most of the thing:

```rust
#[tune(
    operations(Conv2dDirect, Conv2dIm2col),
    key = create_key,
    inputs = bench_inputs
)]
pub fn conv2d<R: JitRuntime, E: FloatElement, I: IntElement>(
    key: JitAutotuneKey,
    input: JitTensor<R, E, 4>,
    weights: JitTensor<R, E, 4>,
    bias: Option<JitTensor<R, E, 1>>,
    options: ConvOptions<2>,
) -> JitTensor<R, E, 4>;
```

What do you think?

nathanielsimard (approved):

PR #139, opened 2024-09-20
Extract variable allocation strategy to trait to allow backends to use more optimal strategies

wingertge:
Extracts the variable allocation/reuse strategy into a trait to allow backends to generate variables in a way that's beneficial to the output format.

* Splits local creation into `create_local_binding` and `create_local_variable` for intermediates and mutable variables respectively.
* Implements `ReuseAllocator`, which replicates the current behaviour, as well as  `HybridAllocator` which uses SSA for intermediates, and the normal reuse allocator for variables
* Implements `LocalBinding` generation for `wgpu` only for now. CUDA is a future goal, which would also enable switching local allocators per kernel, but I want the `burn` test sweet to work on CUDA before I can comfortably port
* Switches the default local allocator for `wgpu` to `HybridAllocator`, which performs as well as or better than `ReusingAllocator` and produces more optimized SPIR-V

## Testing
Both `cubecl` and `burn` fully pass their test suites. Behaviour of `HybridAllocator` is validated with runtime tests.


crates/cubecl-core/src/frontend/operation/base.rs
```diff
@@ -14,8 +14,8 @@ pub(crate) fn binary_expand<F>(
 where
     F: Fn(BinaryOperator) -> Operator,
 {
-    let lhs_var: Variable = *lhs;
-    let rhs_var: Variable = *rhs;
+    let lhs = lhs.consume();
```
nathanielsimard, line 17: Ok so I guess consume reduces the ref count potentially making the variable reusable, removing the need to manually check `can_mut`

crates/cubecl-core/src/ir/local_allocator.rs
```diff
@@ -0,0 +1,117 @@
...
+        // Insert it in the variable pool for potential reuse
+        let new = ExpandElement::Managed(Rc::new(root.borrow_mut().create_local(item)));
+        self.pool.insert(new.clone());
+
+        new
+    }
+
+    fn create_local_binding(&self, root: ScopeRef, scope: ScopeRef, item: Item) -> ExpandElement {
+        self.create_local_variable(root, scope, item)
+    }
```
nathanielsimard, line 94: I'm not sure I undetstand the fundamental difference between `local_variable` and `local_binding`.

nathanielsimard: What is the logic behind calling `local_binding` vs `local_variable` ? When do you know you should do one vs the other?

crates/cubecl-core/src/codegen/compiler.rs
```diff
@@ -17,6 +17,7 @@ pub trait Compiler: Sync + Send + 'static + Clone + Default + core::fmt::Debug {
     fn compile(kernel: KernelDefinition, mode: ExecutionMode) -> Self::Representation;
     /// The size of the given element in bytes.
     fn elem_size(elem: Elem) -> usize;
+    fn local_allocator() -> impl LocalAllocator;
```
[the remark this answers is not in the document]
wingertge: It was added in one of the semi-recent Rust versions, I believe it does work like an invisible associated type but I'm not sure about the implementation details. It was added in the same version that added async functions in traits, Rust 1.75. It just works like it does outside of traits in terms of actual usage.

crates/cubecl-core/src/ir/local_allocator.rs
wingertge: Local bindings are things that are only assigned to once. So intermediates, immutable variables, etc. The term that's mostly used these days is "binding", which is why I called it that. It would be `let x = ` in `wgsl`, as opposed to "variables", which are a `var x` in `wgsl` and can be reassigned. In the generated SPIR-V, `binding`s would become normal SSA ops (i.e. `%32 = AddOp %30 %31`), while `variable`s would become a variable declaration, along with a `LoadOp` for accessing it and a `StoreOp` for writing to it. An `AddOp` with a `variable` would result in code like this
```rust
%30 = LoadOp %3
%31 = LoadOp %5
%32 = AddOp %30 %31
StoreOp %32 %3
```

wingertge: The important difference is that bindings don't necessarily exist after compilation, they're just labels for intermediate values. Variables do exist at runtime and must be used with load/store.

wingertge: As for when you'd use either: you'd use a local binding by default, and only use a variable when you need to reassign the value (i.e. `let mut` or `vectorized_empty`).

crates/cubecl-core/src/frontend/operation/base.rs
```diff
@@ -55,34 +43,21 @@ pub(crate) fn binary_expand_no_vec<F>(
...
-    let out = if lhs.can_mut() && item_lhs == item {
-        lhs
-    } else if rhs.can_mut() && item_rhs == item {
-        rhs
-    } else {
-        context.create_local(item)
-    };
-
-    let out_var = *out;
+    let output = context.create_local_binding(item);
```
nathanielsimard, line 53: If I do the following:

```rust
let mut out = x + y;
```
Do we call this function? Would it be problematic since the out value may be reused?

wingertge: Creating a mutable local calls `Init::init` on the initializer, which assigns it to a new mutable local variable. So in `wgpu` this would be
```rust
let _10 = _8 + _9;
l_1_6 = _10;
```
or in SPIR-V
```
%10 = AddOp %8 %9
StoreOp %10 %out_var
```
See this code
```rust
pub fn init_expand<F>(context: &mut CubeContext, input: ExpandElement, func: F) -> ExpandElement
where
    F: Fn(UnaryOperator) -> Operator,
{
    if input.can_mut() {
        return input;
    }

    let input_var: Variable = *input;
    let item = input.item();

    let out = context.create_local_variable(item);
    let out_var = *out;

    let op = func(UnaryOperator {
        input: input_var,
        out: out_var,
    });

    context.register(op);

    out
}
```
`can_mut` will always return false for plain variables (i.e. local bindings).

nathanielsimard: Perfect, that answer my question :)

nathanielsimard (approved):

PR #142, opened 2024-09-23
Add simple fusion example

nathanielsimard:


examples/fusing/src/lib.rs
```diff
@@ -0,0 +1,88 @@
...
+    outputs: &mut Sequence<Array<F>>,
+    #[comptime] ops: Sequence<Operation>,
+) {
+    #[unroll]
+    for index in 0..ops.len() {
+        let op = comptime! { ops.index(index) };
+        let input = inputs.index(op.input_index);
+        let output = outputs.index_mut(op.output_index);
+
+        if op.kind == 0 {
```
louisfd, line 22: Yeah, it's definitely time to support enums

crates/cubecl-macros/src/parse/expression.rs
```diff
@@ -118,11 +118,8 @@ impl Expression {
                     && args.iter().all(|arg| arg.is_const())
                     && method.method != "runtime"
                 {
-                    let receiver = receiver.as_const(context).unwrap();
-                    let method = &method.method;
-                    let args = args.iter().map(|it| it.to_tokens(context));
                     Expression::Verbatim {
-                        tokens: quote![#receiver.#method(#(#args),*)],
+                        tokens: quote![#method],
```
louisfd, line 122: Not sure why you can safely simplify this

louisfd (approved):
Awesome PoC for fusion in cube :D

crates/cubecl-macros/src/parse/expression.rs
nathanielsimard: It was actually a bug, when the arguments as well as the method are const, we need to call the normal method with the normal arguments, not mix both (`sequence.index(u32)`, not `sequence.index(ExpandedElemTyped::from_lit(u32)` nor `sequence.__index_expand(...)`).

PR #147, opened 2024-09-24
Add support for comptime arrays available as runtime constants

wingertge:
Adds a new function to `Array`, `from_data` which allows creating arrays from comptime data. Any type implementing `IntoIterator<Item = CubePrimitive>` can be used and the data will automatically be cast to the correct type. The auto-casting is required because you currently cannot pass anything containing floats as comptime variables (because floats don't implement `Eq` or `Hash`).

If actual floats are required, a `comptime!` function could be used to bitcast the input.

# Testing
Adds codegen and runtime tests for constant arrays.


crates/cubecl-cuda/tests/constant_array.cu
```diff
@@ -0,0 +1,34 @@
...
+  uint idxGlobal =
+      (absoluteIdx.z * gridDim.x * blockDim.x * gridDim.y * blockDim.y) +
+      (absoluteIdx.y * gridDim.x * blockDim.x) + absoluteIdx.x;
+  uint rank = info[0];
+  uint rank_2 = rank * 2;
+  const float arrays_0[3] = {
+      uint(3),
+      uint(5),
+      uint(1),
+  };
```
nathanielsimard, line 18: isn't strange that the values are casted to `uint`?

crates/cubecl-wgpu/tests/constant_array.wgsl
```diff
@@ -0,0 +1,28 @@
+@group(0)
+@binding(0)
+var<storage, read_write> output_0_global: array<f32>;
+
+@group(0)
+@binding(1)
+var<storage, read_write> info: array<u32>;
+
+const arrays_0: array<f32, 3> = array(f32(3u),f32(5u),f32(1u),);
```
nathanielsimard, line 9: Looks better in wgsl

crates/cubecl-cuda/tests/constant_array.cu
wingertge: I don't know what's up with that either, that's what the backend does when I just print the value. I might just force a cast to the output type.

crates/cubecl-wgpu/tests/constant_array.wgsl
wingertge: `wgsl` uses the `fmt_cast` function I introduced in the variable allocator PR to add casts where necessary. CUDA doesn't have that rn.

nathanielsimard (approved):
LGTM, we simply need to fix the conflicts

PR #152, opened 2024-09-26
Add select (ternary)

wingertge:
## Changes

Adds a `select` function that represents ternaries in CUDA, or `select` in `wgsl`.
I decided to use the name from `wgsl` because it represents the concept quite well, but used the argument order of ternaries because that's much more intuitive. `wgsl` uses `select(or_else, then, cond)` but that's entirely down to it being derived from the `mix` interpolation function in GLSL. So I think we should keep the more intuitive `cond`, `then`, `or_else` order.

## Reasoning
Select limits branching and can allow for completely SSA kernels. `wgsl` will execute both branches for example, then (afaik) executes `then & cond + or_else & !cond`. This allows conditional assign without any branching or intermediate variables.

## Important caviat
Since the goal of this is to *avoid* branching, both values will be evaluated regardless of the condition. This cannot safely be used for things like bounds checking (unless we know the access is safe regardless of the condition, i.e. checking if convolution position is in padding).

## Testing

Adds runtime tests to ensure `select` works properly.

### DIFF
--- a/crates/cubecl-core/src/frontend/operation/branch.rs
+++ b/crates/cubecl-core/src/frontend/operation/branch.rs
@@ -1,5 +1,12 @@
 use crate::prelude::CubePrimitive;
 
+/// Executes both branches, *then* selects a value based on the condition. This *should* be
+/// branchless, but might depend on the compiler.
+///
+/// # Safety
+///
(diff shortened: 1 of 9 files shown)


nathanielsimard (approved):
LGTM

PR #156, opened 2024-10-04
Fix wgpu memory corruption for CubeCount::Dynamic

ArthurBrussee:
A previous PR #56 has made it a bit trickier to deal with wgpu resources, and turns out there's another memory corruption possible now :) Resources used for `CubeCount::Dynamic` were not registered as "in use" which means it was possible for a copy operation & dispatch to be ordered incorrectly. 

Another problem is that `get_resource` is public which means users might add operations using a buffer themselves. These should also be tracked! The fix is to always register a buffer as 'used' if it's used in `get_resource`. 

This fix also seems to alleviate the need for the mysterious "double use for used copy buffers" that was previously needed, mystery solved!

#56 has introduced two memory corruptions now, so it's clearly a bit too flaky... some simpler strategy to enable fast wgpu copying might be needed but don't have much inspiration atm!


crates/cubecl-wgpu/src/compute/server.rs
```diff
@@ -196,11 +194,21 @@ where
...
-        self.memory_management.get_resource(
-            binding.memory,
-            binding.offset_start,
-            binding.offset_end,
-        )
+        // Keep track of any buffer that might be used in the wgpu queue, as we cannot copy into them
+        // after they have any outstanding compute work. Calling get_resource repeatedly
+        // will add duplicates to this, but that is ok.
+        let handle = self.memory_management.get(binding.memory.clone());
+        self.storage_in_flight.push(handle.id);
```
nathanielsimard, line 201: We could remove duplicates by using a set? If the `storage_in_flight` is cleared often it's probably not worth it.

nathanielsimard: I think we should actually probably create a data structure with methods. We use a slice in the memory management functions, but it could be a struct with the proper methods. Then we could change the data structure more easily (Vec under size of 10-20, a set if more than that).

crates/cubecl-wgpu/src/compute/server.rs
```diff
@@ -373,18 +364,7 @@ where
...
-            // For some unknown reason, we have to make sure
-            // a buffer isn't used more than once not just in the current
-            // submission, but also in the next one.
-            //
-            // This really needs a better explanation of why this is, or
-            // some investigation, maybe it's a wgpu bug.
-            x.1 += 1;
-            x.1 < 2
-        });
+        self.storage_in_flight.clear();
```
nathanielsimard, line 367: Why syncing means that we can remove the external readonly buffers from the list?

nathanielsimard (approved):

PR #157, opened 2024-10-06
IR Optimizer and changes to SPIR-V compiler

wingertge:
- Adds an optimizer that operates on the `cubecl` IR and makes use of it in the SPIR-V compiler. 
- Adds a debug representation for the IR so it can be more easily inspected.
- Adds `Versioned` variable type for SSA transformer
- Adds `InitLine` operation to create a vector from a set of variables
- Adds `Copy` and `CopyBulk` operations to IR for direct memory to memory copies. `Copy` is currently created by the optimizer, while `CopyBulk` requires calling a new function.
- Adds `create_local_binding` to scope to create intermediate bindings for things like the `for` loop condition
- Adds `index_unchecked` and `index_assign_unchecked` methods to frontend to somewhat ammortize Vulkan's lower maximum vectorization for data loading where in-bounds is guaranteed.


crates/cubecl-core/src/frontend/operation/copy.rs
```diff
@@ -0,0 +1,34 @@
+use crate::prelude::{CubePrimitive, Slice, SliceMut};
+
+pub fn copy_bulk<C: CubePrimitive>(
```
nathanielsimard, line 3: It would be great to add documentation here.

crates/cubecl-opt/src/block.rs
```diff
@@ -0,0 +1,53 @@
+use std::{cell::RefCell, collections::HashSet, rc::Rc};
+
+use cubecl_core::ir::{Operation, Variable};
+use petgraph::graph::NodeIndex;
+use stable_vec::StableVec;
+
+use crate::{version::PhiInstruction, ControlFlow, Optimizer, Program};
+
+#[derive(Default, Debug, Clone)]
+pub struct BasicBlock {
```
nathanielsimard, line 10: I think it's important to add documentation on the new data structures added to the new crate.

crates/cubecl-opt/src/block.rs
```diff
@@ -0,0 +1,53 @@
...
+
+impl Optimizer {
+    pub fn visit_all(
+        &mut self,
+        mut visit_read: impl FnMut(&mut Self, &mut Variable) + Clone,
+        mut visit_write: impl FnMut(&mut Self, &mut Variable) + Clone,
+    ) {
+        for node in self.program.node_indices().collect::<Vec<_>>() {
+            let phi = self.program[node].phi_nodes.clone();
+            let ops = self.program[node].ops.clone();
```
nathanielsimard, line 27: If it's only one operation, I would call the variable `op` instead of `ops`.

crates/cubecl-opt/src/lib.rs
```diff
@@ -0,0 +1,359 @@
+use std::{
```
nathanielsimard, line 1: I would add a description of the lib and how it can be used to optimize the cubecl ir.

crates/cubecl-opt/src/lib.rs
```diff
@@ -0,0 +1,359 @@
...
+#[derive(Default, Clone, Copy, PartialEq, Eq, Debug)]
+struct Range {
+    lower_bound: Option<i64>,
+    upper_bound: Option<i64>,
+}
+
+#[derive(Debug, Clone)]
+pub struct Optimizer {
+    program: Program,
+    pub current_block: Option<NodeIndex>,
```
nathanielsimard, line 99: Is there a reason to have this as public? Is it mutated by each optimization then reset after? It's a bit weird to have a counter/ptr as a public member of the optimizer.

crates/cubecl-opt/src/lib.rs
```diff
@@ -0,0 +1,359 @@
...
+                }
+                mut other => {
+                    self.visit_operation(&mut other, |_, _| {}, |opt, var| opt.write_var(var));
+                    self.current_block_mut().ops.borrow_mut().push(other);
+                }
+            }
+        }
+    }
+
+    fn compile_procedure(&mut self, proc: Procedure, mut scope: Scope) {
```
nathanielsimard, line 308: If the optimizer also responsible for recompiling the scope?

crates/cubecl-opt/src/passes/array_copy_propagate.rs
```diff
@@ -0,0 +1,121 @@
...
+            }
+            replace_const_arrays(opt, arr_id, &vars);
+            changes.inc();
+        }
+    }
+}
+
+#[derive(Clone, Copy, Debug)]
+struct Array {
+    id: (u16, u8),
```
nathanielsimard, line 32: Why two numbers for the id? Is it the variable id then the vectorization factor? Could it be 2 fields?

crates/cubecl-opt/src/passes/composite.rs
```diff
@@ -0,0 +1,124 @@
...
+use cubecl_core::ir::{
+    BinaryOperator, Item, LineInitOperator, Operation, Operator, UnaryOperator, Variable,
+};
+use stable_vec::StableVec;
+
+use crate::{AtomicCounter, Optimizer};
+
+use super::OptimizationPass;
+
+pub struct CompositeMerge;
```
nathanielsimard, line 12: It would be cool to add documentation for each optimization pass, saying what they do, why they are useful and an example of the changes made on the IR.

crates/cubecl-opt/src/passes/integer_range_analysis.rs
```diff
@@ -0,0 +1,263 @@
...
+            lower_bound: Some(0),
+            upper_bound: Some(upper),
+        }
+    }
+}
+
+#[derive(Default, Clone, Debug)]
+pub struct IntegerRangeAnalysis;
+
+impl OptimizationPass for IntegerRangeAnalysis {
```
nathanielsimard, line 28: It doesn't feel right that an analysis implements an optimization pass. I would expect two traits, one for analysis and one for the optimization. Optimizations might use the results of the analyses stored in the optimizer's state. 

```
Step 1 => Initialize the optimizer
Step 2 => Run all analyses and store their results in the optimizer's state.
Step 3 => Run all optimization pass that can access all analyses previously computed.
Step 4 => Generate or return the transformed CubeCL IR.
Step 5 => Enjor speed 😎
```

It might be the current workflow, it's just not named that way.

crates/cubecl-opt/src/passes/integer_range_analysis.rs
```diff
@@ -0,0 +1,263 @@
...
+                    _ => {}
+                }
+            }
+        }
+    }
+}
+
+pub(crate) fn range_of(opt: &Optimizer, var: &Variable) -> Range {
+    match var {
+        Variable::Versioned {
```
nathanielsimard, line 102: What is a versioned variable? Is it only for the optimization?

crates/cubecl-opt/src/version.rs
```diff
@@ -0,0 +1,194 @@
...
+}
+
+#[derive(Debug, Clone)]
+pub struct PhiEntry {
+    pub block: NodeIndex,
+    pub value: Variable,
+}
+
+#[derive(Debug, Clone)]
+pub struct PhiInstruction {
```
nathanielsimard, line 28: What is Phi refering to? If there is a link to some information about the concept, that would be great!

crates/cubecl-opt/src/passes/array_copy_propagate.rs
wingertge: It's the variable ID and depth. Since I'm working directly on the IR I need to mirror the ID scheme of the IR.

crates/cubecl-opt/src/passes/integer_range_analysis.rs
wingertge: The reason they're combined is because some analyses should run after certain optimizations but before certain other optimizations that depend on them. For example, integer analysis is greatly simplified by first doing assignment merging and constant evaluation, but must be done before in bounds transformation. Maybe I can just rename the trait to `OptimizerPass` to make it a bit more clear it's just anything the optimizer executes as part of its optimization loop.

crates/cubecl-opt/src/passes/integer_range_analysis.rs
wingertge: Yes, it's a local variable that has been transformed into an immutable by versioning it with each assignment. It should be treated the same as a `LocalBinding`, but needs the extra version number to make SSA transformation possible (since we don't have the `CubeContext` to allocate new binding IDs when we get to the optimizer, so we have to instead version).

crates/cubecl-opt/src/block.rs
wingertge: It's the set of all operations in the block. So it's multiple. We iterate through it below.

crates/cubecl-opt/src/lib.rs
wingertge: Yes, the optimizer completely takes over compiling of the scope. Crates using the optimizer should only need to work off the optimizer's `BasicBlock`s. This means it also needs to push the procedure instructions.

nathanielsimard (approved):

PR #158, opened 2024-10-06
Rework & add ExclusivePages as an option for MemoryManagement 

ArthurBrussee:
# Overview 

The current memory strategy is not entirely compatible with wgpu - sub slices can't have different read / read_write annotations, which causes issues with the current allocator.

There is a "simple" allocator to work around this but it's not very performant nor does it have some things like managing alignment properly. This PR reworks the MemoryManagement setup a bit and adds a new option to use "ExclusivePages" which does not allocate any sub-slices.

## Performance

Memory wise, I suspect this allocator does have bad worst case scenarios. Allocating increasingly big allocations can leave a "trail" of old allocations in pools do not get deallocated. However, I tested this on a practical example (training splats for 1000/2000/3000 steps) and so far saw a bit _less_ memory use compared to the slices allocator. The allocator uses the bin layout of [this](https://github.com/sebbbi/OffsetAllocator) allocator which is claiming some nice properties so perhaps that helps.

The CPU performance should be slightly better than memory management with sub slices as looking for a page is simpler. Additionally, pools of the right size are now found with a binary search.

## Rework

The ExclusivePages allocator is based on the previous "small" memory pool. The slices allocator still uses one of these pools for the smallest allocations where offsets don't work.

This PR also reworks some related coda:
- There is a MemoryDeviceProperties struct now. Servers no longer have a generic Properties or FeatureSet. Instead a ClientProperties holds both a set of features and memory properties uniformally.
- The cuda device used to be lazily initialized, but that wasn't compatible with the need to construct the client with a known max memory size. However, the current wmma feature detection already immediatly instantiated the client anyway, so this lazy init could be removed.
- The memory pools don't have a "rounding strategy" anymore. I believe that for the slice allocator no rounding was actually broken anyway, and didn't make much sense to use imo.
- There's a new `MemoryUsage` struct with some stats, that can be read from a ComputeClient.
- Removed 'Chunk' in pools - it was redundant with memory page, which is a more standard terminology
- Added a method to initialize CUDA with RuntimeOptions like wgpu.
- Removed StorageUtilization. Only custom allocs (which aren't really used) used StorageUtilization::Full. It's easier to just represent everything as a slice which may or may not be the whole storage.
- Removed MemoryPoolHandle / MemoryPoolBinding and just use slice handles directly.
- Added some more tests to the MemoryManagement.

## Future

- After this PR I'd like to delete the old simple memory management and potentially the MM: MemoryManagement in a bunch of places. I think it's likely better to configure the current dynamic memory allocator with more options rather than allowing wholesale replacements. It's always possible to dispatch cheaply to another allocator anyway.

- When simple memory management is removed something needs to tell cube whether the memory allocator is wgpu spec compliant and whether it can add proper read/read_write tags.

- Probably need to measure memory usage / performance in a bit more detail. I also think this allocator might be a hard sell when pages can never be deallocated, so perhaps it needs a way for an occasional page deallocation again.


crates/cubecl-cuda/src/runtime.rs
```diff
@@ -17,15 +17,81 @@ use crate::{
...
+type Channel = MutexComputeChannel<Server>;
 
-const MEMORY_OFFSET_ALIGNMENT: u32 = 32;
+static RUNTIME: ComputeRuntime<CudaDevice, Server, Channel> = ComputeRuntime::new();
 
-type Server = CudaServer<DynamicMemoryManagement<CudaStorage>>;
+const MEMORY_OFFSET_ALIGNMENT: usize = 32;
+
+/// Initialize a client on the given device with the given options. This function is useful to configure the runtime options
+/// or to pick a different graphics API. On wasm, it is necessary to use [`init_async`] instead.
```
nathanielsimard, line 38: I think the comment is not up to date.

crates/cubecl-runtime/src/feature_set.rs
```diff
@@ -0,0 +1,41 @@
+use crate::memory_management::MemoryDeviceProperties;
+use std::collections::BTreeSet;
+
+/// The set of [features](Feature) supported by a [runtime](Runtime).
+#[derive(Debug)]
+pub struct ClientProperties<Feature: Ord + Copy> {
```
nathanielsimard, line 6: I don't think the name is right, yes it's accessed by the client, but isn't it better to name it `DeviceProperties` instead?

crates/cubecl-runtime/src/memory_management/base.rs
```diff
@@ -1,5 +1,56 @@
+use core::fmt::Debug;
+
 use crate::storage::{ComputeStorage, StorageHandle, StorageId};
 
+/// Amount of memory in use by this allocator
+/// and statistics on how much memory is reserved and
+/// wasted in total.
+pub struct MemoryUsage {
```
nathanielsimard, line 8: This is gold 🙏

crates/cubecl-runtime/src/memory_management/base.rs
```diff
@@ -1,5 +1,56 @@
...
+        // the 5 biggest allocations, to show when you an OOM.
+        let usage_percentage = (self.bytes_in_use as f32 / self.bytes_reserved as f32) * 100.0;
+        let padding_percentage = (self.bytes_padding as f32 / self.bytes_in_use as f32) * 100.0;
+        writeln!(f, "Memory Usage Report:")?;
+        writeln!(f, "  Number of allocations: {}", self.number_allocs)?;
+        writeln!(f, "  Bytes in use: {} bytes", self.bytes_in_use)?;
+        writeln!(f, "  Bytes used for padding: {} bytes", self.bytes_padding)?;
+        writeln!(f, "  Total bytes reserved: {} bytes", self.bytes_reserved)?;
+        writeln!(f, "  Usage efficiency: {:.2}%", usage_percentage)?;
+        writeln!(f, "  Padding overhead: {:.2}%", padding_percentage)
```
nathanielsimard, line 50: Could we pretty format the bytes (kb, Mb, Gb, etc.)? It would be easier to debug.

crates/cubecl-runtime/src/memory_management/dynamic.rs
```diff
@@ -1,111 +1,207 @@
...
-#[derive(new, Debug)]
-pub struct DynamicMemoryManagementOptions {
-    pools: Vec<MemoryPoolOptions>,
-    min_chunk_alignment_offset: usize,
-}
+// Bin sizes as per https://github.com/sebbbi/OffsetAllocator/blob/main/README.md
+// This guarantees that _for bins in use_, the wasted space is at most 12.5%. So as long
+// as bins have a high use rate this should be fairly efficient. That said, currently slices in
+// bins don't deallocate, so there is a chance more memory than needed is used.
+const EXP_BIN_SIZES: [usize; 200] = [
```
nathanielsimard, line 19: Is there a function that can generate those bins? It would be cool to have it here as a comment!

crates/cubecl-runtime/src/memory_management/dynamic.rs
```diff
@@ -1,111 +1,207 @@
...
-/// Reserves and keeps track of chunks of memory in the storage, and slices upon these chunks.
-pub struct DynamicMemoryManagement<Storage> {
-    min_chunk_alignment_offset: usize,
-    small_memory_pool: SmallMemoryPool,
-    pools: Vec<MemoryPool>,
-    options: Vec<MemoryPoolOptions>,
-    storage: Storage,
+enum DynamicPool {
+    Sliced(SlicedPool),
+    Simple(ExclusiveMemoryPool),
```
nathanielsimard, line 12: Could we rename Simple to Exclusive?

crates/cubecl-runtime/src/memory_management/memory_pool/exclusive_pool.rs
```diff
@@ -0,0 +1,137 @@
...
+///
+/// - Only one slice is supported per page, due to the limitations in WGPU where each buffer should only bound with
+///   either read only or read_write slices but not a mix of both.
+/// - The pool uses a ring buffer to efficiently manage and reuse pages.
+///
+/// Fields:
+/// - `pages`: A hashmap storing the allocated pages by their IDs.
+/// - `slices`: A hashmap storing the slices by their IDs.
+/// - `ring_buffer`: A vector used as a ring buffer to manage page reuse.
+/// - `index`: The current position in the ring buffer.
```
nathanielsimard, line 19: Is it necessary to add the comment on the fields? That comment could become false and outdated quite soon.

crates/cubecl-cuda/src/runtime.rs
ArthurBrussee: Fixed

crates/cubecl-runtime/src/feature_set.rs
ArthurBrussee: Yeah makes sense, fixed!

crates/cubecl-runtime/src/memory_management/base.rs
ArthurBrussee: Yeah good idea! Added some formatting. Went with base 1000 but some other tools (nvidia-smi) use kibi / mebibytes, not kilo/megabytes. Idk what's easier for people

crates/cubecl-runtime/src/memory_management/dynamic.rs
ArthurBrussee: Done

crates/cubecl-runtime/src/memory_management/memory_pool/exclusive_pool.rs
ArthurBrussee: Oh yeah the original code had this but agreed, don't like it, reminds me of python :P

crates/cubecl-runtime/src/memory_management/base.rs
nathanielsimard: Good enough I guess

nathanielsimard (approved):
🚀

PR #160, opened 2024-10-08
Delete simple memory management

ArthurBrussee:
After #158, it's probably always better to use ExclusivePages rather than simple memory management.

This PR removes the old simple allocator all together and switches wasm to ExclusivePages by default.

This PR has also meant there is no need for servers anymore to be generic over their memory management. I think future memory strategies are better integrated into MemoryManagement, sharing some common code & dealing with all the memory quirks, rather than being wholesaled replaceable. This then means Handle's and Bindings don't have to depend on the runtime, which solves some quirks around eg. CubeCount. That does have the potential to propagate to a _lot_ of places in cubecl as a bunch of types are generic over the runtime there because of the handle. I think in some places there that is desirable as otherwise there's a ton of places where the runtime needs to be specified, so I've slapped a PhantomData on some of these now.

Otherwise, there was a feature you could enable called simple-memory-management which affected how cube compiled. I think ideally this would just depend automatically on how your memory allocator is configured, but, that doesn't really work out. The problem is that memory strategy is per client right now, which means that it would couple cube shaders to a specific client, and that doesn't seem great. Maybe allocator configuration needs to be global / once at startup only... not sure. Currently I've just added/renamed the current feature to exclusive-memory-only.

Lastly, a somewhat related change is to remove ResourceKind::Slice / Full as "full" alloc was removed in #158 .

### DIFF
--- a/crates/cubecl-runtime/src/server.rs
+++ b/crates/cubecl-runtime/src/server.rs
@@ -143,11 +143,13 @@ impl Clone for Binding {
     }
 }
 
-/// Provides launch information specifying the number of work groups to be used by a compute shader.
+/// Specifieds the number of cubes to be dispatched for a kernel.
+///
+/// This translates to eg. a grid for CUDA, or to num_workgroups for wgsl.
 pub enum CubeCount {
-    /// Dispatch x,y,z work groups.
+    /// Dispatch a known count of x, y, z cubes.
     Static(u32, u32, u32),
-    /// Dispatch work groups based on the values in this buffer. The buffer should contain a u32 array [x, y, z].
+    /// Dispatch an amount based on the values in this buffer. The buffer should contain a u32 array [x, y, z].
     Dynamic(Binding),
 }
 
--- a/crates/cubecl-runtime/tests/dummy/server.rs
+++ b/crates/cubecl-runtime/tests/dummy/server.rs
@@ -17,8 +17,8 @@ use super::DummyKernel;
 /// The dummy server is used to test the cubecl-runtime infrastructure.
 /// It uses simple memory management with a bytes storage on CPU, without asynchronous tasks.
 #[derive(new, Debug)]
-pub struct DummyServer<MM = MemoryManagement<BytesStorage>> {
-    memory_management: MM,
+pub struct DummyServer {
+    memory_management: MemoryManagement<BytesStorage>,
 }
 
 impl ComputeServer for DummyServer {


crates/cubecl-runtime/src/server.rs
```diff
@@ -143,3 +142,29 @@ impl<Server: ComputeServer> Clone for Binding<Server> {
         }
     }
 }
+
+/// Provides launch information specifying the number of work groups to be used by a compute shader.
+pub enum CubeCount {
+    /// Dispatch x,y,z work groups.
+    Static(u32, u32, u32),
+    /// Dispatch work groups based on the values in this buffer. The buffer should contain a u32 array [x, y, z].
```
nathanielsimard, line 150: I would avoid using work group here, since it's a different name for cuda. We could provide a list of equivalent terms in different languages (grid for cuda, num_workgroup for wgsl, etc.).

crates/cubecl-runtime/tests/dummy/server.rs
```diff
@@ -16,32 +17,27 @@ use super::DummyKernel;
 /// The dummy server is used to test the cubecl-runtime infrastructure.
 /// It uses simple memory management with a bytes storage on CPU, without asynchronous tasks.
 #[derive(new, Debug)]
-pub struct DummyServer<MM = SimpleMemoryManagement<BytesStorage>> {
+pub struct DummyServer<MM = MemoryManagement<BytesStorage>> {
     memory_management: MM,
 }
```
nathanielsimard, line 22: We could simply remove the generic parameter here.

crates/cubecl-runtime/build.rs
```diff
@@ -4,5 +4,6 @@ fn main() {
     // Setup cfg aliases
     cfg_aliases! {
         autotune_persistent_cache: { all(feature = "std", any(target_os = "windows", target_os = "linux", target_os = "macos")) },
+        exclusive_memory_only: { any(feature = "exclusive-memory-only", target_family = "wasm") },
```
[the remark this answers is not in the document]
ArthurBrussee: on wasm, exclusive memory is not a preference, but just broken if you don't use it, so better to compile it away imo.

Otherwise, maybe could also configure it with an environment variable in RuntimeOptions, like tasks_max is done, as well, but to be able to compile a cube shader you need to know whether all usages of it are spec compliant, so changing it later on doesn't really work. Kernels are re-used between clients(?) so if two are configured differently they now can't share a kernel.

crates/cubecl-runtime/src/server.rs
ArthurBrussee: Ah this was just copied, but yeah fixed up the docs a bit. This is user facing so used the 'cube' terminology and explained it a bit.

crates/cubecl-runtime/tests/dummy/server.rs
ArthurBrussee: Nice missed that one

PR #163, opened 2024-10-10
Add block merging and fix break parsing

wingertge:
Some fixes for the optimizer around parsing break, along with a block merging pass that merges `BasicBlocks` left over from constant branches that were removed.

All tests pass in `cubecl` several more tests pass in `burn`.


nathanielsimard (approved):
LGTM

PR #164, opened 2024-10-10
Add a WgpuPoll loop

ArthurBrussee:
Currently reading back a tensor on wgpu requires doing a full wgpu::maintain::Wait. Ideally something in the background just polls, and the read will resolve when it's done. In the browser, the event loop handles polling.

This PR adds a WgpuPoll which handles polling in the background. It uses Arcs to hand out as handles to check whether something still needs polling and keeps polling in a busy loop until everything is done.

I had hoped this would improve performance overall but it doesn't seem like it really does, however, when doing some multi threaded things it does mean the other thread can make progress instead of also being stalled, which does help a bit in my case (one thread doing inference, one doing training).

This code seems to deadlock when not done juuust right. I've run all cube tests on a loop and it seems ok now but well.

### DIFF
--- a/crates/cubecl-wgpu/src/compute/server.rs
+++ b/crates/cubecl-wgpu/src/compute/server.rs
@@ -16,6 +16,72 @@ use cubecl_runtime::{
 use hashbrown::HashMap;
 use wgpu::{CommandEncoder, ComputePass, ComputePipeline, ShaderModuleDescriptor};
 
+#[cfg(not(target_family = "wasm"))]
+mod poll {
+    #[derive(Debug)]
+    pub struct WgpuPoll {
+        active_handle: std::sync::Arc<()>,
+        cancel_sender: std::sync::mpsc::Sender<()>,
+    }
+
+    impl WgpuPoll {
+        pub fn new(device: std::sync::Arc<wgpu::Device>) -> Self {
+            let active_handle = std::sync::Arc::new(());
+            let thread_check = active_handle.clone();
+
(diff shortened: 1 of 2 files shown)


crates/cubecl-wgpu/src/compute/server.rs
```diff
@@ -16,6 +16,72 @@ use cubecl_runtime::{
...
+                if std::sync::Arc::strong_count(&thread_check) > 2 {
+                    device.poll(wgpu::MaintainBase::Poll);
+                } else {
+                    // Do not cancel thread while someone still needs to poll.
+                    if cancel_receiver.try_recv().is_ok() {
+                        break;
+                    }
+                }
+                std::thread::yield_now();
+            });
```
nathanielsimard, line 45: Not sure we should always spawn a new thread, could we actually have a pooling thread instead owned by the server?

ArthurBrussee: Well the server owns the WgpuPoll, there's one thread per server, not sure what you mean? 

Could have the spawn in the server and move it into WgpuPoll but idk, seperate type seems to work it to not copy more code to the spir-v server.

One thing I probably should do is park/unpark the thread when needed so it's not looping the entire time.

nathanielsimard (approved):

crates/cubecl-wgpu/src/compute/server.rs
nathanielsimard: I was confused, got it now thanks 😅

PR #167, opened 2024-10-12
Support autotuning on WebGPU, more precise timings on wgpu

ArthurBrussee:
This PR fixes autotuning when running on WebGPU. Previously, autotuning tried to measure time by syncing the client, doing some operations, and syncing again. On WebGPU, however, device.poll(wgpu::maintain::Wait) is just ignored, so the timings were wrong. This can lead to quite bad performance - e.g., selecting a naive reduction for getting the mean of 250K elements in my case!

Fixing this required two main changes:

1. Make client.sync() async. Have the autotuner kick off a future and wait for results to come in.
2. That almost works but turns out to be quite imprecise. After an async wait, the browser can schedule some other work, which gets included in the time measurements. This is also a problem on native when using threads. Instead, use time queries to measure the actual GPU execution time. This should be a good deal more precise as well.

This required some reworks:

- Autotune cache now has a "pending" state. On native, however, I've decided not to thread the autotune results yet. Doing so could be better for startup times, but it means benchmarks would now have a "warmup" effect, and multi-threaded tuning would be less accurate, especially if you're not measuring GPU time like on CUDA.
- client.sync() is now async. Flushing really doesn't need to be, so that is a separate function now. I think for Burn we should remove SyncType entirely and have people use the client if they want to flush. It doesn't make sense for all backends. Similarly, I think Burn's Backend::sync() can stay synchronous and just warn on WASM to drop down to the lower-level client API if needed.
- Not strictly needed, but removed the boxing of the futures. Turns out it's not needed, as long as you mark the future as 'static.
- The autotuner doesn't use the Benchmark trait anymore. The Benchmark trait could be changed to have an async option as well, and on wgpu could use the real GPU time to be more accurate, but I didn't want to mess with existing benchmarks to keep things comparable.

With that, timings now work on WebGPU and are generally a bit more precise / have less variance now when using wgpu.

For the future, I also wonder if it would make sense to make a "benchmark" function with a limited scope. That would mean you could be sure no other GPU work is introduced while benchmarking.


crates/cubecl-runtime/src/tune/tuner.rs
```diff
@@ -1,16 +1,16 @@
-#[cfg(target_family = "wasm")]
-use web_time::Duration;
+use async_channel::{Receiver, Sender};
 
-#[cfg(not(target_family = "wasm"))]
-use core::time::Duration;
 use core::{any::Any, mem::ManuallyDrop};
-#[cfg(feature = "std")]
-use std::panic::{catch_unwind, resume_unwind, AssertUnwindSafe};
+use web_time::Duration;
```
nathanielsimard, line 4: Do we always want to use duration from `web_time` ? Maybe we should put those things in `cubecl_common` ?

crates/cubecl-runtime/src/tune/tuner.rs
```diff
@@ -47,115 +52,146 @@ impl<K: AutotuneKey> Tuner<K> {
...
+        while let Ok((key, fastest_index)) = self.results_rec.try_recv() {
+            self.tune_cache.cache_insert(key.clone(), fastest_index);
+            #[cfg(autotune_persistent_cache)]
+            {
+                let checksum = autotune_operation_set.compute_checksum();
+                self.tune_cache
+                    .persistent_cache_insert(key, checksum, fastest_index);
+                self.tune_cache.save();
+            }
+        }
```
nathanielsimard, line 85: What is the difference between using the channel and having a `Hit` ?

crates/cubecl-runtime/src/tune/tuner.rs
```diff
@@ -47,115 +52,146 @@ impl<K: AutotuneKey> Tuner<K> {
...
-                    self.run_benchmark(op, client)
-                } else {
-                    Ok(BenchmarkDurations::new(Vec::from([Duration::MAX])))
+        self.tune_cache.mark_pending(key.clone());
+
+        let set = autotune_operation_set.clone();
+        let client = client.clone();
+        let sender = self.results_send.clone();
+
+        let fut = async move {
```
nathanielsimard, line 111: I think the async block could become another method, it's starting to become a long 😅

crates/cubecl-runtime/src/tune/tuner.rs
```diff
@@ -47,115 +52,146 @@ impl<K: AutotuneKey> Tuner<K> {
...
-                fastest_tunable = Some(i);
-            }
+        #[cfg(all(not(target_family = "wasm"), feature = "std"))]
+        {
+            use futures_lite::FutureExt;
+            AssertUnwindSafe(tuner.sample_durations())
+                .catch_unwind()
+                .await
+                .map_err(|e| {
+                    println!("Caught error while benchmarking, falling back to next operation.");
```
nathanielsimard, line 192: It should be a log instead.

crates/cubecl-runtime/tests/integration_test.rs
```diff
@@ -250,8 +250,8 @@ fn autotune_cache_different_keys_return_a_cache_miss() {
         dummy::CacheTestAutotuneOperationSet::new(client.clone(), shapes_1, handles_1);
     let cache_test_autotune_kernel_2 =
         dummy::CacheTestAutotuneOperationSet::new(client.clone(), shapes_2, handles_2);
-    autotune_execute(&client, Box::new(cache_test_autotune_kernel_1));
-    autotune_execute(&client, Box::new(cache_test_autotune_kernel_2));
+    autotune_execute(&client, Arc::new(cache_test_autotune_kernel_1));
```
nathanielsimard, line 253: The reason to use a `Box` was so that tensors stored in the autotune operation set have their ownership intact to allow for inplace operations. But I don't think it matters to compute benchmarks, and in fact probably contributed to more variance (we still cloned the tensors because we run 10 samples).

crates/cubecl-wgpu-spirv/Cargo.toml
```diff
@@ -23,13 +23,13 @@ cubecl-spirv = { path = "../cubecl-spirv", version = "0.2.0" }
...
 
 bytemuck = { workspace = true }
-pollster = { workspace = true }
 wgpu = { version = "22.0.0", features = [
     "fragile-send-sync-non-atomic-wasm",
     "spirv",
 ] }
 
 async-channel = { workspace = true }
+futures-lite = { workspace = true }
```
nathanielsimard, line 32: What is the difference between `pollster` and `futures-lite`?

crates/cubecl-wgpu/src/compute/server.rs
```diff
@@ -132,6 +138,12 @@ impl WgpuServer {
             tasks_max,
             logger: DebugLogger::new(),
             poll: WgpuPoll::new(device.clone()),
+            query_set: device.create_query_set(&QuerySetDescriptor {
+                label: Some("burn queries"),
```
nathanielsimard, line 142: burn => cubecl ;)

crates/cubecl-wgpu/src/compute/server.rs
```diff
@@ -371,12 +381,22 @@ impl ComputeServer for WgpuServer {
...
+            let pass = self
+                .encoder
                 .begin_compute_pass(&wgpu::ComputePassDescriptor {
                     label: None,
-                    timestamp_writes: None,
+                    timestamp_writes: Some(wgpu::ComputePassTimestampWrites {
+                        query_set: &self.query_set,
+                        beginning_of_pass_write_index: start_write,
+                        end_of_pass_write_index: end_write,
+                    }),
```
nathanielsimard, line 396: Do we always want to keep track of timestamps? Does it slow down compute?

crates/cubecl-runtime/src/tune/tuner.rs
ArthurBrussee: Ah yes - web_time just compiles to normal time on non-wasm platforms and strips away all it's code, so besides a tiny compilation overhead there's no reason not to, so don't think moving it helps much either, but lmk!

crates/cubecl-runtime/src/tune/tuner.rs
ArthurBrussee: Hm the code before the async block isn't doing much except cloning for the captures in the block. I've tried instead to shorten the async block a bit & move out some stuff in it.

crates/cubecl-runtime/src/tune/tuner.rs
ArthurBrussee: Ah yes - made it a log::warn because panic unwinds aren't supported on wasm/no-std so shouldn't rely on them too much.

crates/cubecl-runtime/tests/integration_test.rs
ArthurBrussee: Hmm I see yeah, tricky. Don't think there's a good way around this as long as the set needs to be shared with another thread (/ future). I guess in the future it'd be possible to run the benchmark on something like Tensor::empty_like(a) if the tensor seems to be possible to in-place otherwise.

crates/cubecl-wgpu-spirv/Cargo.toml
ArthurBrussee: Ah yes forgot this in the PR description - we now need futures-lite for some extension methods anyway (mainly the futures panic unwinding), and it has a block_on method much like pollster. Think it's a bit more actively supported and don't need two deps.

crates/cubecl-wgpu/src/compute/server.rs
ArthurBrussee: Lol woops

crates/cubecl-wgpu/src/compute/server.rs
ArthurBrussee: Yes :/ It's a bit unfortunate - see my comment about maybe needing a special benchark function that runs kernels in a scope, would be a bit cleaner.

That said, as far as I can measure there's no overhead, it only requires writing a 8 byte timestamp each tasks_mask kernels so not a lot of work.

Issue #171, opened 2024-10-15
Mac: Features(TIMESTAMP_QUERY) are required but not enabled on the device


utensil:
On Github Actions runner `macos-latest`, a simple gelu example using CubeCL have just become resulting in the following error (and it was fine around 5 hours ago):

```
thread 'test_shader' panicked at /Users/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/wgpu-22.1.0/src/backend/wgpu_core.rs:3411:5:
wgpu error: Validation Error

Caused by:
  In Device::create_query_set
    Features Features(TIMESTAMP_QUERY) are required but not enabled on the device
```

<details>
<summary>stack backtrace:</summary>

```
   0: rust_begin_unwind
             at /rustc/9322d183f45e0fd5a509820874cc5ff27744a479/library/std/src/panicking.rs:665:5
   1: core::panicking::panic_fmt
             at /rustc/9322d183f45e0fd5a509820874cc5ff27744a479/library/core/src/panicking.rs:74:14
   2: wgpu::backend::wgpu_core::default_error_handler
             at /Users/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/wgpu-22.1.0/src/backend/wgpu_core.rs:3411:5
   3: core::ops::function::Fn::call
             at /rustc/9322d183f45e0fd5a509820874cc5ff27744a479/library/core/src/ops/function.rs:79:5
   4: <alloc::boxed::Box<F,A> as core::ops::function::Fn<Args>>::call
             at /rustc/9322d183f45e0fd5a509820874cc5ff27744a479/library/alloc/src/boxed.rs:2468:9
   5: wgpu::backend::wgpu_core::ErrorSinkRaw::handle_error
             at /Users/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/wgpu-22.1.0/src/backend/wgpu_core.rs:3397:17
   6: wgpu::backend::wgpu_core::ContextWgpuCore::handle_error
             at /Users/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/wgpu-22.1.0/src/backend/wgpu_core.rs:296:9
   7: wgpu::backend::wgpu_core::ContextWgpuCore::handle_error_nolabel
             at /Users/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/wgpu-22.1.0/src/backend/wgpu_core.rs:308:9
[...]

PR #174, opened 2024-10-15
Fix for wgpu backend for machines without timstamp queries.

ArthurBrussee:
Fixes https://github.com/tracel-ai/cubecl/issues/171

This uses a dummy readback, instead of timestamps, for platforms that don't support it. I'm not 100% sure how many machines out there this really is - anything past Big Sur should in theory support this but I guess CI doesn't, maybe some metal weirdness.

Also simplifies the time bookkeeping a bit.

### DIFF
--- a/crates/cubecl-wgpu/src/compute/server.rs
+++ b/crates/cubecl-wgpu/src/compute/server.rs
@@ -1,4 +1,9 @@
-use std::{future::Future, num::NonZero, time::Duration};
+use std::{
+    future::Future,
+    num::NonZero,
+    pin::Pin,
+    time::{Duration, Instant},
+};
 
 use crate::compiler::wgsl::WgslCompiler;
 
@@ -109,8 +114,8 @@ pub struct WgpuServer {
     storage_locked: MemoryLock,
 
     duration_profiled: Duration,
-    query_set: QuerySet,
-    query_started: bool,
+    command_start_time: Option<Instant>,
+    query_set: Option<QuerySet>,
 }
 
 fn create_encoder(device: &wgpu::Device) -> CommandEncoder {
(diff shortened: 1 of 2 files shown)


crates/cubecl-wgpu/src/compute/server.rs
```diff
@@ -267,30 +283,30 @@ impl WgpuServer {
...
+            // is not yet implemented on wgpu https://github.com/gfx-rs/wgpu/issues/6395
+            //
+            // For now, instead do a dummy readback. This *seems* to wait for the entire
+            // queue to be done.
+            let dummy = self.empty(32);
+            let fut = self.read(dummy.binding());
+            Box::pin(async move {
+                fut.await;
+                start_time.elapsed()
+            })
```
nathanielsimard, line 309: Could we simply `poll::wait` the device here instead?

ArthurBrussee: Maybe but then this path doesn't for WebGPU again, and the timestamp queries extension is technically optional, so you could hit this.

on_submitted_work_done is the real fix, hope wgpu has it soon.

crates/cubecl-wgpu/src/compute/server.rs
```diff
@@ -127,6 +132,16 @@ impl WgpuServer {
...
+        let queries = if device.features().contains(wgpu::Features::TIMESTAMP_QUERY) {
+            Some(device.create_query_set(&QuerySetDescriptor {
+                label: Some("CubeCL profile queries"),
+                ty: QueryType::Timestamp,
+                count: 2,
+            }))
+        } else {
+            None
+        };
+
```
utensil, line 144: A quick feedback, the fix in this PR now works on Github Actions, locally on M1 too. Thanks!

PR #176, opened 2024-10-16
Fix no-std + wasm infinite loop

nathanielsimard:
### Fix wasm infinite loop

Reading the documentation, it's not possible to block on a future in `wasm`, this is why we had `read_sync` and `try_read_sync`. However the `block_on` function provided by embassy is essentially an infinite loop, which would never finish on `wasm`. I simply reuse the `read_sync` function instead.

### Fix no-std

The `Waker` trait used in `read_sync` isn't the the same when real atomics are available, so I simply fixed that. I'm unsure if we should expose `read_sync` and `try_read_sync` anymore since we can use `block_on` instead.

We could expose `try_block_on` which would call `catch_unwind` in `std` environments and `try_read_sync` in `wasm` environments. But I'm unsure yet if it's a good idea.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -66,9 +66,6 @@ strum = { version = "0.26.3", features = ["derive"] }
 tracel-xtask = { version = "~1.0" }
 
 portable-atomic = { version = "1.9" }
-portable-atomic-util = { version = "0.2.2", features = [
-    "alloc",
-] } # alloc is for no_std
 pretty_assertions = "1.4"
 
 # Async
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -65,6 +65,7 @@ syn = { version = "2", features = ["full", "extra-traits", "visit-mut"] }
 strum = { version = "0.26.3", features = ["derive"] }
 tracel-xtask = { version = "~1.0" }
 
+portable-atomic = { version = "1.9" }
 portable-atomic-util = { version = "0.2.2", features = [
     "alloc",
 ] } # alloc is for no_std
--- a/crates/cubecl-common/src/future.rs
+++ b/crates/cubecl-common/src/future.rs
@@ -3,12 +3,17 @@ use core::future::Future;
 
 /// Block until the [future](Future) is completed and returns the result.
 pub fn block_on<O>(fut: impl Future<Output = O>) -> O {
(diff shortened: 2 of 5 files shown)


crates/cubecl-common/src/future.rs
```diff
@@ -3,12 +3,17 @@ use core::future::Future;
...
+
+    #[cfg(all(not(target_family = "wasm"), not(feature = "std")))]
     {
         embassy_futures::block_on(fut)
     }
 
-    #[cfg(feature = "std")]
+    #[cfg(all(not(target_family = "wasm"), feature = "std"))]
     {
         futures_lite::future::block_on(fut)
```
ArthurBrussee, line 18: Maybe worth switching all the block_ons to embassy if we're using it

Cargo.toml
```diff
@@ -65,9 +65,7 @@ syn = { version = "2", features = ["full", "extra-traits", "visit-mut"] }
 strum = { version = "0.26.3", features = ["derive"] }
 tracel-xtask = { version = "~1.0" }
 
-portable-atomic-util = { version = "0.2.2", features = [
-    "alloc",
-] } # alloc is for no_std
+portable-atomic = { version = "1.9" }
```
ArthurBrussee, line 68: Don't think you need this at all anymore?

crates/cubecl-common/src/future.rs
nathanielsimard: When `std`, we're using `futures-lite` instead.

Cargo.toml
nathanielsimard: We kind of do, because we're activating the `portable-atomic` feature of `spin` and it fails the burn CI if we don't activate the right feature flag on `portable-atomic` when native atomic aren't available... yeah I know it's painful 😅

crates/cubecl-common/src/future.rs
ArthurBrussee: Ok but... why :D Seems better to keep one async runtime, even if the runtime is just all block_on's. It's not a big difference though.

nathanielsimard: We can't catch unwind with embassy, which is great for autotune. Also we don't pull the runtime for embassy, it's just a loop with a busy waiting.

PR #177, opened 2024-10-16
Fix deadlock

ArthurBrussee:
Fix deadlock when running sync().

This was a bad bug after #174 

The intention was to write timestamp 0 and 1 for the first compute pass, and only 1 for every pass afterwards.

Instead, because we immediatly set the start time we only ever write timestamp 1, which combined with the following problem https://github.com/gfx-rs/wgpu/issues/3612 meant a deadlock.

Also update the wgpu-spirv server to this version, and move the profiling code so it really is just at the start/end of execute instead of interwoven.

Lastly, use web_time::Instant. This isn't available in std but wgpu doesnt' work on no_std anyway. Would be nice to add to cubecl_common but seems there's no good core version.


nathanielsimard (approved):
Yup it fixes the problem thank you so much 🙏

PR #181, opened 2024-10-17
Add prefetching to into_contiguous

wingertge:
Adds prefetching to the `into_contiguous` kernel, with a heuristic to find the number of elements per unit by default and the option to explicitly pass the number. This improves performance by up to 2x.

# Performance
There's currently no dedicated benchmark for `into_contiguous`, but the time for executing both `into_contiguous` kernels for `implicit_gemm` went from ~1.25-1.3ms with no prefetching to ~700-750µs with prefetch 8 for the large (16x16x512x512) input tensor, and prefetch 1 for the much smaller (64x16x3x3) weight tensor.
The Nsight profiler shows extreme memory latency stalls with no prefetch (86% of stall reasons, and it's the second most frequent stall overall), with prefetch the stalls are mostly balanced and all the most common stalls are now in the GEMM kernel.

### DIFF
--- a/crates/cubecl-linalg/src/tensor/contiguous.rs
+++ b/crates/cubecl-linalg/src/tensor/contiguous.rs
@@ -29,29 +29,70 @@ fn into_contiguous_kernel<N: CubePrimitive>(
     input: &Tensor<Line<N>>,
     output: &mut Tensor<Line<N>>,
     #[comptime] rank: Option<u32>,
+    #[comptime] elems_per_thread: u32,
 ) {
-    let offset_output = ABSOLUTE_POS;
+    let offset_output = ABSOLUTE_POS * elems_per_thread;
+    let vec = vectorization_of(input);
 
     if offset_output >= output.len() {
         return;
     }
 
-    let offset_input = index_offset_with_layout::<N, N>(
-        input,
-        output,
-        offset_output,
-        0,
-        rank.unwrap_or_else(|| output.rank()),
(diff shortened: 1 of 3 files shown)


crates/cubecl-linalg/src/tensor/contiguous.rs
```diff
@@ -61,11 +74,41 @@ pub fn into_contiguous<R: Runtime, E: CubePrimitive>(
...
+
+    // TODO: Benchmark to find good default prefetch, for now preserve existing behaviour
+    into_contiguous_prefetch(client, input, elems_per_unit)
+}
+
+/// Make a jit tensor contiguous.
+pub fn into_contiguous_prefetch<R: Runtime, E: CubePrimitive>(
+    client: &ComputeClient<R::Server, R::Channel>,
+    input: TensorHandleRef<'_, R>,
+    elems_per_unit: u32,
```
nathanielsimard, line 95: The vectorization factor could be passed as input, we're already calculating it in the `into_contiguous` function. Also is there a reason why this function is public?

wingertge: To make it possible to override the `elems_per_unit` based on autotuning or other heuristics that may more more specific to that use case. That's why I don't want to pass the vectorization: the function is supposed to be simple to use from things like `burn` so it can be tuned (auto or not) easily.

nathanielsimard (approved):

PR #184, opened 2024-10-18
Merge `cubecl-wgpu` and `cubecl-wgpu-spirv`

wingertge:
Merge the two `wgpu` backends into one using a new `WgpuCompiler` trait that handles compilation (specifically for the mode issues around `robustness2`), creating the device (for custom features and extensions), registering supported features and creating the pipeline.


crates/cubecl/Cargo.toml
```diff
@@ -19,26 +19,25 @@ default = [
...
 ]
+linalg = ["dep:cubecl-linalg"]
 std = ["cubecl-core/std", "cubecl-wgpu?/std", "cubecl-cuda?/std"]
 template = ["cubecl-core/template"]
 
 # Runtimes
 cuda = ["cubecl-cuda"]
 wgpu = ["cubecl-wgpu"]
-wgpu-spirv = ["cubecl-wgpu-spirv"]
+wgpu-spirv = ["cubecl-wgpu/spirv"]
```
nathanielsimard, line 33: I would put `["wgpu", "cubecl-wgpu/spirv"]` so that the wgpu feature flag is activated.

crates/cubecl-wgpu/src/compiler/base.rs
```diff
@@ -0,0 +1,27 @@
...
+
+    fn create_pipeline(
+        server: &mut WgpuServer<Self>,
+        kernel: CompiledKernel<Self>,
+        mode: ExecutionMode,
+    ) -> Arc<ComputePipeline>;
+
+    #[expect(async_fn_in_trait, reason = "Only used locally")]
+    async fn request_device(adapter: &Adapter) -> (Device, Queue);
+    fn register_features(adapter: &Adapter, device: &Device, props: &mut DeviceProperties<Feature>);
```
nathanielsimard, line 26: I'm not familiar with `#[expect]`, what does it do? Also, would it be better to have only one function `request_device` that comes with all features registered?

```rust
async fn request_device(adapter: &Adapter) -> (Device, Queue, DeviceProperties<Features>);
```

wingertge: `expect` was added a few rust versions ago, I'll have to actually check if it fits with the MSRV. Might just do `allow` for now. It's `allow` but warns you if the lint is not actually needed, for example because things change and it becomes unnecessary to have it there.

wingertge: Looks like it's 1.81 MSRV, but burn is currently set to 1.80 for its minimum version. Same goes for the lint reason which was added at the same time. Should I make it an `allow` with no reason for now?

wingertge: As for merging the functions, it's not possible without changing the inialization order `wgpu` uses right now. `request_device` is used in `select_device`, but the features are set on the `Client`, not the device.

nathanielsimard (approved):

PR #186, opened 2024-10-19
Feat/cube impl

nathanielsimard:
Now we support adding methods to structs!


crates/cubecl-macros/src/parse/cube_impl.rs
```diff
@@ -0,0 +1,237 @@
...
+                    vec![CubeImplItem::Fn(func), CubeImplItem::FnExpand(func_expand)]
+                }
+            }
+            _ => vec![CubeImplItem::Other],
+        };
+
+        Ok(res)
+    }
+
+    pub fn func(&mut self) -> Option<&mut KernelFn> {
```
wingertge, line 65: These should probably be `as_func` etc to keep with Rust conventions.

wingertge:
Looks pretty good overall, I really like the vec of enums to more easily handle sequences of items in codegen. Makes things much more modular and reusable.

PR #188, opened 2024-10-20
Fix casting, powf and alignment for SPIR-V

wingertge:
Fix bugs in vectorized `powf` and casting to `bool`.
Quick fix for alignment crash.


crates/cubecl-wgpu/src/compute/storage.rs
```diff
@@ -73,6 +73,12 @@ impl WgpuStorage {
 impl ComputeStorage for WgpuStorage {
     type Resource = WgpuResource;
 
+    // 32 bytes is enough to handle a double4 worth of alignment.
+    // See: https://github.com/gfx-rs/wgpu/issues/3508
+    // NB: cudamalloc and co. actually align to _256_ bytes. Worth
+    // trying this in the future to see if it reduces memory coalescing.
+    const ALIGNMENT: usize = 32;
```
nathanielsimard, line 80: Is it wise to make the alignment a constant here? The alignment might depend on runtime device properties, and requiring it to be known at compile time could cause some issues.

wingertge: I actually don't know how to fetch it for HIP, and I know for `wgpu` it is a constant in the API. I don't believe this is device dependent.

nathanielsimard: ok, well let's start with that and we can always change it later if it causes problems.

ArthurBrussee: Ah for wgpu when the MemoryDeviceProperties is created it does do a max of storage::ALIGNMENT and what wgpu reports alignment needs to be, I guess this is more like "minimal alignment". 

I added it because the padding in queue.write_buffer_with relies on alignment being >= 4 and alignment % 4 == 0, so eaiser to see that's the case this way.

I also think in the future we might want to overalign to 256 bytes - that's what CUDA does by default appearantly claiming it helps performance (memory coalescing), though this seems less relevant on latest gen hardware.

PR #190, opened 2024-10-21
Allow users to specify timing method (cpu/gpu) to use when running a benchmark

jbelanich:
Changes to benchmark.rs stemming from [this conversation](https://discord.com/channels/1038839012602941528/1286391740739358802/1296921922432663663). 

**The problem:** I'm seeing the time required to read timestamps from the GPU completely wash out benchmark results for an FFT kernel I'm writing (which typically has timings in the 10-100 microseconds). Ideally I'd like to be able to have my benchmarks report just those GPU timestamps, rather than the CPU timing of retrieving those GPU timestamps. However, the current implementation of the `Benchmark` trait doesn't allow this.

**Solution:** Since both CPU and GPU timings are useful under different contexts, this change allows the caller of `Benchmark::run` to decide which timing method to use by providing a `TimingMethod` argument. A side-benefit to this is that it is now more explicit how `BenchmarkDurations` is computed. To maintain previous behavior, the matmul and unary benchmarks are set to use `TimingMethod::CPU`.

### DIFF
--- a/crates/cubecl-common/src/benchmark.rs
+++ b/crates/cubecl-common/src/benchmark.rs
@@ -7,9 +7,32 @@ use serde::{Deserialize, Serialize};
 
 use super::stub::Duration;
 
+/// How a benchmark's execution times are measured.
+#[derive(Debug, Default, Clone, Copy, Serialize, Deserialize)]
+pub enum TimingMethod {
+    /// Time measurements come from CPU timing of execution + sync
+    /// calls.
+    #[default]
+    CPU,
+    /// Time measurements come from hardware reported timestamps
+    /// coming from a sync call.
+    GPU,
+}
+
+impl Display for TimingMethod {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        match self {
+            TimingMethod::CPU => f.write_str("cpu"),
+            TimingMethod::GPU => f.write_str("gpu"),
+        }
+    }
+}
+
 /// Results of a benchmark run.
 #[derive(new, Debug, Default, Clone, Serialize, Deserialize)]
 pub struct BenchmarkDurations {
+    /// How these durations were measured.
+    pub timing_method: TimingMethod,
     /// All durations of the run, in the order they were benchmarked
     pub durations: Vec<Duration>,
 }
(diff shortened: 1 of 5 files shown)


crates/cubecl-common/src/benchmark.rs
```diff
@@ -7,9 +7,32 @@ use serde::{Deserialize, Serialize};
...
+/// How a benchmark's execution times are measured.
+#[derive(Debug, Default, Clone, Copy, Serialize, Deserialize)]
+pub enum TimingMethod {
+    /// Time measurements come from CPU timing of execution + sync
+    /// calls.
+    #[default]
+    CPU,
+    /// Time measurements come from hardware reported timestamps
+    /// coming from a sync call.
+    GPU,
```
nathanielsimard, line 19: I think CPU & GPU is a bit confusing, since the execution isn't on the CPU. I would rename those based on the scope, maybe `Full` and `Kernel` or `Total` and `Device`. I'm open to better names 😅. Kernel is kinda bad, since we may benchmark multiple kernels at the same time.

nathanielsimard: Also, we could have two `sync` methods, one that reads the timestamps and returns a duration, and another sync that doesn't read the timestamps. Maybe `sync_profiled() -> Result<Duration, SyncProfiledError>` and `sync() `. At some point we might have the option to start collecting timestamps, and calling `sync_profiled` would return an error, no timestamps recorded. So maybe instead of `sync_profiled` is could be `read_timestamps`, unsure ...

jbelanich: > I think CPU & GPU is a bit confusing, since the execution isn't on the CPU. I would rename those based on the scope, maybe `Full` and `Kernel` or `Total` and `Device`. I'm open to better names 😅. Kernel is kinda bad, since we may benchmark multiple kernels at the same time.

Got it, makes sense. How do you feel about DeviceOnly and Full? DeviceOnly implies there is more to benchmark, and Full implies...well, everything. If Full is too vague, maybe DeviceAndOverhead to be more concrete?

jbelanich: > Also, we could have two `sync` methods, one that reads the timestamps and returns a duration, and another sync that doesn't read the timestamps. Maybe `sync_profiled() -> Result<Duration, SyncProfiledError>` and `sync() `. At some point we might have the option to start collecting timestamps, and calling `sync_profiled` would return an error, no timestamps recorded. So maybe instead of `sync_profiled` is could be `read_timestamps`, unsure ...

Yeah, having the option to sync without profiling would be nice. It is a little strange that right now sync is forced to also profile, and must incur a profiling cost 😅. Especially since on wgpu it seems that such overhead is non-trivial.

nathanielsimard (approved):

PR #193, opened 2024-10-22
Add features for each supported datatype

wingertge:
Adds a feature for each `Elem` type supported by the backend. Preparation for adding 8 and 16 bit integer types.
Adds a page to the book describing optional hardware/backend features and how to query their availability.
Also fixes the features set by SPIR-V, it was missing a feature required for accessing 16 bit types from storage buffers.

### DIFF
--- a/crates/cubecl-wgpu/src/compiler/spirv.rs
+++ b/crates/cubecl-wgpu/src/compiler/spirv.rs
@@ -4,9 +4,9 @@ use ash::{
     khr::cooperative_matrix,
     vk::{
         ComponentTypeKHR, DeviceCreateInfo, DeviceQueueCreateInfo,
-        PhysicalDeviceCooperativeMatrixFeaturesKHR, PhysicalDeviceVulkan12Features,
-        PhysicalDeviceVulkanMemoryModelFeatures, ScopeKHR, EXT_ROBUSTNESS2_NAME,
-        KHR_COOPERATIVE_MATRIX_NAME,
+        PhysicalDevice16BitStorageFeatures, PhysicalDeviceCooperativeMatrixFeaturesKHR,
+        PhysicalDeviceShaderFloat16Int8Features, PhysicalDeviceVulkanMemoryModelFeatures, ScopeKHR,
+        EXT_ROBUSTNESS2_NAME, KHR_COOPERATIVE_MATRIX_NAME,
     },
 };
 use cubecl_core::{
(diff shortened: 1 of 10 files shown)


crates/cubecl-wgpu/src/compiler/spirv.rs
```diff
@@ -178,9 +179,12 @@ impl WgpuCompiler for SpirvCompiler<GLCompute> {
 fn request_device(
     wgpu_adapter: &wgpu::Adapter,
     adapter: &vulkan::Adapter,
-    features: Features,
+    mut features: Features,
     limits: Limits,
 ) -> (wgpu::Device, wgpu::Queue) {
+    // This registers only f16 but not u8/i8, so remove so we can manually add them
+    features.remove(Features::SHADER_F16);
```
nathanielsimard, line 186: I think we should check if it is present before adding it to the device properties no? It is hardcoded as available in the `register_types` function, but it might not always be available?

crates/cubecl-wgpu/src/compiler/spirv.rs
```diff
@@ -344,7 +380,7 @@ fn dump_spirv(compiled: &CompiledKernel<VkSpirvCompiler>, name: &str, id: cubecl
         hash::{DefaultHasher, Hash, Hasher},
     };
 
-    if let Ok(dir) = std::env::var("CUBCEL_DEBUG_SPIRV") {
+    if let Ok(dir) = std::env::var("CUBECL_DEBUG_SPIRV") {
```
nathanielsimard, line 383: I feel like at some point it would be very nice to have a revamp `DebugLogger` where we would save stuff in a database file as well. Just an idea.

cubecl-book/src/core-features/features.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Overview
+
+### Features
+
+Also requires device support
+
+| Feature | CUDA | ROCm | WGPU (WGSL) | WGPU (SPIR-V) |
+| ------- | ---- | ---- | ----------- | ------------- |
+| Subcube | ✔️   | ✔️   | ✔️          | ✔️            |
```
nathanielsimard, line 20: There is a format error in the table.

nathanielsimard (changes requested):

crates/cubecl-wgpu/src/compiler/spirv.rs
wingertge: All the datatype features in Vulkan are mandatory as far as I can tell. I don't believe they're device specific, they're handled by the driver.

crates/cubecl-wgpu/src/compiler/spirv.rs
wingertge: Yeah I was thinking about this and couldn't come up with a great way to handle it, so this was my temporary fix

cubecl-book/src/core-features/features.md
wingertge: That's actually technically correct, it formats based on character count, but emojis are displayed as wider than normal characters. So the formatter automatically changes it to this.

nathanielsimard (approved):

PR #194, opened 2024-10-22
Fix an overflow when max page == 4gb on wasm

ArthurBrussee:
Running some wasm code on mac hit a weird error. This turned out to be due to usize, 32 bit on wasm, overflowing as max_page_size was 4gb, which when rounded up overflows.

To fix this:
- For slices allocator, round down max page size bucket.
- For exclusive allocator, don't make a bucket of max size at all. There'll be a bucket close to it already, it's not really needed.
- Convert memory addresses to u64 instead of usize. This matches what wgpu  does - GPU adresses don't have to fit in CPU addressable space in the first place, and eases concerns about other overflows - it's a bit too scary to have some usize::MAX floating around when doing math on them! I've left client.empty() as it's more public API, and the API does slightly overpromise, as all tensor reading functions can of course still only do a Vec<> with a usize of length.

### DIFF
--- a/crates/cubecl-runtime/src/memory_management/memory_manage.rs
+++ b/crates/cubecl-runtime/src/memory_management/memory_manage.rs
@@ -17,7 +17,7 @@ enum DynamicPool {
 // This guarantees that _for bins in use_, the wasted space is at most 12.5%. So as long
 // as bins have a high use rate this should be fairly efficient. That said, currently slices in
 // bins don't deallocate, so there is a chance more memory than needed is used.
-const EXP_BIN_SIZES: [usize; 200] = [
+const EXP_BIN_SIZES: [u64; 200] = [
     128, 144, 160, 176, 192, 208, 224, 240, 256, 288, 320, 352, 384, 416, 448, 480, 512, 576, 640,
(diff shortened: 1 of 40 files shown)


nathanielsimard (approved):
I wasn't sure when to put `usize` vs when to put `u64`, but I like the question "would a vector could be created from this size?"

crates/cubecl-runtime/src/memory_management/memory_manage.rs
```diff
@@ -116,11 +112,11 @@ impl<Storage: ComputeStorage> MemoryManagement<Storage> {
             MemoryConfiguration::SubSlices => {
                 // Round chunk size to be aligned.
                 let memory_alignment = properties.alignment;
-                let max_page = round_up_to_multiple(properties.max_page_size, memory_alignment);
+                let max_page = properties.max_page_size;
```
AsherJingkongChen, line 115: Doesn't `max_page` need to be aligned?

ArthurBrussee: Doesn't always seem like it, I'm guessing because wgpu clamps to 4gb to align things. In any case would like to allow aligning to higher align than wgpu requires

PR #195, opened 2024-10-22
Refactor/timestamps

nathanielsimard:
The timestamps now work with and without profiling, with better error handling to identify the cause of timestamps not working. We can disable timestamps in all runtimes to reduce overhead (mostly for wgpu). We maintain an API that is mostly compatible with the benchmarks written in Burn, which rely on full execution time, to remain comparable to other backends not based on CubeCL.

### DIFF
--- a/crates/cubecl-common/src/benchmark.rs
+++ b/crates/cubecl-common/src/benchmark.rs
@@ -34,7 +34,7 @@ pub enum TimestampsError {
     /// Collecting timestamps is disabled, make sure to enable it.
     Disabled,
     /// Collecting timestamps isn't available.
-    Unavailabled,
+    Unavailable,
     /// En unknown error occured while collecting timestamps.
     Unknown(String),
 }
@@ -179,7 +179,7 @@ pub trait Benchmark {
     /// Wait for computation to complete and return hardware reported
     /// computation duration.
     fn sync_elapsed(&self) -> TimestampsResult {
-        Err(TimestampsError::Unavailabled)
+        Err(TimestampsError::Unavailable)
     }
 
     /// Run the benchmark a number of times.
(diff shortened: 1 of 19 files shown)


crates/cubecl-common/src/benchmark.rs
```diff
@@ -28,6 +28,20 @@ impl Display for TimingMethod {
     }
 }
 
+/// Error that can occured when collecting timestamps from a device.
+#[derive(Debug)]
+pub enum TimestampsError {
+    /// Collecting timestamps is disabled, make sure to enable it.
+    Disabled,
+    /// Collecting timestamps isn't available.
+    Unavailabled,
```
ArthurBrussee, line 37: Typo -> Unavailable

crates/cubecl-wgpu/src/compute/server.rs
```diff
@@ -230,48 +187,106 @@ impl<C: WgpuCompiler> WgpuServer<C> {
...
-            })
-        } else {
-            // TODO: This should work queue.on_submitted_work_done() but that
-            // is not yet implemented on wgpu https://github.com/gfx-rs/wgpu/issues/6395
-            //
-            // For now, instead do a dummy readback. This *seems* to wait for the entire
-            // queue to be done.
+        #[cfg(target_family = "wasm")]
+        {
             let dummy = self.empty(32);
```
ArthurBrussee, line 195: I'd leave the comment about on_submitted_work_done here

crates/cubecl-wgpu/src/compute/server.rs
```diff
@@ -230,48 +187,106 @@ impl<C: WgpuCompiler> WgpuServer<C> {
...
+                        .map(|x| u64::from_le_bytes(x.try_into().unwrap()))
+                        .collect::<Vec<_>>();
+                    let delta = u64::checked_sub(data[1], data[0]).unwrap_or(1);
+                    let duration = Duration::from_secs_f64(delta as f64 * period);
+
+                    Ok(duration)
+                })
+            }
+            TimestampMethod::StartTime(start_time) => {
+                // TODO: This should work queue.on_submitted_work_done() but that
```
ArthurBrussee, line 277: I think you could just use self.sync_queue() here right?

PR #197, opened 2024-10-23
Fix: Concurrent timestamps

nathanielsimard:
Not a pretty fix, but should avoid timetamps errors.

```
enable_timestamps
enable_timestamps

run_stuff_that_needs_timestamps // OK

disable_timestamps
run_stuff_that_needs_timestamps // ERROR

disable_timestamps


nathanielsimard:
Didn't work

PR #206, opened 2024-10-25
refactoring

adamnemecek:


nathanielsimard:
I think there are too many conflicts right now to be merged. The refactor was done at the same time the IR was actually being refactored. :cry:

adamnemecek:
If I fix it, will you consider merging it? Or will there be a better time in the future?

nathanielsimard:
> If I fix it, will you consider merging it? Or will there be a better time in the future?

Yup nothing wrong with using `Self` when possible :)

louisfd:
Been open for too long

PR #207, opened 2024-10-26
Feat/more types support

wingertge:
Adds support for `u8`, `u16`, `u64`, `i8`, `i16`, `i64` (in CUDA/HIP), `f64` (in CUDA/HIP). Adds `tf32` for use in CUDA CMMA. Adds `flex32` type that is marked as `RelaxedPrecision` in SPIR-V, which lets it operate at `f16` precision if the device supports it while still remaining a 32 bit type. Implemented as `f32` in all other backends.

Parameterizes tests to ensure different types all work.


crates/cubecl-core/src/frontend/element/float/tensor_float.rs
```diff
@@ -0,0 +1,297 @@
...
+
+#[inline]
+pub(crate) const fn tf32_to_f32(i: u32) -> f32 {
+    // TODO: Replace mem::transmute with from_bits() once from_bits is const-stabilized
+    // If NaN, keep current mantissa but also set most significiant mantissa bit
+    if i & 0x7FFFu32 > 0x7F80u32 {
+        unsafe { mem::transmute::<u32, f32>((i | 0x0200u32) << 13) }
+    } else {
+        unsafe { mem::transmute::<u32, f32>(i << 13) }
+    }
```
nathanielsimard, line 111: Is there some tests for that?

crates/cubecl-core/src/ir/kernel.rs
```diff
@@ -131,29 +144,45 @@ impl Elem {
                 FloatKind::BF16 => core::mem::size_of::<half::bf16>(),
                 FloatKind::F32 => core::mem::size_of::<f32>(),
                 FloatKind::F64 => core::mem::size_of::<f64>(),
+                FloatKind::Relaxed => core::mem::size_of::<f32>(),
+                FloatKind::TF32 => core::mem::size_of::<f32>(),
```
nathanielsimard, line 148: That function can be made `const` I think. It can help rust unroll some loops probably.

crates/cubecl-core/src/ir/kernel.rs
```diff
@@ -170,19 +199,35 @@ impl Display for Elem {
             Self::Float(kind) => match kind {
                 FloatKind::F16 => f.write_str("f16"),
                 FloatKind::BF16 => f.write_str("bf16"),
+                FloatKind::Relaxed => f.write_str("minf16"),
```
nathanielsimard, line 202: Why not `flex32` since it's a 32 bits type that can be relaxed.

crates/cubecl-core/src/runtime_tests/cmma.rs
```diff
@@ -42,6 +42,41 @@ pub fn kernel_simple_1(lhs: &Array<f16>, rhs: &Array<f16>, out: &mut Array<f32>)
     cmma::store(out.as_slice_mut(), &c, 16, cmma::MatrixLayout::RowMajor);
 }
 
+#[cube(launch)]
+/// Executes Out = Lhs @ Rhs.T
+pub fn kernel_simple_tf32<F: Cast>(lhs: &Array<F>, rhs: &Array<F>, out: &mut Array<f32>) {
+    let a = cmma::Matrix::<F>::from_slice(
+        cmma::MatrixIdent::A,
+        16,
+        16,
```
nathanielsimard, line 51: The function is generic over `F`, I think for clarity it would be better to hardcode `tf32`.

crates/cubecl-core/src/runtime_tests/mod.rs
```diff
@@ -17,21 +17,105 @@ pub mod unary;
...
+        $crate::testgen_untyped!();
+    };
+    ($f_def:ident: [$($float:ident),*], $i_def:ident: [$($int:ident),*], $u_def:ident: [$($uint:ident),*]) => {
+        use $crate::Runtime;
+
+        ::paste::paste! {
+            $(mod [<$float _ty>] {
+                type FloatT = $float;
+                type IntT = $i_def;
+                type UintT = $u_def;
```
nathanielsimard, line 41: I would prefer `T` to be the full name. Is it `T` for template, type or test?

crates/cubecl-cpp/src/shared/element.rs
```diff
@@ -530,17 +556,25 @@ impl<D: Dialect> Item<D> {
 impl<D: Dialect> Elem<D> {
     pub fn size(&self) -> usize {
```
nathanielsimard, line 557: Can be const

crates/cubecl-core/src/frontend/element/float/tensor_float.rs
wingertge: I actually need to change that, from testing it seems to actually have a compatible layout to `f32`, it's not shifted

crates/cubecl-core/src/ir/kernel.rs
wingertge: I missed that when I renamed it

crates/cubecl-core/src/runtime_tests/cmma.rs
wingertge: The problem is that there's some weirdness with trait resolution which made it not work using `tf32` directly for some reason. So I had to make it generic.

crates/cubecl-core/src/runtime_tests/mod.rs
wingertge: type, I used the C++ convention I guess (`float32_t`). I can just spell it out I guess, it's not used in that many places

nathanielsimard (approved):

PR #210, opened 2024-10-29
Allow for overriding device, return device setup

ArthurBrussee:
It'd be nice to allow for automatically overriding the wgpu device cube uses when using WgpuDevice::BestAvailable.

This PR adds a CUBECL_WGPU_DEVICE env var that be overriden. The names are the same as the enum - just implemented manually for now, I don't think it's worth pulling in more complex enum parsing machinery.

Additionally, this PR:
- Returns a "WgpuSetup" from init_(a)sync, to allow people to intercept what devices are actually being used
- Uses wpgu logic to select the highest performing device
- Remove some leftover code in the wpu-spirv server that also created devices. Only C::request_device needs to be different between these two backends.

### DIFF
--- a/crates/cubecl-wgpu/src/runtime.rs
+++ b/crates/cubecl-wgpu/src/runtime.rs
@@ -46,9 +46,9 @@ impl Runtime for WgpuRuntime<WgslCompiler> {
 
     fn client(device: &Self::Device) -> ComputeClient<Self::Server, Self::Channel> {
         RUNTIME.client(device, move || {
-            let (adapter, device_wgpu, queue) =
+            let setup =
                 future::block_on(create_wgpu_setup::<AutoGraphicsApi, WgslCompiler>(device));
-            create_client(adapter, device_wgpu, queue, RuntimeOptions::default())
+            create_client(setup, RuntimeOptions::default())
         })
     }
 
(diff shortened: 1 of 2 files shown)


crates/cubecl-wgpu/src/runtime.rs
```diff
@@ -90,95 +90,92 @@ impl Default for RuntimeOptions {
...
-pub fn init_existing_device(
-    adapter: Arc<wgpu::Adapter>,
-    device: Arc<wgpu::Device>,
-    queue: Arc<wgpu::Queue>,
-    options: RuntimeOptions,
-) -> WgpuDevice {
-    let device_id = WgpuDevice::Existing(device.as_ref().global_id());
-    let client = create_client(adapter, device, queue, options);
+#[derive(Clone)]
+pub struct WgpuSetup {
```
nathanielsimard, line 94: I would add some docs, since it's a public type.

crates/cubecl-wgpu/src/runtime.rs
```diff
@@ -90,95 +90,92 @@ impl Default for RuntimeOptions {
...
     RUNTIME.register(&device_id, client);
     device_id
 }
 
 /// Initialize a client on the given device with the given options. This function is useful to configure the runtime options
 /// or to pick a different graphics API. On wasm, it is necessary to use [`init_async`] instead.
-pub fn init_sync<G: GraphicsApi>(device: &WgpuDevice, options: RuntimeOptions) {
-    future::block_on(init_async::<G>(device, options));
-}
+pub fn init_sync<G: GraphicsApi>(device: &WgpuDevice, options: RuntimeOptions) -> WgpuSetup {
```
nathanielsimard, line 110: Instead of init I would use `setup`, since it returns the `WgpuSetup`.

crates/cubecl-wgpu/src/runtime.rs
```diff
@@ -90,95 +90,92 @@ impl Default for RuntimeOptions {
...
-    device: &WgpuDevice,
-) -> (wgpu::Device, wgpu::Queue, wgpu::Adapter) {
-    #[cfg(target_family = "wasm")]
-    let adapter = select_adapter::<G>(device).await;
-
-    #[cfg(not(target_family = "wasm"))]
-    let adapter = select_adapter::<G>(device);
-
+pub async fn create_wgpu_setup<G: GraphicsApi, C: WgpuCompiler>(device: &WgpuDevice) -> WgpuSetup {
+    let (instance, adapter) = select_adapter::<G>(device).await;
```
nathanielsimard, line 160: Hmm yeah it's confusing with the functions `init_async` and `init_sync`. Maybe instead of `init` we could have `register_setup_sync` and `register_setup_async`. Also we could remove `wgpu` from the function name here, since it's in the wgpu project: `create_setup` is probably sufficient.

crates/cubecl-wgpu/src/runtime.rs
```diff
@@ -250,60 +251,78 @@ fn select_adapter<G: GraphicsApi>(device: &WgpuDevice) -> wgpu::Adapter {
...
+                };
+
+                if override_device.is_none() {
+                    log::warn!("Unknown CUBECL_WGPU_DEVICE override {var}");
+                }
+                override_device
+            })
+    } else {
+        None
+    };
```
nathanielsimard, line 292: This could be in a single function I think, pretty well scoped.

crates/cubecl-wgpu/src/runtime.rs
ArthurBrussee: Yeah this could do with better naming. Register didn't feel right. This function does create new things, not just sign them up somewhere. Initialization is better imo (or maybe create). I went for this in the end:

- init_sync -> init_device_sync.
- init -> init_device
- init_existing_device -> init_device_on_setup
- create_client -> create_client_on_setup, and is now private, doesn't need to be public
- create_wgpu_setup -> create_setup_for_device, and is now private, doesn't need to be public

Lmk how that looks.

nathanielsimard (approved):

crates/cubecl-wgpu/src/runtime.rs
```diff
@@ -90,95 +82,112 @@ impl Default for RuntimeOptions {
...
+    pub adapter: Arc<wgpu::Adapter>,
+    /// The wpgu device Burn will use. Nb: There can only be one device per adapter.
+    pub device: Arc<wgpu::Device>,
+    /// The queue Burn commands will be submittd to.
+    pub queue: Arc<wgpu::Queue>,
+}
+
+/// Create a `WgpuDevice` on an existing `WgpuSetup`. Useful when you want to share
+/// a device between CubeCL and other wgpu libraries.
+pub fn init_device_on_setup(setup: WgpuSetup, options: RuntimeOptions) -> WgpuDevice {
```
AsherJingkongChen, line 102: They are confusing, are they all creating and registering devices?

- `fn init_device_on_setup(WgpuSetup) -> WgpuDevice`
- `async fn init_device(WgpuDevice) -> WgpuSetup`
- `fn init_device_async(WgpuDevice) -> WgpuSetup`

ArthurBrussee: Init_device and init_device_async are for initializing a cube WgpuDevice, and create all the wgpu resources.

init_device_on_setup initializes a cube WgpuDevice given existing wgpu resources.

Sorry if they are a bit confusing, maybe other names could be better!

AsherJingkongChen: @ArthurBrussee I have figured out their usage. However, I revised function names based on suggestions from @nathanielsimard in #214.

PR #211, opened 2024-10-29
Refactor for wgpu v23 compatibility with an example of wgpu device sharing

AsherJingkongChen:
### Related Issues/PRs

- tracel-ai/burn#2435
- gfx-rs/wgpu#6134
- gfx-rs/wgpu#6465

### Checklist

- [x] Confirmed that `cargo xtask test --ci` script has been executed.
- [x] Made sure the book is up to date with changes in this PR.
- [x] Checked for any updates or changes in dependencies.

#### 1. **Modified `entry_point` Type**
   - **Change**: Updated the type of `entry_point` from `&str` to `Option<&str>`.
   - **Reason**: Clippy's suggestion

#### 2. **Updated `WgpuDevice::Existing` Enum Type**
   - **Change**: Altered the type of `WgpuDevice::Existing` from `wgpu::Id<wgpu::Device>` to `u32`.
   - **Reason**: After the changes introduced in [gfx-rs/wgpu#6134](https://github.com/gfx-rs/wgpu/pull/6134), `wgpu::Id` no longer exists. Using a `u32` ensures a unique identifier for `WgpuDevice::Existing`.

   > **Note**: We avoid using pointers as identifiers due to the possibility of reallocation at the same memory address. To maintain uniqueness, a monotonically increasing value is now used internally.

### Examples

1. **Added `device_sharing` Example**
   - **Description**: This example demonstrates the device sharing feature.
   - **How to Run**: Use the following command to run the example:
     ```shell
     cargo run --example device_sharing --features wgpu
     ```


crates/cubecl-wgpu/src/runtime.rs
```diff
@@ -96,7 +96,7 @@ pub fn init_existing_device(
     queue: Arc<wgpu::Queue>,
     options: RuntimeOptions,
 ) -> WgpuDevice {
-    let device_id = WgpuDevice::Existing(device.as_ref().global_id());
+    let device_id = WgpuDevice::Existing(ptr::from_ref(device.as_ref()) as usize);
```
ArthurBrussee, line 99: This seems a bit dangerous! No guarantee your next device doesn''t land on the same memory adress - not unlikely even depending how you allocate stuff.

I think maybe just keeping an AtomicU64 which is incremented might be easier. Do have to be careful not to register something twice but that's already the case I think.

(sorry for the driveby comment - I added this code originally)

AsherJingkongChen: The code should explain your concerns, it does reallocation on the same address:

> No guarantee your next device doesn't land on the same memory address.

```rust
fn main() {
    let mut x = String::from("Hello, world!");
    let addr_x1 = &x as *const _ as usize;
    drop(x);
    x = String::from("Bonjour, le monde!");
    let addr_x2 = &x as *const _ as usize;
    assert_eq!(addr_x1, addr_x2);
}
```

ArthurBrussee: Well, exactly - two different strings and yet the adress is the same. So, in this case - you could feasible create two _different_ wgpu devices and yet have Cube think the are the same WgpuDevice, that's not the good! The id should be unique per device.

ArthurBrussee:
In theory now that the OOB behaviour has been fixed this should work now, if you have time to re test on metal would be amazing :) Thank you!

nathanielsimard:
> In linux CI, there is a segfault without message, would it come from CUDA? I can't resolve it now.

I don't think the CI is building CUDA.

nathanielsimard:
@AsherJingkongChen If you merge main, does it fix the CI?

crates/cubecl-wgpu/src/compiler/spirv.rs
```diff
@@ -280,7 +280,7 @@ fn request_device(
         adapter
             .device_from_raw(
                 vk_device,
-                true,
+                None,
```
AsherJingkongChen, line 283: I think I am doing the equivalent change here.

AsherJingkongChen:
I don't get it why did it pass CI. LOL.

nathanielsimard (approved):

syl20bnr:
> I don't get it why did it pass CI. I just updated feature flags. LOL.

This is because the CI errors was caused by the `spirv` feature which is not yet fully functional on CPU.

PR #213, opened 2024-10-29
Fix asynchronous autotuning for wasm

ArthurBrussee:
Fix tuning on wasm. This uses a slightly different approach from https://github.com/tracel-ai/cubecl/pull/203. In principle all autotuning is asynchronous, it just happens that native blocks on completion to make sure the message is sent. In the future, if needed (for eg. startup times), the tuning can happen on a thread (the Future is now Send as well).

This required some futzing with mutability, including with how checksums are handled. There is now a validate_checksum which writes the checksum state, instead of fastest having mut access or having a fastest_with_checksum. The checksum is now an Option<bool> to also store that we checked but failed.

The locking for the local tuner is still done carefully which I think prevents all the deadlocks, but, I haven't worked on that much so please do double check it!

> read only, check fast cache
> Write, create values if needed, validate checksums
> Read only, run autotune if needed
> Write, listen for messages of new results (which may or may not arrive immediately from step 3)

Tested on native & Wasm


nathanielsimard (approved):
Tested on native. It works!

PR #214, opened 2024-10-29
Adjust naming and comments

AsherJingkongChen:
There were 3 `init_device*` previously.

I found that it can be easier to identify, so I made this change.

| **Previous** | **Current** | **Type** |
|--|--|--|
| init_device_on_setup | init_device | `Fn(Setup) -> Device` |
| init_device_async | init_setup | `Fn(Device) -> Setup` |
| init_device | init_setup_async | `Fn(Device) -> Future<Setup>` |

### DIFF
--- a/crates/cubecl-wgpu/src/runtime.rs
+++ b/crates/cubecl-wgpu/src/runtime.rs
@@ -82,45 +82,43 @@ impl Default for RuntimeOptions {
     }
 }
 
-/// A complete setup used to run wgpu on a GPU.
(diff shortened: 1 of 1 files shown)


nathanielsimard:
I think the names are better, what do you think @ArthurBrussee?

ArthurBrussee:
Clearly previous names were confusing so all for it! Thanks for fixing them

nathanielsimard (approved):

PR #215, opened 2024-10-30
Fix setup for SPIR-V

wingertge:
The setup for SPIR-V was still using the old removed functions

### DIFF
--- a/crates/cubecl-runtime/src/memory_management/memory_lock.rs
+++ b/crates/cubecl-runtime/src/memory_management/memory_lock.rs
@@ -1,3 +1,6 @@
+#[cfg(not(feature = "std"))]
+use alloc::collections::BTreeSet;
+#[cfg(feature = "std")]
 use std::collections::HashSet;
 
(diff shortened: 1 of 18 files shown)


crates/cubecl-runtime/src/memory_management/memory_lock.rs
```diff
@@ -6,7 +9,10 @@ use crate::storage::StorageId;
 /// used for allocations currently.
 #[derive(Debug, Default)]
 pub struct MemoryLock {
+    #[cfg(feature = "std")]
     locked: HashSet<StorageId>,
+    #[cfg(not(feature = "std"))]
+    locked: BTreeSet<StorageId>,
```
nathanielsimard, line 15: We should always use BTreeSet I think, simpler and storage id is pretty much just a number.

nathanielsimard (changes requested):

nathanielsimard (approved):

PR #219, opened 2024-10-31
Add subcube and mma support for HIP compiler

syl20bnr:
On our AMD machine MMA matmul kernel is ~50% faster than Tiling2D. This is roughly the same as SPIR-V.

The `arch.rs` file list all the supported data format and matrix sizes.

There are still stability issues:
- random synchronization hangs after kernel launch in matmul benches, it happens both on full and half precision
- constant OOM error in half precision with MMA in matmul benches

### DIFF
--- a/crates/cubecl-common/src/benchmark.rs
+++ b/crates/cubecl-common/src/benchmark.rs
@@ -28,14 +28,14 @@ impl Display for TimingMethod {
     }
 }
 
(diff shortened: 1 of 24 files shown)


nathanielsimard (commented):
LGTM waiting for https://github.com/tracel-ai/cubecl/pull/220 to be fixed before merging

nathanielsimard (approved):

PR #220, opened 2024-10-31
Matmul components

louisfd:
I have made a massive refactor of Matmul kernels, with several components to make it modular. 

Throughout this new code, I use the "plane" word, which will soon be the new official name of the subcube. 

There are now four levels of Matmul, which can be composed. Here they are in increasing order of the problem they can solve. 

- Tile level: corresponds to the instruction where actual compute occurs. It always uses a single plane There are two implementations at the moment: 
   - Accelerated, which essentially encapsulates CMMA instruction
   - PlaneMMA, which uses cooperation in the plane but no specialized hardware [it's still pretty slow, to be optimized]
- Stage level: a matmul between staged data, meaning data that has been moved to local but fixed-size memory (typically, shared memory). It uses a whole Cube but for only one set of data. There is one implementation, where each plane is responsible for a whole row of tiles, and has an accumulator for each column. 
- Global level: a matmul between global lhs and rhs, but restricted to a subset of rows of lhs and columns of rhs, as large as the underlying stage. Uses one Cube and manages out of bounds. There is one implementation, which is the classic matmul loop. If there ever is plane specialization for double buffering, for instance, it will be a new implementation. 
- Batch level: a matmul between whole tensors, including batches. Uses the whole hypercube. For now there is one implementation, which launches one cube for each underlying global matmul. Next implementations will be adaptative and stream-k, which give the responsibility of several global matmuls to each Cube, matching the number of SMs of the GPU. 

The new way is in general much more flexible than the previous cmma (now `cmma_old`). It allows for instructions of different shapes and column major is now supported. However I'm not yet sure it is always at least as efficient as the old one, so I'm leaving it for now. Similarly for tiling2d, which is still much faster than PlaneMma.


crates/cubecl-linalg/src/matmul/components/problem.rs
```diff
@@ -0,0 +1,126 @@
...
+    pub lhs_layout: MatrixLayout,
+    pub rhs_layout: MatrixLayout,
+    pub lhs_line_size: u8,
+    pub rhs_line_size: u8,
+    pub out_line_size: u8,
+    pub _element: PhantomData<EG>,
+}
+
+impl<EG: Numeric> MatmulProblem<EG> {
+    #[cfg(feature = "export_tests")]
```
nathanielsimard, line 25: We could remove export tests here no?

crates/cubecl-linalg/src/matmul/kernels/cmma_old/config/base.rs
```diff
@@ -40,9 +41,9 @@ pub struct CmmaConfig {
 impl Default for CmmaConfig {
     fn default() -> Self {
         Self::new(
-            64,
-            32,
-            64,
+            128,
+            16,
+            128,
```
nathanielsimard, line 46: Should revert to previous, but not use the default config when launching cmma

crates/cubecl-linalg/src/matmul/kernels/cmma_old/config/mod.rs
```diff
@@ -2,6 +2,7 @@ mod base;
 mod predefined;
 mod strategy;
 
-pub(crate) use base::*;
+pub use base::*;
+#[cfg(feature = "export_tests")]
```
nathanielsimard, line 6: Don't need the feature flag.

PR #225, opened 2024-11-04
Remove leftover dead phi nodes from branch elimination

wingertge:
In rare cases, invalid phi nodes can be left over after branch elimination, which causes SPIR-V segfaults. This pass removes these invalid phis and replaces them with a simple copy.


nathanielsimard (approved):
LGTM

PR #227, opened 2024-11-05
Dot product of numeric lines + minor stuff

louisfd:
No major feature here. While trying to improve the PlaneMMA algorithm, I made some few adjustments here and there:
- small refactor of tile config
- Enable dot operation for lines of numeric
- Add some unrolling, which seems to gives some more speed

Unfortunately my other lined algorithm turned out hopeless, so I reverted to the previous


crates/cubecl-linalg/src/matmul/components/tile/plane.rs
```diff
@@ -165,22 +163,40 @@ impl<I: Numeric, O: Numeric, T: tile::Config, const M: u32, const N: u32, const
...
+
+            if comptime!(line_size == 1) {
+                #[unroll]
+                for col in 0..num_lines {
+                    slice[col + offset] = Line::cast_from(out[col]);
+                }
+                // It seems we must repeat this loop without unroll
+                for col in 0..num_lines {
+                    slice[col + offset] = Line::cast_from(out[col]);
+                }
```
nathanielsimard, line 189: Can we just remove the unroll instead?

louisfd: Sadly no, I know it makes no sense 😢 This is the bug that appears:
`---- tests::plane_mma_matmul::f32_ty::test_stage_matmul_s8x1x1 stdout ----
thread 'tests::plane_mma_matmul::f32_ty::test_stage_matmul_s8x1x1' panicked at /Users/louisfortier-dubois/Projects/cubecl/crates/cubecl-linalg/src/matmul/tests/cmma_matmul/matmul_test_launcher.rs:204:9:
Values differ more than epsilon: index=38 actual=-0.000000000000000007134327, expected=1.3014265, difference=1.3014265, epsilon=0.0001

---- tests::plane_mma_matmul::flex32_ty::test_stage_matmul_s8x1x1 stdout ----
thread 'tests::plane_mma_matmul::flex32_ty::test_stage_matmul_s8x1x1' panicked at /Users/louisfortier-dubois/Projects/cubecl/crates/cubecl-linalg/src/matmul/tests/cmma_matmul/matmul_test_launcher.rs:204:9:
Values differ more than epsilon: index=38 actual=-25291050000000, expected=1.3014265, difference=25291050000000, epsilon=0.8192`

Really don't know what's special with index 38 here.

PR #228, opened 2024-11-05
Ensure `LoopBreak` is updated when merging blocks

wingertge:
Fix oversight left over from when I added `LoopBreak`, which causes SPIR-V compiler panic in some cases.


louisfd (approved):
LGTM

PR #230, opened 2024-11-05
Matmul/transpose loader

louisfd:
Added the possibility of having different layouts in global memory and in stage memory, with the loader doing the transpose. 
This is because some tile matmuls should be faster in specific layouts.


crates/cubecl-linalg/src/matmul/components/global/tensor_view/cyclic_loading.rs
```diff
@@ -0,0 +1,125 @@
...
+#[cube]
+impl PlaneMapper for CyclicLoading {
+    fn plane_id() -> u32 {
+        UNIT_POS_Y
+    }
+
+    fn plane_unit() -> u32 {
+        UNIT_POS_X
+    }
+}
```
nathanielsimard, line 26: What is the plane mapper?

louisfd: Components must implement this trait when they use unit positions in the algorithm. It's a way to abstract the algorithms from unit positions. 

For now it's always plane_id=UNIT_POS_Y (which plane) and plane_unit=UNIT_POS_X (which unit within the plane)
So for  a cube dim of (32, 8, 1) plane_id is in 0..8 and plane_unit is in 0..32. 

It might be a bit overkill since every implementation of the plane mapper is the same for now, but the idea is that if we have plane specialization, for double buffering for instance, we can write the algorithm as if all planes contributed to the algorithm. For instance if we had planes 0..4 do something else and planes 4..8 performing the algorithm, then the plane mapper could map 4..8 to 0..4 so that the offset doesn't impact the algorithm. 

But that's future work anyway, for now it's really just abstracting unit positions in the algorithm.

crates/cubecl-linalg/src/matmul/kernels/cmma_matmul/config.rs
```diff
@@ -14,15 +15,28 @@ pub(crate) type CmmaSmmConfig<T> = stage::row_accumulate::Config<T>;
...
+    /// Ensure the inputs to tile matmul are in specified layout
+    ///
+    /// # Notes
+    ///
+    /// First item is for LHS, second item is for RHS
+    /// If None, the layout will be the same as in global memory
+    /// If enforced layout is different from global memory,
+    /// transpose will be done at loading from global memory to stage,
+    /// and stage will not be vectorized.
+    pub enforced_tile_layout: (Option<MatrixLayout>, Option<MatrixLayout>),
```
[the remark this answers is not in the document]
louisfd: Exactly, in that case the instruction can't assume the layout it's given as input and must be valid in both case (or eventually panic if we don't want the instruction to bother with the transpose)

nathanielsimard (approved):

PR #231, opened 2024-11-06
Perf/cuda fense

nathanielsimard:
Use cuda events instead of synching the stream for reading and syncing the server. At some point we might also create a pooling thread, but unsure it's necessary for now.

### DIFF
--- a/crates/cubecl-cuda/src/compute/fense.rs
+++ b/crates/cubecl-cuda/src/compute/fense.rs
@@ -1,0 +1,54 @@
+use cudarc::driver::sys::{CUevent_flags, CUevent_st, CUevent_wait_flags, CUstream_st};
+
+/// A fense is a simply an [event](CUevent_st) created on a [stream](CUevent_st) that you can wait
+/// until completion.
+///
+/// This is useful for doing synchronization outside of the compute server, which is normally
+/// locked by a mutex or a channel. This allows the server to continue accepting other tasks.
+pub struct Fense {
+    stream: *mut CUstream_st,
+    event: *mut CUevent_st,
+}
+
+// If we don't close the stream or destroy the event, it is safe.
+//
+// # Safety
+//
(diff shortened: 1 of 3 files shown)


crates/cubecl-cuda/src/compute/fense.rs
```diff
@@ -0,0 +1,54 @@
+use cudarc::driver::sys::{CUevent_flags, CUevent_st, CUevent_wait_flags, CUstream_st};
+
+/// A fense is a simply an [event](CUevent_st) created on a [stream](CUevent_st) that you can wait
+/// until completion.
+///
+/// This is useful for doing synchronization outside of the compute server, which is normally
+/// locked by a mutex or a channel. This allows the server to continue accepting other tasks.
+pub struct Fense {
```
louisfd, line 8: Pretty sure it should be Fence 😄

crates/cubecl-cuda/src/compute/fense.rs
```diff
@@ -0,0 +1,54 @@
+use cudarc::driver::sys::{CUevent_flags, CUevent_st, CUevent_wait_flags, CUstream_st};
+
+/// A fense is a simply an [event](CUevent_st) created on a [stream](CUevent_st) that you can wait
```
louisfd, line 3: is a simply an -> remove a

crates/cubecl-cuda/src/compute/fense.rs
```diff
@@ -0,0 +1,54 @@
...
+pub struct Fense {
+    stream: *mut CUstream_st,
+    event: *mut CUevent_st,
+}
+
+// If we don't close the stream or destroy the event, it is safe.
+//
+// # Safety
+//
+// Since streams are never closed and we destroy the event after waiting, which comsumes the
```
louisfd, line 17: consumes

louisfd (changes requested):
Change fense to fence, then LGTM

PR #232, opened 2024-11-06
Perf/cuda fence

nathanielsimard:


crates/cubecl-cuda/src/compute/fence.rs
```diff
@@ -0,0 +1,54 @@
...
+pub struct Fence {
+    stream: *mut CUstream_st,
+    event: *mut CUevent_st,
+}
+
+// If we don't close the stream or destroy the event, it is safe.
+//
+// # Safety
+//
+// Since streams are never closed and we destroy the event after waiting, which comsumes the
```
louisfd, line 17: consumes

louisfd (approved):
just two typos :)

PR #236, opened 2024-11-07
Refactor metadata to make it more simple and flexible

wingertge:
Refactors metadata into a format that's more flexible and extensible, to allow for proper per-tensor rank support, buffer length, etc.
Adds tests for all kinds of metadata.
Refactors the `wgpu` pipieline creation to manually specifiy the bind group layout. This allows unused tensors without any bind group mismatch errors. Required now that `info` is no longer automatically accessed for the rank. Fixes some problems with the optimizer/SPIR-V that were uncovered with the new matmul tests.

See `metadata.rs` for details on the conceptual metadata representation.

## Breaking changes

There's a minor breaking change anywhere that uses `TensorArg::from_raw_parts`. It now requires either a generic type or a raw element size in bytes via `from_raw_parts_and_size`. The latter is required for fusion since types aren't statically known there. The element size is used to properly calculate `buffer_len`.


crates/cubecl-core/src/codegen/integrator.rs
```diff
@@ -212,19 +212,23 @@ fn is_contiguous(strides: &[usize]) -> bool {
 /// Information related to an input.
 #[derive(Clone, Debug)]
 pub enum InputInfo {
-    Array { item: Item, visibility: Visibility },
-    Scalar { elem: Elem, size: usize },
+    Array {
+        item: Item,
+        visibility: Visibility,
+        has_extended_meta: bool,
```
nathanielsimard, line 218: I would put some docs, saying  _when_ we need to set it to true.

crates/cubecl-core/src/codegen/metadata.rs
```diff
@@ -0,0 +1,144 @@
...
+//!     logical_lengths: [u32; N],
+//! }
+//!
+//! struct ExtendedMeta<const N: usize> {
+//!     ranks: [u32; N],
+//!     shape_offsets: [usize; N],
+//!     stride_offsets: [usize; N],
+//!     shapes: Vec<u32>,
+//!     strides: Vec<u32>
+//! }
```
nathanielsimard, line 21: Maybe we could have more precise names for each metadata block:

```rust
 struct Metadata<const NB: usize, const NT: usize> {
    buffer: BufferMetadata<NB>,
    tensor: TensorMetadata<NT>,
}
```

crates/cubecl-core/src/codegen/metadata.rs
```diff
@@ -0,0 +1,144 @@
...
+// Extended Metadata
+const RANK: u32 = 0;
+const SHAPE_OFFSETS: u32 = 1;
+const STRIDE_OFFSETS: u32 = 2;
+
+#[derive(Clone, Debug, Default)]
+pub struct Metadata {
+    num_meta: u32,
+    num_extended_meta: u32,
+}
```
nathanielsimard, line 42: Not sure what this struct does really 😅

crates/cubecl-core/src/codegen/metadata.rs
```diff
@@ -0,0 +1,144 @@
...
+
+#[derive(Default)]
+pub struct MetadataBuilder {
+    buffer_lens: Vec<u32>,
+    lengths: Vec<u32>,
+    ranks: Vec<u32>,
+    shapes: Vec<Vec<u32>>,
+    strides: Vec<Vec<u32>>,
+}
```
nathanielsimard, line 92: This is the way!

crates/cubecl-core/src/codegen/metadata.rs
```diff
@@ -0,0 +1,144 @@
...
+
+impl MetadataBuilder {
+    pub fn with_array(&mut self, buffer_len: u32, len: u32) {
+        self.buffer_lens.push(buffer_len);
+        self.lengths.push(len);
+    }
+
+    pub fn with_tensor(
```
nathanielsimard, line 100: I think the order is important. So maybe adding a comment on that one.

nathanielsimard: The order you called the builder I mean

crates/cubecl-core/src/frontend/container/tensor/base.rs
```diff
@@ -34,7 +31,7 @@ mod metadata {
             unexpanded!()
         }
 
-        /// The length of the buffer representing the tensor.
+        /// The number of unvectorized elements in the tensor.
```
nathanielsimard, line 34: Wrong comment updates I think! It is vectorized!

crates/cubecl-wgpu/src/compiler/wgsl/compiler.rs
```diff
@@ -10,25 +10,28 @@ use cubecl_core::{
...
+    ComputePipeline, DeviceDescriptor, PipelineLayoutDescriptor, ShaderModuleDescriptor,
+    ShaderStages,
+};
 
 /// Wgsl Compiler.
 #[derive(Clone, Default)]
 pub struct WgslCompiler {
     num_inputs: usize,
     num_outputs: usize,
+    metadata: Metadata,
```
nathanielsimard, line 27: So the `Metadata` type is acting like the metadata of the metadata 😅. Maybe we could name the type `MetadataHeader` ?

nathanielsimard (changes requested):

crates/cubecl-core/src/codegen/metadata.rs
wingertge: I kind of wanted to decouple it from the specific `Tensor` name because we may have other types with extended metadata at some point. Offsets allow for dynamically sized metadata so you could theoretically make any kind of data structure that has additional metadata associated with it, as long as it could be encoded as series of u32. That's why I wanted to use the more generic `ExtendedMetadata`

crates/cubecl-core/src/codegen/metadata.rs
wingertge: Forgot to add docs, but also it's just a helper that calculates the offset on the compiler side so we can change the layout without changing every usage in the backend.

crates/cubecl-core/src/codegen/metadata.rs
wingertge: I added one to the builder at the top, it's the same order as usual (same order as the pipeline bindings)

crates/cubecl-wgpu/src/compiler/wgsl/compiler.rs
wingertge: It represents the metadata conceptually on the compiler side, so compilers don't need to do their own offset calculations. So I think the name is appropriate. It's the dynamic equivalent of `offset_of!`.

nathanielsimard (approved):

PR #237, opened 2024-11-07
Matmul: CPU reference with same precisions

louisfd:
CPU reference was always in the same precision of the input/outputs for the whole computation, while in the gpu algorithm it can go through several precisions. I made the reference match the same types when possible. When impossible I made epsilon less permissive. Now all matmul tests should pass on cuda.


nathanielsimard (approved):
LGTM

PR #238, opened 2024-11-07
Matmul/tensor reference

louisfd:
Matmul now takes tensor references as input, rather than owned tensors. 

It used to take owned tensors because (somehow) cube allowed it with the way I used them, and it simplified the tensor view, which could own it instead of owning a reference. 

But for further improvements I needed the tensor as a reference. In the tensor view I turned the Tensor into a pointer and used unchecked indexing.


nathanielsimard (approved):
LGTM

PR #240, opened 2024-11-08
Fix exclusive memory only and custom wgsl kernels

wingertge:
Fix to layout creation for `exclusive_memory_only` and for custom WGSL kernels.


nathanielsimard (approved):
LGTM

PR #242, opened 2024-11-08
Matmul: batch one_to_many + refactor configuration

louisfd:
Edit: one_to_many works

I started working on one_to_many batch matmul, where one cube is responsible of several passes of global matmul (doesn't work yet, but nothing uses it). 
I realized it was not easy to test another batch strategy, because launching with custom configurations was difficult, and in particular tests had no flexibility. 
So I did a huge refactor, now all "Algorithms" for matmul are specified via the trait Algorithm, which has several associated types. 
Tests also use this, declaring their own algorithm trait. They underwent a great job of uniformization and nomenclature.


crates/cubecl-linalg/src/matmul/kernels/matmul/algorithm/base.rs
```diff
@@ -0,0 +1,55 @@
...
+            Self::ES,
+            Self::EG,
+            <<Self::GlobalMatmul as global::Matmul<Self::EG, Self::ES>>::Lhs as global::Loader<
+                Self::EG,
+                Self::ES,
+            >>::StageReader,
+            <<Self::GlobalMatmul as global::Matmul<Self::EG, Self::ES>>::Rhs as global::Loader<
+                Self::EG,
+                Self::ES,
+            >>::StageReader,
```
nathanielsimard, line 29: We should use type aliases for that.

crates/cubecl-linalg/src/matmul/tests/test_macros/cmma/matmul_algorithm.rs
```diff
@@ -0,0 +1,3240 @@
...
+// global: g[m]x[n]x[k], with m,n,k the whole matrix dimensions
+// stage: s[m]x[n]x[k], with m,n,k the number of tiles along those dims
+// tile: t[m]x[n]x[k], with m,n,k the tile dimensions. tile algorithm is given by macro arguments
+// layouts: [r/c][r/c], r=row, c=col, respectively for lhs and rhs
+// line size: ln[v], with v the line size of all tensors. if different then ln[v_lhs]x[v_rhs]x[v_out]
+// Other specifications may be appended at the end
+
+#[allow(missing_docs)]
+#[macro_export]
+macro_rules! matmul_test_define {
```
nathanielsimard, line 12: Is it a refactor of all matmul tests?

PR #244, opened 2024-11-11
Add topology properties to client

wingertge:
Adds topology properties to the client to directly read out things like subcube size from the device instead of assuming a warp size of 32.

### DIFF
--- a/crates/cubecl-runtime/src/memory_management/mod.rs
+++ b/crates/cubecl-runtime/src/memory_management/mod.rs
@@ -81,6 +81,19 @@ pub struct MemoryDeviceProperties {
 }
 
 /// Properties of the device related to topology.
+///
+/// # Subcube size min/max
+///
+/// This is a range of possible values for the subcube size.
+///
+/// For Nvidia GPUs and HIP, this is a single fixed value.
+///
+/// For wgpu with AMD GPUs this is a rannge of possible values, but the actual configured value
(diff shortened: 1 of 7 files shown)


crates/cubecl-runtime/src/memory_management/mod.rs
```diff
@@ -79,3 +79,12 @@ pub struct MemoryDeviceProperties {
...
 }
+
+/// Properties of the device related to topology.
+#[derive(Debug, Clone)]
+pub struct TopologyProperties {
+    /// The minimum size of a subcube on this device
+    pub subcube_size_min: u32,
+    /// The maximum size of a subcube on this device
+    pub subcube_size_max: u32,
+}
```
nathanielsimard, line 90: Is the first dimension in the Cube responsible for the subcube size? For example, if I have an AMD GPU with a max of 64 and a min of 32, and I launched a kernel with (32, 8, 1), would I have 8 subcubes of 32, or 4 of 64? It would be great to add the answer here as a comment.

wingertge: I don't believe there's an actual answer that's documented anywhere. In HIP it's a compiler flag that determines it, wgpu is still very in flux and not very well defined.  For Intel, it depends on the registers. Really, it's quite unreliable and the only definitive number is the one fetched at runtime. But this at least provides a rough estimate for determining things like cube size, with the exact details left to each kernel.

nathanielsimard (approved):

PR #246, opened 2024-11-11
Matmul: slice level

louisfd:
Introduction of a new level of matmul, which I've put less forward than the others: the slice matmul, that goes between the batch matmul and the global matmul. As its name implies, it does a matmul on slices of the tensor inputs, using one cube but performing several executions for it, to fill all the slice. 

The way we iterate in the slice can follow row-major, col-major or swizzle, the latter is supposed to maximize cache L2 use. 

TODO:
- Row/Col/Swizzle for cube dispatch, like we had in cmma_old. I tried to do it but had difficulty finding a beautiful way. 
- Initialize global matmul's accumulator outside, so we can reuse it.


crates/cubecl-linalg/src/matmul/components/batch/span.rs
```diff
@@ -0,0 +1,155 @@
...
+pub struct SpanDim {
+    start: u32,
+    end: u32,
+    step: u32,
+}
+
+#[cube]
+impl SpanDim {
+    pub fn new(shape: u32, stage: u32, cube_pos: u32, num_cubes: u32) -> SpanDim {
+        let num_stages = (shape + stage - 1) / stage;
```
nathanielsimard, line 32: Would be great to add a bit more docs here.

PR #248, opened 2024-11-11
[Fix] Merge vector definitions for `F32` and `TF32`

wingertge:
Merges vector definitions for `F32` and `TF32` since they're both represented by `float` in C++.
Fixes small oversight in `f16` optimized remainder


nathanielsimard (approved):
LGTM

PR #255, opened 2024-11-12
Refactor/slice lifetime

nathanielsimard:
Since we really need to work with slices a lot, and lifetimes don't add much to safety / correctness, I made them simpler without lifetime.


louisfd (approved):
LGTM

PR #256, opened 2024-11-13
Add max_bindings hardware limit

ArthurBrussee:
Keep track of max_bindings in client properties.

Rather than adding something new I've renamed TopologyProperties to HardwareProperties. Maybe a slight misnomer as it's not just hardware but hardware + software that determine some of these limits. It's nice to have a place for this as we need more hardware limits.


nathanielsimard (approved):
LGTM, but there are conflicts to solve before merging.

ArthurBrussee:
Rebased it now!

PR #267, opened 2024-11-16
Add Runtime trait method to get default kernel/shader extension

torsteingrindvik:
Useful in cases where kernels are compiled and written to files in a generic setting.


torsteingrindvik:
I used this locally to write out cu, spv, and wgsl files which worked nicely.
I don't have HIP tools installed to try that, but from https://github.com/ROCm/rocm-examples/blob/develop/HIP-Basic/saxpy/main.hip it looks like "hip" is the extension they use so I assume that's what CubeCL uses too.

nathanielsimard (approved):

PR #268, opened 2024-11-17
Inform users that launch kernels cannot have return types and suggest…

### LINKED ISSUE #216: Kernels should give a helpful error message if users put return values on them
Example:

```rust
#[cube(launch_unchecked)]
pub fn gelu_array(...) -> u32 {  // NOTE: Cannot actually get return value
  ...

  123u32
}
```

Kernels which have `cube(launch)` or `cube(launch_unchecked)` are allowed to have return values, but there isn't a way to get return values.

If users put return values on kernels, the proc macro should detect this and produce an error message instructing the user on what to do instead.

(Instead: Likely go via an output handle etc.)

torsteingrindvik:
… what to do.

Fixes #216 

For example here:

```rust
use cubecl::prelude::*;

#[cube(launch_unchecked)]
pub fn bad<F: Float>(input: &Array<Line<F>>, output: &mut Array<Line<F>>) -> u8 {
        output[ABSOLUTE_POS] = input[ABSOLUTE_POS];
    }

    123
}

fn main() {}
```

When compiling, this is shown in the terminal:

image

There is an error msg in the outer `#[cube(..)]` position too:

image

Been a while since I did work on proc macros, let me know if there is an elegant way to get around that

### DIFF
--- a/crates/cubecl-macros/src/parse/kernel.rs
+++ b/crates/cubecl-macros/src/parse/kernel.rs
@@ -1,10 +1,11 @@
 use crate::{expression::Block, paths::prelude_type, scope::Context, statement::Pattern};
 use darling::{ast::NestedMeta, util::Flag, FromMeta};
 use proc_macro2::TokenStream;
+use quote::ToTokens;
 use std::iter;
 use syn::{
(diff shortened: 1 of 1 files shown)


nathanielsimard (approved):
LGTM

PR #273, opened 2024-11-19
Matmul: customizable cube dispatch

louisfd:
Cubes can work on other places than their CUBE_POS naturally suggest.


nathanielsimard (approved):
LGTM

PR #274, opened 2024-11-19
Reduce on cuda

maxtremblay:
The initial goal was to support testing the reduction algorithms on CUDA. This lead to some work on the warp operations in the compiler to support vectorization. I also added a bunch of related tests.

### DIFF
--- a/crates/cubecl-core/src/runtime_tests/plane.rs
+++ b/crates/cubecl-core/src/runtime_tests/plane.rs
@@ -101,7 +101,7 @@ pub fn test_plane_sum<TestRuntime: Runtime, F: Float + CubeElement + Sized>(
             kernel_sum::launch::<F, TestRuntime>(
                 &client,
                 cube_count,
-                CubeDim::new(32, 1, 1),
+                CubeDim::new(plane_size, 1, 1),
                 handle,
             )
         },
@@ -140,7 +140,7 @@ pub fn test_plane_prod<TestRuntime: Runtime, F: Float + CubeElement>(
             kernel_prod::launch::<F, TestRuntime>(
                 &client,
                 cube_count,
-                CubeDim::new(32, 1, 1),
+                CubeDim::new(plane_size, 1, 1),
                 handle,
             )
         },
@@ -176,7 +176,7 @@ pub fn test_plane_max<TestRuntime: Runtime, F: Float + CubeElement>(
             kernel_max::launch::<F, TestRuntime>(
                 &client,
                 cube_count,
-                CubeDim::new(32, 1, 1),
+                CubeDim::new(plane_size, 1, 1),
                 handle,
             )
         },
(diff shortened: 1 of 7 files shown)


crates/cubecl-core/src/runtime_tests/plane.rs
```diff
@@ -75,132 +74,296 @@ pub fn kernel_broadcast<F: Float>(output: &mut Tensor<F>) {
...
 }
 
 pub fn test_plane_prod<TestRuntime: Runtime, F: Float + CubeElement>(
     client: ComputeClient<TestRuntime::Server, TestRuntime::Channel>,
+    vectorization: u8,
 ) {
+    let plane_size = 32;
+    let input: Vec<f32> = (0..plane_size * vectorization as u32)
+        .map(|x| match x % 3 {
+            0 => 0.5,
```
louisfd, line 118: A bit picky on my part, but I think you should pick other values. 0.5 and 2. will cancel each other, while 1.0 doesn't affect the result. Therefore it would be possible to arrive at the right result coincidentally

crates/cubecl-core/src/runtime_tests/plane.rs
```diff
@@ -75,132 +74,296 @@ pub fn kernel_broadcast<F: Float>(output: &mut Tensor<F>) {
...
+        &expected,
+        vectorization,
         client.clone(),
-        |cube_count: CubeCount, cube_dim, handle| {
-            kernel_sum::launch::<F, TestRuntime>(&client, cube_count, cube_dim, handle)
+        |cube_count, handle| {
+            kernel_sum::launch::<F, TestRuntime>(
+                &client,
+                cube_count,
+                CubeDim::new(32, 1, 1),
```
louisfd, line 104: We should use plane_size instead of 32

crates/cubecl-core/src/runtime_tests/plane.rs
```diff
@@ -75,132 +74,296 @@ pub fn kernel_broadcast<F: Float>(output: &mut Tensor<F>) {
...
+    vectorization: u8,
 ) {
+    let plane_size = 32;
+    let mut input: Vec<f32> = (0..plane_size * vectorization as u32)
+        .map(|x| (x % 5) as f32) // the predicate is x < 5 which is always satisfied at this step.
+        .collect();
+
+    for k in 0..vectorization as usize {
+        if k % 2 == 0 {
+            input[4 * vectorization as usize + k] = 10.0; // Make all even batches false by setting an element to be > 5.
```
louisfd, line 233: What's the 4 for?

louisfd (approved):
I have minor comments about tests, otherwise LGTM

crates/cubecl-core/src/runtime_tests/plane.rs
maxtremblay: changed to 0.75, 1.25 and 1.75 (I want number that are exactly representable to avoid floating-point errors)

crates/cubecl-core/src/runtime_tests/plane.rs
maxtremblay: I want to problematic entries to be in the middle. So, I push them by 4 * vectorization. I don't have a good justification, it just feel like a more robust test to me, but it is somewhat subjective.

PR #277, opened 2024-11-19
Read many buffers at once

nathanielsimard:
The read operation can now read multiple bindings at the same time, increasing GPU utilization.

### DIFF
--- a/crates/cubecl-core/src/runtime_tests/assign.rs
+++ b/crates/cubecl-core/src/runtime_tests/assign.rs
@@ -24,7 +24,7 @@ pub fn test_kernel_assign_scalar<R: Runtime, F: Float + CubeElement>(
         unsafe { ArrayArg::from_raw_parts::<F>(&handle, 2, vectorization) },
     );
 
-    let actual = client.read(handle.binding());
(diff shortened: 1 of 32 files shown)


nathanielsimard:
> I think this could also just be done by joining futures of multiple reads?

Not really, because a single read flushes the queue. In this case, we flush only after all copies are pushed into the encoder.

ArthurBrussee:
I guess but that sounds a tiny bit faster at best? Worst case could just check if task_count == 0 and skip the flush? Waiting for multiple things at once is a good thing but yeah, if we're paying all the complexity for async already anyway...

nathanielsimard:
> I guess but that sounds a tiny bit faster at best? Worst case could just check if task_count == 0 and skip the flush? Waiting for multiple things at once is a good thing but yeah, if we're paying all the complexity for async already anyway...

It's the difference between 90% utilization and 99-100% utilization 😅. Maybe on wasm it's not that bad, but on native it is. What we could do is delay the mapping of buffers until the next compute task, but it's hard to know when you can do that without creating deadlocks.

For simplicity we could simply remove the `read` method and only keep `read_many`.

PR #278, opened 2024-11-19
Propagate kernel name via KernelSettings

### LINKED ISSUE #234: CUDA kernels names shouldn't all be "kernel"
Hardcoded here:

1. https://github.com/tracel-ai/cubecl/blob/main/crates/cubecl-cuda/src/compute/server.rs#L366
2. https://github.com/tracel-ai/cubecl/blob/main/crates/cubecl-cpp/src/shared/kernel.rs#L125

I think that perhaps either
1. The kernel name should always be equal to the Rust name, or
2. It should default to the Rust name but let the user name it via some API or via the proc macro (e.g. `#[cube(launch, name = "foo")]`

The reason for having names is that if profiling with Nsight tools we can see which kernels are which. This is diffifcult when the names are the same.

torsteingrindvik:
Fixes #234 

Names now show up!

image

Propagating this down from the proc macro until both `crates/cubecl-cpp/src/shared/kernel.rs` as well as `crates/cubecl-cuda/src/compute/server.rs` could get a hold of the name required a bit of `kernel_name` field duplication in the relevant data structures.
If there's a more straight forward way to propagate it I'm happy to do that, but I couldn't see any obvious way.

Also notice from the image that the same four names show up twice, but with different performance characteristics.
This is because the example I ran first goes through f32 then f16.
I hope to support automatically adding a suffix in a follow-up if it's not too hairy.

### DIFF
--- a/crates/cubecl-core/src/compute/kernel.rs
+++ b/crates/cubecl-core/src/compute/kernel.rs
@@ -7,6 +7,7 @@ use cubecl_runtime::ExecutionMode;
 /// A kernel, compiled in the target language
 pub struct CompiledKernel<C: Compiler> {
     pub name: Option<&'static str>,
+    pub kernel_name: String,
     /// Source code of the kernel
     pub source: String,
     /// In-memory representation of the kernel
@@ -188,6 +189,7 @@ pub struct KernelTask<C: Compiler, K: Kernel> {
 impl<C: Compiler, K: Kernel> CubeTask<C> for KernelTask<C, K> {
     fn compile(&self, mode: ExecutionMode) -> CompiledKernel<C> {
         let gpu_ir = self.kernel_definition.define();
+        let kernel_name = gpu_ir.kernel_name.clone();
         let cube_dim = gpu_ir.cube_dim;
         let lower_level_ir = C::compile(gpu_ir, mode);
         let shared_mem_bytes = lower_level_ir.shared_memory_size();
@@ -199,6 +201,7 @@ impl<C: Compiler, K: Kernel> CubeTask<C> for KernelTask<C, K> {
             cube_dim,
             shared_mem_bytes,
             debug_info: None,
+            kernel_name,
         }
     }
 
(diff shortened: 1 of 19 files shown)


ArthurBrussee:
That's awesome! Had the same problem, but in wgpu. Would it be possible to use the names as a label for the shader on wgpu too?

torsteingrindvik:
> That's awesome! Had the same problem, but in wgpu. Would it be possible to use the names as a label for the shader on wgpu too?

Yes, from what I've done in wgpu I know that the entry point can be named by users somewhere in the pipeline setup.
Then similarly the CubeCL code that produces wgsl source code should match the pipeline setup.

Do you have a debugger for wgpu that shows just "main" "main" "main" etc.? Asking because I don't know what the best way of seeing compute shader calls from wpgu is.


@maintainers: If you see this PR and you're happy please go ahead and merge. I'll then do the wgpu in a follow-up as I'm out of time right now. Else I'll continue here next time I'm coding

crates/cubecl-core/src/compute/kernel.rs
```diff
@@ -7,6 +7,7 @@ use cubecl_runtime::ExecutionMode;
 /// A kernel, compiled in the target language
 pub struct CompiledKernel<C: Compiler> {
     pub name: Option<&'static str>,
+    pub kernel_name: String,
```
nathanielsimard, line 10: What is the difference between `name` and `kernel_name` ?

torsteingrindvik: Sure. I added that back, but changed the names to "entrypoint_name" (the new one) and "debug_name" (the old one).

To make it less confusing I added some docs:

image

torsteingrindvik:
Added a commit to make wgpu + wgsl/spirv also use the kernel name as their entrypoint names.

nathanielsimard (approved):
LGTM Thanks 🙏

PR #279, opened 2024-11-20
Add HIP wmma intrinsic compiler

syl20bnr:
Tested with `cargo test` in `cubecl-hip` crate and with benches.
Note that benches hang sometimes so you can interrupt and relaunch them until they don't hang.

For intrinsic:

```sh
cargo bench --bench matmul --features hip
```

For rocwmma:

```sh
cargo bench --bench matmul --features hip-rocwmma
```

Support for different wavefront size depending on the hardware will come in a follow up PR.


crates/cubecl-linalg/src/matmul/kernels/cmma_old/availability.rs
```diff
@@ -23,6 +23,10 @@ pub fn check_cmma_availability<R: Runtime, F: CubePrimitive>(
     cmma_config: &CmmaConfig,
 ) -> Result<(), UnavailabilityReason> {
     let tile_dim: TileDimension = cmma_config.tile_dimension_strategy.into();
+    println!("tile dim m: {:?}", tile_dim.m);
+    println!("tile dim n: {:?}", tile_dim.n);
+    println!("tile dim k: {:?}", tile_dim.k);
+    println!("c type: {:?}", F::as_elem());
```
nathanielsimard, line 29: Should remove the println!

crates/cubecl-linalg/src/matmul/tests/cmma_old/test_cases.rs
```diff
@@ -103,6 +103,8 @@ pub(crate) fn test_cmma<R: Runtime, F: Float + CubeElement + Display>(
     device: &R::Device,
 ) -> Result<(), String> {
     let client = R::client(device);
+    println!("Device properties: {:?}", client.properties());
+    println!("Config: {:?}", config);
```
nathanielsimard, line 107: Same

nathanielsimard (changes requested):

PR #280, opened 2024-11-20
Implement a reduction accross lines

maxtremblay:


crates/cubecl-std/src/reduce/sum.rs
```diff
@@ -10,6 +10,8 @@ pub struct ReduceConfig {
 
 /// Compute the sum of all elements of `input` and write it to the first element of `output`.
 ///
+/// This doesn't reduce values accross lines. For a version that does, use [reduce_sum_lined].
```
louisfd, line 13: across

louisfd (approved):
LGTM

PR #289, opened 2024-11-21
Add missing mma.h include in wmma CUDA compiler

syl20bnr:
Regression introduced in the last commit for HIP intrinsic.


louisfd (approved):
LGTM

PR #294, opened 2024-11-22
Query num planes

maxtremblay:
I add a new UNIT_POS_PLANE constant to query to index of a unit within a plane. Then, I reimplement reduce_sum using plane-base primitives instead of cube dimensions. 

I tested the implement locally for cuda and wgpu and on the amd machine for hip and wgpu.


nathanielsimard (approved):
LGTM

PR #296, opened 2024-11-22
Consistent OOB behaviour for wgpu

ArthurBrussee:
Currently, CubeCL relies on wgpu to have a consistent behaviour for OOB indexing (read-0, write-discard). While this so far has been true, it's not true in Dawn on WebGPU. The WebGPU specification defines an OOB as a "dynamic error" that might result in anything including program termination. In practice it's not that severe - but some backends do have differing behaviour like clamping the index instead of discarding a write.

This PR changes Cube to use an indexing mode much like cubecl-cpp, and insert checks manually. For performance, I still do a read-oob-0 by using a `select()` which hopefully compiles to a conditional move. This isn't quite correct as the WebGPU spec allows any kind of behaviour when indexing OOB, but in practice it can't do much besides picking a random in bound index, so this is practically safe.

On Vulkan + Spir-V we're relying on Vulkan robustness, this doens't change anything there.

In the future we could also disable these checks on Vulkan with robustness when using WGSL (or any other platforms where this behavious is guaranteed).

This might unblock https://github.com/tracel-ai/cubecl/pull/211, as mentioned here: https://github.com/tracel-ai/burn/pull/2435 metal had new issues as the OOB behaviour in WGPU changes.

Very much not my area so please have Genna sign this off!

### DIFF
--- a/crates/cubecl-core/src/runtime_tests/index.rs
+++ b/crates/cubecl-core/src/runtime_tests/index.rs
@@ -1,0 +1,56 @@
+use crate as cubecl;
+
+use cubecl::prelude::*;
+
+#[cube(launch)]
+pub fn kernel_assign<F: Float>(output: &mut Array<F>) {
+    if UNIT_POS == 0 {
+        let item = F::new(5.0);
+        // Assign normally.
+        output[0] = item;
+
+        // out of bounds write should not show up in the array.
+        output[2] = F::new(10.0);
+
+        // out of bounds read should be read as 0.
+        output[1] = output[2];
+    }
+}
+
+pub fn test_kernel_index_scalar<R: Runtime, F: Float + CubeElement>(
+    client: ComputeClient<R::Server, R::Channel>,
+) {
+    let handle = client.create(F::as_bytes(&[F::new(0.0), F::new(1.0), F::new(123.0)]));
+    let handle_slice = handle.clone().offset_end(1);
+    let vectorization = 1;
+
+    kernel_assign::launch::<F, R>(
+        &client,
+        CubeCount::Static(1, 1, 1),
+        CubeDim::default(),
+        unsafe { ArrayArg::from_raw_parts::<F>(&handle_slice, 3, vectorization) },
+    );
+
+    let actual = client.read_one(handle.binding());
+    let actual = F::from_bytes(&actual);
+
+    assert_eq!(actual[0], F::new(5.0));
+    assert_eq!(actual[1], F::new(0.0));
+    assert_eq!(actual[2], F::new(123.0));
+}
+
+#[allow(missing_docs)]
+#[macro_export]
+macro_rules! testgen_index {
+    () => {
+        use super::*;
+
+        #[test]
+        fn test_assign_index() {
+            let client = TestRuntime::client(&Default::default());
+            cubecl_core::runtime_tests::index::test_kernel_index_scalar::<TestRuntime, FloatType>(
+                client,
+            );
+        }
+    };
+}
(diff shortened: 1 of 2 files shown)


crates/cubecl-wgpu/src/compiler/wgsl/instructions.rs
```diff
@@ -1012,51 +1035,47 @@ fn index(
...
+    };
+
+    if let Some(ind) = index {
+        if let Some(len) = len {
+            // Note: This is technically not 100% allowed. According to the WebGPU specification,
+            // any OOB access is a "dynamic error" which allows "many possible outcomes". In practice,
+            // both wgpu and Dawn handle this by either returning dummy data or clamping the index
+            // to valid bounds. This means it's harmless to use in a select.
+            let out_item = out.item();
+            value = format!("select({out_item}(0), {value}, {ind} < {len})");
```
nathanielsimard, line 1073: Not sure this is going to work with vectorization.

crates/cubecl-wgpu/src/compiler/wgsl/compiler.rs
```diff
@@ -720,267 +732,310 @@ impl WgslCompiler {
...
+                        false => cube::Metadata::Length { var: lhs },
+                    };
+
+                    instructions.push(self.compile_metadata(length, Some(array_len)));
+                    instructions.push(wgsl::Instruction::CheckedIndex {
+                        len: self.compile_variable(array_len),
+                        lhs: self.compile_variable(lhs),
+                        rhs: self.compile_variable(rhs),
+                        out: self.compile_variable(out),
+                    });
```
nathanielsimard, line 899: I don't think it's necessary right now, but I think having this implemented with a `cpa` kernel like `CheckedIndexAssign` would be easier.

nathanielsimard: I'm also not sure when we don't have `BufferLength`

crates/cubecl-wgpu/tests/unary_bench.wgsl
```diff
@@ -33,22 +33,46 @@ for (var l_2_2: u32 = 0u; l_2_2 < 256u; l_2_2++) {
...
+}
 } else {
-let _11 = input_0_global[id];
-let _12 = input_1_global[id];
+var l_3_0: u32;
+l_3_0 = info[0u];
+let _11 = select(vec4<f32>(0), input_0_global[id], id < l_3_0);
+var l_3_1: u32;
+l_3_1 = info[1u];
+let _12 = select(vec4<f32>(0), input_1_global[id], id < l_3_1);
```
nathanielsimard, line 61: I didn't know `vec4<f32>(0)` was supported syntax!

crates/cubecl-wgpu/src/compiler/wgsl/instructions.rs
ArthurBrussee: Just with the syntax of vec4(0)? Or more that indexing needs to take vectorization into account? I'm really not sure how vecorization works, if  you could add a test for it that'd be amazing!

crates/cubecl-wgpu/src/compiler/wgsl/compiler.rs
ArthurBrussee: +1 to cpa someday.

And yeah this behaviour is copied from the CPP version. Personally I feel like that kind of thing should fail to compile, rather than creating a language with strange edge cases of UB. I'm not sure what kind of situation means you don't have buffer length

nathanielsimard (approved):

PR #299, opened 2024-11-22
Improve tests for cubecl-reduce

maxtremblay:
I introduce new macros to run the tests on different vector, cube and line sizes.


louisfd (approved):
LGTM

PR #300, opened 2024-11-23
Fix hardware properties for WebGPU

ArthurBrussee:
- max_bindings was reading the wrong thing. That's the number of bind groups, not how many buffers you can have per kernel... On desktop this number was only 8, on WebGPU it's only 4. In reality on desktop it's >>>large, and on Chrome it's currently 10. Should improve fusion a bit on desktop, and a good amount on WebGPU.

- Min/max plane size were reported as dummy 0 values, which was causing issues. For now, disable subgroups until wgpu no longer reports dummy values here.


nathanielsimard (approved):
LGTM

PR #301, opened 2024-11-23
Make spirv feature to compile on macOS

syl20bnr:
Added some guidance in the `build.rs` file.

Lots of tests are failling though because of some device feature not present.

Also, not related to spir-v, there is one wgpu test failing `test_buffer_len_offset` with message:

> Buffer offset 5248 does not respect device's requested `min_storage_buffer_offset_alignment` limit 256

The offset 5248 changes each time we execute the test.

### DIFF
--- a/crates/cubecl-core/src/runtime_tests/metadata.rs
+++ b/crates/cubecl-core/src/runtime_tests/metadata.rs
@@ -256,14 +256,22 @@ macro_rules! testgen_metadata {
         }
 
         #[test]
-        fn test_buffer_len() {
+        fn test_buffer_len_discontiguous() {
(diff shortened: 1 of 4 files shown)


syl20bnr:
cc @wingertge

wingertge:
Does it mention the device feature somewhere? I'm enabling a bunch of features by default because as far as I can tell they should be supported on all devices, but the compatibility layer might not have them all available. The main enforced features are the Vulkan Memory model (that one seems the most likely to be missing from a compat layer but also the hardest to remove), f16/i8 support and 16 bit buffer support. CMMA is gated behind a check for device capability so it's not that.

PR #302, opened 2024-11-23
Kernel names with float suffix

torsteingrindvik:
Makes kernel entrypoint naming more powerful:

image

A suffix is added to the kernel name if the kernel has a `X: Float` bound (the generic name is irrelevant, only the bound is important).

Strategy:

- When parsing the kernel, inspect all generics
- Look for a generic type with the Float bound
- Use `core::any::type_name::<X>` where X is the generic type with the correct bound
- Clean up that string in order to have a short suffix e.g. `_f16` or `_f32`
- If no generic had the Float bound, use the base name with no suffix

This allows very clean reporting in tools such as Nsight Compute when testing kernels which are float generic.

I intentionally did not attempt to inspect generics in the `where X: Float` position as I don't foresee myself using that- let me know if this is important.

### DIFF
--- a/crates/cubecl-macros/src/generate/kernel.rs
+++ b/crates/cubecl-macros/src/generate/kernel.rs
@@ -1,7 +1,7 @@
 use darling::usage::{CollectLifetimes as _, CollectTypeParams as _, GenericsExt as _, Purpose};
 use proc_macro2::TokenStream;
 use quote::{format_ident, quote, ToTokens};
-use syn::Ident;
+use syn::{Ident, TypeParamBound};
 
 use crate::{
     parse::kernel::{KernelBody, KernelFn, KernelParam, KernelReturns, KernelSignature, Launch},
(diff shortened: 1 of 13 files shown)


crates/cubecl-macros/src/generate/kernel.rs
```diff
@@ -213,6 +213,77 @@ impl Launch {
...
+
+                            let type_name = core::any::type_name::< #type_param_ident >();
+                            let type_name = shorten(type_name);
+
+                            format!("{}_{type_name}", #base_name)
+                        }
+                    );
+                }
+            }
+        }
```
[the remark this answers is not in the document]
torsteingrindvik: I added support for `Numeric`. 
If multiple generic params exist I now just append the suffixes in order of appearance.

I added a WGSL test and a CUDA test to ensure this works, see `naming.cu` and `naming.wgsl`.

Also I discovered the CUDA tests don't run in CI, so I re-generated them such that they now pass. For example `#include <mma.h>` isn't on main and a few other things.

crates/cubecl-macros/src/generate/kernel.rs
```diff
@@ -213,6 +213,88 @@ impl Launch {
         }
     }
 
+    /// Returns the kernel entrypoint name.
+    /// Appropriate for usage in source code such as namiing the CUDA or WGSL entrypoint.
```
nathanielsimard, line 217: typo `namiing` => `naming`

torsteingrindvik: Ah whoops.
Done :+1:

torsteingrindvik: Tests are good again.

PR #303, opened 2024-11-24
Fix/bool cast

wingertge:
Fixes the boolean casting in SPIR-V to match other backends in treating all non-zero values as `true`, instead of just `1`.
Fixes `plane_all` and `plane_any` on SPIR-V by unrolling vectorized values, since the SPIR-V instruction only supports scalar output.
Fixes various test failures resulting from other changes.


nathanielsimard (approved):
LGTM

nathanielsimard:
@wingertge After fixing the merge conflicts we can merge!

PR #309, opened 2024-11-26
[Feat] Ground work to make GEMM components usable for convolution

wingertge:
Adds a new `AccumulatorLoader` trait and methods to load the accumulator, and makes a few type public that used to be private, so they can be used in `burn`.


louisfd (approved):
LGTM

PR #310, opened 2024-11-26
Matmul: tilewise loader

louisfd:
Addition of the tilewise loader, which existed in cmma old but was never ported. It needs the number of planes to be the same as the number of tiles. Possible to have it only on lhs or rhs, or both, or none. 

Combined with renaming of width/height


nathanielsimard (approved):
LGTM

PR #311, opened 2024-11-26
Cast fragment

nathanielsimard:


crates/cubecl-core/src/runtime_tests/cmma.rs
```diff
@@ -151,16 +174,53 @@ pub fn test_simple_1<R: Runtime>(
...
+}
+
 pub fn test_simple_tf32<R: Runtime>(
     client: ComputeClient<R::Server, R::Channel>,
     cube_dimensions: CubeDim,
 ) {
     if !client.properties().feature_enabled(Feature::Cmma {
-        a: Elem::Float(FloatKind::TF32),
-        b: Elem::Float(FloatKind::TF32),
+        a: Elem::Float(FloatKind::F16),
```
louisfd, line 219: Isn't this a test specifically for TF32?

crates/cubecl-cpp/src/cuda/wmma/cuda_compiler.rs
```diff
@@ -59,8 +59,7 @@ impl WmmaCompiler<CudaDialect<Self>> for CudaWmmaCompiler {
     fn supported_wmma_combinations(arch: &Self::Architecture) -> SupportedWmmaCombinations {
         let mut result: SupportedWmmaCombinations = vec![];
         if arch.version >= WMMA_MINIMUM_VERSION {
-            //                                   m   n   k
-            let tdims = vec![(16, 16, 16), (32, 16, 8), (8, 16, 32)];
+            let tdims = vec![(16, 16, 16), (32, 16, 8), (8, 16, 32), (32, 8, 16)];
```
louisfd, line 62: I think it should be [(16, 16, 16), (32, 8, 16), (8, 32, 16)]

louisfd (changes requested):

PR #312, opened 2024-11-27
Matmul: kernel select heuristic

louisfd:
I added heuristics to determine which kernel to launch. 
- Choose between 16x16x16, 32x8x16 and 8x32x16 tile instructions
- Choose stage dims as a function of input size vs tile size, number of tensor cores expected and number of SM (streaming multiprocessor) expected on the device. Last two are hardcoded to something average for now. The heuristic tries to have around as many cubes as SMs, then the largest amount of tensor cores. For large inputs, it will fix the number of tensor cores to its max and potentially way more cubes than SMs. But for smaller inputs it may choose to use less tensor cores to maximize SM occupancy. 

There are also some refactors around launching and advanced configs.

### DIFF
--- a/crates/cubecl-core/src/frontend/cmma.rs
+++ b/crates/cubecl-core/src/frontend/cmma.rs
@@ -299,9 +299,9 @@ pub mod load {
 /// Load the matrix with the provided array using the stride with an explicit layout.
 /// Explicit layouts are required when loading accumulators.
 #[allow(unused_variables)]
-pub fn load_with_layout<C: CubeType>(
+pub fn load_with_layout<C: CubePrimitive, V: CubePrimitive>(
     mat: &Matrix<C>,
-    value: &Slice<C>,
+    value: &Slice<V>,
     stride: u32,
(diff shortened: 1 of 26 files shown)


crates/cubecl-linalg/src/matmul/kernels/matmul/algorithm/selector.rs
```diff
@@ -0,0 +1,286 @@
...
+                    R,
+                    EG,
+                    StandardAlgorithm<EG, ES, EA, S8x8x2, Accelerated8x32x16<ES, EA>>,
+                >(client, lhs, rhs, out, problem),
+                _ => panic!("No configuration found for this stage size. "),
+            },
+            _ => panic!("No configuration found for instruction shapes."),
+        }
+    }
+}
```
nathanielsimard, line 122: Maybe this can be macros, it's pattern matching so that might be cleaner in a way, very optional.

nathanielsimard (approved):

PR #314, opened 2024-11-27
Import reduce naive from burn

maxtremblay:
This is the first PR to import the reduction algorithms implemented by @wingertge in Burn. 

Currently, this only import the naive implementation. The others will follow soon. I also added a bunch of tests, support for lines and refactor the code a little bit. However, the logic of the implementation follow the one from Burn.

Notice that the tests currently ignore ArgMax and ArgMin since I have some issues with Line that most be solved first.


crates/cubecl-reduce/src/test.rs
```diff
@@ -33,212 +35,392 @@ macro_rules! testgen_reduce {
...
+            .iter()
+            .map(|s| match s.cmp(&dim_stride) {
+                std::cmp::Ordering::Equal => 1,
+                std::cmp::Ordering::Greater => s / dim_shape,
+                std::cmp::Ordering::Less => *s,
+            })
+            .collect()
+    }
+
+    fn random_input_values<F: Float>(&self) -> Vec<F> {
```
louisfd, line 410: Quite a duplication from matmul test_utils. This would be a good case for cubecl-std, under a test cfg. Also, the matmul one has a seed as argument which can prevent bug proneness when we have several inputs. I think we should have an Option<u32> and use 123456789 when None. Anyway, not a blocker for this PR.

louisfd (approved):
I have a few comments, but it's already good

crates/cubecl-reduce/src/naive.rs
```diff
@@ -0,0 +1,213 @@
...
+    }
+
+    // Compute the first index where to start the reduction for the current thread.
+    // First, compute the coordinate corresponding to the ABSOLUTE_POS element of the output tensor
+    // Then, use the strides of the input tensor to find the index of the same coordinate
+    // in the input tensor.
+    let mut offset_input = 0;
+    for axis in 0..input.rank() {
+        let coordinate = (ABSOLUTE_POS / output.stride(axis)) % output.shape(axis);
+        offset_input += coordinate * input.stride(axis);
```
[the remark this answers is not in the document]
maxtremblay: The shape of the `dim` axis in output is always 1, so for that particular axis, the coordinate is always 0. Thus, it has a null effect on `offset_input`.

crates/cubecl-reduce/src/test.rs
maxtremblay: I can create an issue?

crates/cubecl-reduce/src/naive.rs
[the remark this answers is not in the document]
louisfd: No, I was just making sure nothing was lost from the original

crates/cubecl-reduce/src/test.rs
louisfd: Yeah, maybe an issue about creating cubecl-std with common stuff between reduce and matmul.

PR #316, opened 2024-11-27
Matmul/auto

nathanielsimard:
Revisited the matmul error handling to allow changes in the implementation based on errors. This enables the `auto` strategy, which first tries `accelerate` and then falls back to `tiling2d`.

### DIFF
--- a/crates/cubecl-linalg/src/matmul/components/problem.rs
+++ b/crates/cubecl-linalg/src/matmul/components/problem.rs
@@ -98,7 +98,7 @@ impl MatmulProblem {
             }
         }
 
-        if self.n % self.out_line_size as usize == 1 {
+        if self.n % self.out_line_size as usize != 0 {
             return Err(MatmulInvalidProblem::InvalidLineSizeOut {
                 size: self.n as u32,
                 line_size: self.out_line_size,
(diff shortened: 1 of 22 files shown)


crates/cubecl-linalg/src/matmul/components/problem.rs
```diff
@@ -36,36 +38,73 @@ impl MatmulProblem {
...
+                    });
+                }
+            }
+        }
 
-        assert!(match self.rhs_layout {
-            MatrixLayout::RowMajor => self.n % self.rhs_line_size as usize == 0,
-            MatrixLayout::ColMajor => self.k % self.rhs_line_size as usize == 0,
-        });
+        if self.n % self.out_line_size as usize == 1 {
```
louisfd, line 101: Should keep != 0 because it could be 2 or 3...

crates/cubecl-linalg/src/matmul/kernels/error.rs
```diff
@@ -0,0 +1,127 @@
...
+                f,
+                "the lhs tensor can't be read with line size={line_size} and dimension={size}"
+            ),
+            MatmulInvalidProblem::InvalidLineSizeRhs { size, line_size } => write!(
+                f,
+                "The rhs tensor can't be read with line size={line_size} and dimension={size}"
+            ),
+            MatmulInvalidProblem::InvalidLineSizeOut { size, line_size } => write!(
+                f,
+                "The out tensor can't be read with line size={line_size} and dimension={size}"
```
louisfd, line 94: Written not read

louisfd (changes requested):

PR #317, opened 2024-11-28
Add Support for cast instruction in hip wmma intrinsic compiler

syl20bnr:


syl20bnr:
Converting to draft, I want to fix other things that are not quite right in the intrinsic compiler.

crates/cubecl-cpp/src/hip/wmma/intrinsic_compiler.rs
```diff
@@ -51,7 +52,10 @@ impl WmmaCompiler<HipDialect<Self>> for WmmaIntrinsicCompiler {
     ) -> std::fmt::Result {
         match fragment.ident {
             FragmentIdent::A | FragmentIdent::B => write!(f, "half16"),
-            FragmentIdent::Accumulator => write!(f, "float8"),
+            FragmentIdent::Accumulator => match fragment.elem {
```
nathanielsimard, line 55: Can we use `bfloat16` in the instruction?

nathanielsimard (approved):
Nice docs

PR #319, opened 2024-11-28
Fix CMMA stride on SPIR-V

wingertge:
SPIR-V treats strides as the number of *vectorized* elements, while CUDA bases the stride off the scalar element. This means SPIR-V strides were off by a factor of `line_size`. This checks for vectorized values/outputs and adjusts the stride accordingly.

In the future we should consider whether using the lined size for strides makes more sense since it has the same semantics as indexing and length operations, which currently work in terms of number of lines, not number of scalar elements.

## Testing

All CMMA tests now pass again, and performance is as expected for SPIR-V, no longer 3-5x slower than CUDA.


nathanielsimard (approved):
LGTM

PR #327, opened 2024-11-29
Fix xtask command with last version

syl20bnr:
Fix custom commands with latest tracel-xtask version `1.1.7`


syl20bnr:
Although the tests fail they are not related to this change, I tested with the previous version.

PR #329, opened 2024-11-29
Import reduce shared

maxtremblay:
- Import the shared reduction from Burn.
- Add support for Line operation.
- Implement a bunch of tests (also update the test framework to be robust against the small precision of bf16 and f16)
- Impose that ArgMax and ArgMin always returns the smallest coordinate in case of item equality.

### DIFF
--- a/crates/cubecl-reduce/src/shared.rs
+++ b/crates/cubecl-reduce/src/shared.rs
@@ -3,21 +3,34 @@ use cubecl_core::prelude::*;
 
 use crate::{ReduceArgMax, ReduceArgMin, ReduceMean, ReduceProd, ReduceSum};
 
+// /// A naive implementation of the reduction algorithm.
+// ///
+// /// Each thread with absolute position P is responsible
+// /// to compute the reduction corresponding to index P of the output.
 /// An instruction for the [reduce_shared](reduce_shared) algorithm.
 #[cube]
 pub trait ReduceSharedInstruction<EI: Numeric>: Send + Sync + 'static {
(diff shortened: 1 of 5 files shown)


crates/cubecl-reduce/src/shared.rs
```diff
@@ -0,0 +1,408 @@
...
+            let destination = jump * 2 * UNIT_POS;
+            let origin = jump * (2 * UNIT_POS + 1);
+            if UNIT_POS < num_active_units {
+                RD::merge(&mut accumulator, destination, origin);
+            }
+            jump *= 2;
+            sync_units();
+        }
+    } else {
+        // 25
```
louisfd, line 114: Maybe remove this 25 and the 2* UP

crates/cubecl-reduce/src/shared.rs
```diff
@@ -0,0 +1,408 @@
...
+        }
+    }
+
+    if UNIT_POS == 0 {
+        RD::write_first(output, &accumulator, CUBE_POS, input.shape(reduce_dim));
+    }
+}
+
+#[cube]
+fn div_ceil(a: u32, b: u32) -> u32 {
```
louisfd, line 135: std material

louisfd (approved):
LGTM, just some comments to cleanup

crates/cubecl-reduce/src/shared.rs
maxtremblay: Ha yes. They were notes for me to follow the algorithm :)

PR #335, opened 2024-12-02
Implement more robust memory pool lookup (#243)

### LINKED ISSUE #243: `MemoryManagement<Storage>::reserve()` can panic with index out of bounds
`cubecl_runtime::memory_management::memory_manage::MemoryManagement<Storage>::reserve()` (below) can panic when `size` is bigger than any available `pool`.

https://github.com/tracel-ai/cubecl/blob/3882ed25b47506d49562c501a179b7468e61702e/crates/cubecl-runtime/src/memory_management/memory_manage.rs#L275-L288

In my case (running `burn`) my program attempted to allocate 2_925_527_040 bytes with pools max sizes: [0] 4_028_128 B [1] 4_028_128 B [2] 32_225_024 [3] <my debugger failed to read max_size> [4] 2_062_401_536 B. So `pool_ind` was computed as `5` and `&mut self.pools[pool_ind]` panicked.

The function can panic in edge-cases because per docs `partition_point()` requires that:

> (...) all elements for which the predicate returns true are at the start of the slice and all elements for which the predicate returns false are at the end
(...)
If this slice is not partitioned, the returned result is unspecified and meaningless, as this method performs a kind of binary search.

and the condition seems to not be checked at any point.

The same issue also occurs a few lines lower in `alloc()`.

The solution is probably a simple bounds check or capping `pool_ind` to `pools.len() - 1` or some more complex error handling.

Happy to help with solving the problem.

Quba1:
Fixes #243 

I have implemented a more robust way to find big enough memory pool for reserve/alloc. I think comments explain quite well what happens in the code.

<details>
  <summary>Minimal example of the solution</summary>

```rust
pub fn find_mem_pool(pool: &[usize], size: usize) -> usize {
    let ind = pool
        .binary_search(&size)
        .or_else(|found_index| {
            if found_index < pool.len() {
                Ok(found_index)
            } else {
                Err(found_index)
            }
        })
        .unwrap_or_else(|_| panic!("Cannot find pool big enough to reserve {size} bytes"));

    ind
}

#[cfg(test)]
mod tests {
    use crate::find_mem_pool;

    #[test]
    fn partitioning() {
        let v = [1, 3, 3, 5, 7];

        assert_eq!(find_mem_pool(&v, 0), 0);
        assert_eq!(find_mem_pool(&v, 1), 0);
        assert_eq!(find_mem_pool(&v, 2), 1);
        assert_eq!(find_mem_pool(&v, 3), 2);
        assert_eq!(find_mem_pool(&v, 5), 3);
        assert_eq!(find_mem_pool(&v, 6), 4);
        assert_eq!(find_mem_pool(&v, 7), 4);

        // this panics
        find_mem_pool(&v, 8);
    }
}
```
</details>

I have also added a simple test to check if the code panics when too much memory is requested.

To avoid code repetition in `reserve()` and `alloc()` I've extracted the solution into a separate function. But because `alloc` and `reserve` require `&mut Storage` the finder function needs to be implemented for directly for `pools` (to have two mutable borrows for pools and storage). But as `Vec<DynamicPool>` is a foreign type I had to implement the function via trait, hence the addition of `PoolFind` trait.

I hope this PR helps you. If something needs improvements or corrections, please let me know.

### DIFF
--- a/crates/cubecl-runtime/src/memory_management/memory_manage.rs
+++ b/crates/cubecl-runtime/src/memory_management/memory_manage.rs
@@ -278,20 +278,9 @@ impl<Storage: ComputeStorage> MemoryManagement<Storage> {
         // hard about overflow here.
         self.alloc_reserve_count += 1;
 
-        // Find first pool where size <= p.max_alloc with a binary search.
-        let pool_ind = self.pools.partition_point(|p| size > p.max_alloc_size());
-
-        // Ensure the pool index is in bounds, otherwise there isn't any pool that can fit the
-        // requested allocation
-        if pool_ind >= self.pools.len() {
-            panic!("Unable to find valid pool partition point: No memory pool big enough to reserve {size} bytes.");
-        }
-
-        let pool = &mut self.pools[pool_ind];
-        if pool.max_alloc_size() < size {
-            panic!("No memory pool big enough to reserve {size} bytes.");
-        }
-        pool.reserve(&mut self.storage, size, exclude)
+        self.pools
+            .find_suitable_memory_pool(size)
+            .reserve(&mut self.storage, size, exclude)
     }
 
     /// Bypass the memory allocation algorithm to allocate data directly.
@@ -300,13 +289,9 @@ impl<Storage: ComputeStorage> MemoryManagement<Storage> {
     ///
     /// Can be useful for servers that want specific control over memory.
     pub fn alloc(&mut self, size: u64) -> SliceHandle {
-        // Find first pool where size <= p.max_alloc with a binary search.
-        let pool_ind = self.pools.partition_point(|p| size > p.max_alloc_size());
-        let pool = &mut self.pools[pool_ind];
-        if pool.max_alloc_size() < size {
-            panic!("No memory pool big enough to alloc {size} bytes.");
-        }
-        pool.alloc(&mut self.storage, size)
+        self.pools
+            .find_suitable_memory_pool(size)
+            .alloc(&mut self.storage, size)
     }
 
     /// Bypass the memory allocation algorithm to deallocate data directly.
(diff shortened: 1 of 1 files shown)


crates/cubecl-runtime/src/memory_management/memory_manage.rs
```diff
@@ -352,6 +337,45 @@ impl<Storage: ComputeStorage> MemoryManagement<Storage> {
...
+    /// Finds and returns a memory pool big enough for the requested size.
+    ///
+    /// # Panics
+    ///
+    /// Panics if no pool big enough can be found.
+    fn find_suitable_memory_pool(self, size: u64) -> &'a mut DynamicPool {
+        let pool_ind = self
+            // This finds a (not necessarily first) pool that matches the size exactly
+            // or returns error with index where a matching element could be inserted while maintaining sorted order
+            .binary_search_by_key(&size, |p| p.max_alloc_size())
```
nathanielsimard, line 355: `size` won't be equal to any of the `max_alloc_sizes`, we should find a bucket where `size >= min_slice_size && size <= max_alloc_size`. We don't need binary search since we only have < 10 buckets, max 20. Having the trait is great though.

Quba1: Sorry, I didn't know about the `size >= min_slice_size` condition. But I also can't find `min_slice_size` defined anywhere.

Is the `Vec<DynamicPools>` sorted by `min_slice_size` too? Because then binary search would still be possible. 

Binary search was what `partition_point()` was doing and this implementation only extends that technique. But with such a small array size I do agree that linear search will not impact the performance and will allow for the `size >= min_slice_size` condition - I'll change the implementation.

ArthurBrussee: Hiya - give I wrote this I'll barge in - there is no min_slice_size, that's implicit by the fact that that should have been allocated in the previous pool. Stuff being sorted is therefore a condition anyway - and binary search seems fine? I agree it's a small performance win but... idk, seems clearer & faster. I just didn't quite handle the error condition of the search correctly.

The trait also seems like it could just be a struct? Doubt you'd want multiple possible implementations of this.

Quba1: In case there's only one condition, then the most important consideration for using binary search is that if there are two pools with same `max_alloc_size` then it's not predictable which one will be chosen. The question is if that's an acceptable behaviour.

> The trait also seems like it could just be a struct? Doubt you'd want multiple possible implementations of this.

Struct or newtype would require changing the type of `pools` in `MemoryManagement` and I didn't feel comfortable doing that. So I created the trait, which I admit is a little bit excessive, but necessary for borrow checker.

Quba1:
Hi, just checking up. Is this PR still needed, is it under consideration, should I change something or should it be closed?

ArthurBrussee:
Hiya! Ah I think after #419 and #445 this has been superseded, sorry about that, thank you for the PR though :)

PR #336, opened 2024-12-02
Implement a plane reduction

maxtremblay:
For this reduction algorithm, each plane is responsible of reducing a single tensor element.


louisfd (approved):
I have only a couple of minor comments. Otherwise very good!

PR #349, opened 2024-12-05
major refactor of reduce

maxtremblay:


crates/cubecl-reduce/src/instructions/argmax.rs
```diff
@@ -0,0 +1,153 @@
...
+        accumulator.1 = new_coordinate;
+    }
+
+    fn merge_line<Out: Numeric>(accumulator: Self::Accumulator, _shape_axis_reduce: u32) -> Out {
+        let line_size = accumulator.0.size();
+        let mut max = In::MIN.runtime();
+        let mut coordinate = 0;
+        #[unroll]
+        for k in 0..line_size {
+            if accumulator.0[k] == max && accumulator.1[k] < coordinate {
```
louisfd, line 74: Maybe we should extract accumulator.0[k] and accumulator.1[k] in variables to reuse them?

crates/cubecl-reduce/src/lib.rs
```diff
@@ -1,12 +1,44 @@
+mod config;
+mod error;
 mod instructions;
-mod naive;
-mod plane;
-mod shared;
+mod launch;
+mod strategy;
+// mod naive;
```
louisfd, line 6: Dead code

louisfd (changes requested):
You have a lot of dead code, also minor comments

crates/cubecl-core/src/frontend/container/slice.rs
```diff
@@ -131,7 +131,7 @@ mod indexation {
 
     use super::*;
 
-    impl<E: CubePrimitive> Slice<E> {
+    impl<E: CubeType> Slice<E> {
```
louisfd, line 134: Should not change this

PR #363, opened 2024-12-10
Reduce shared unit

maxtremblay:
Implement the reduction when `shared = true` and `use_planes = false`.

### DIFF
--- a/crates/cubecl-reduce/src/launch.rs
+++ b/crates/cubecl-reduce/src/launch.rs
@@ -1,21 +1,24 @@
 use cubecl_core as cubecl;
 use cubecl_core::prelude::*;
 
-use crate::{LineMode, Reduce, ReduceConfig, ReduceStrategy, SharedAccumulator};
+use crate::{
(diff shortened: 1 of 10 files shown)


crates/cubecl-reduce/src/launch.rs
```diff
@@ -221,20 +330,128 @@ pub fn reduce_slice_plane<N: Numeric, Instr: Reduce<N>>(
...
+/// ```
+///
+/// The outcome is stored in the first element of the accumulator and also returned by this function for convenience.
+///
+/// This assumes that the size of accumulator is exactly CUBE_DIM and leads to undefined behavior if not.
+#[cube]
+pub fn reduce_tree<N: Numeric, Inst: Reduce<N>>(
+    accumulator: &mut Inst::SharedAccumulator,
+) -> Inst::AccumulatorItem {
+    if comptime!(CUBE_DIM.is_power_of_two()) {
```
louisfd, line 416: The if and else clause have gotten quite similar, do you think we could refactor to mitigate duplication?

louisfd (approved):
LGTM, i have only one optional remark.

PR #368, opened 2024-12-12
[Feat] Expanded SPIR-V debug info

wingertge:
Adds line and column numbers as well as proper file names to SPIR-V debug info using the new `DebugInfo::Source` and `DebugInfo::Line` instructions, and a wrapper function to prepend them to semantically relevant instructions.


crates/cubecl-core/src/frontend/debug.rs
```diff
@@ -23,6 +28,37 @@ pub fn debug_call_expand<C>(
...
+        call(context)
+    }
+}
+
+/// Adds source instruction if debug is enabled
+#[track_caller]
+pub fn debug_source_expand(context: &mut CubeContext, name: &str, file: &str, line: u32, col: u32) {
+    if context.debug_enabled {
+        // Normalize to linux separators
+        let file = file.replace("\\", "/");
```
nathanielsimard, line 52: Can we create a path buf instead?

nathanielsimard (approved):
LGTM

crates/cubecl-core/src/frontend/debug.rs
wingertge: I looked this up and apparently Rust has no way to normalize path separators on pathbuf to a specific one, it can only canonicalize a path to an absolute one for the current system (with the current system's separators

PR #380, opened 2024-12-17
Feat/ virtual tensor

nathanielsimard:
Added a virtual tensor type so that we don't have to carry the `MatmulArgs` trait in all matmul components, since it makes it really really hard to reuse those components in other kernels that don't have the same arguments.

### DIFF
--- a/crates/cubecl-linalg/src/matmul/components/batch/shared.rs
+++ b/crates/cubecl-linalg/src/matmul/components/batch/shared.rs
@@ -3,6 +3,7 @@ use cubecl_core::prelude::*;
 
 use crate::matmul::components::global::args::{TensorInput, TensorOutput};
 use crate::matmul::components::{global, MatmulSpec};
+use crate::tensor::{ReadWrite, VirtualTensor};
 
 #[cube]
 /// Execute global matmul on lhs, rhs, writing in out.
@@ -10,7 +11,7 @@ use crate::matmul::components::{global, MatmulSpec};
 pub(crate) fn gmm_execute<MS: MatmulSpec, GMM: global::Matmul<MS>>(
     lhs: TensorInput<MS::EG, MS::Args>,
     rhs: TensorInput<MS::EG, MS::Args>,
-    out: TensorOutput<MS::EG, MS::Args>,
+    mut out: TensorOutput<MS::EG, MS::Args>,
     x_offset: u32,
     y_offset: u32,
     nth_batch: u32,
@@ -27,6 +28,10 @@ pub(crate) fn gmm_execute<MS: MatmulSpec, GMM: global::Matmul<MS>>(
         batch_lhs += tmp % lhs.shape(b) * lhs.stride(b);
         batch_rhs += tmp % rhs.shape(b) * rhs.stride(b);
     }
+
+    let lhs = VirtualTensor::<MS::EG>::new::<TensorInput<MS::EG, MS::Args>>(&lhs);
+    let rhs = VirtualTensor::<MS::EG>::new::<TensorInput<MS::EG, MS::Args>>(&rhs);
+    let out = VirtualTensor::<MS::EG, ReadWrite>::new::<TensorOutput<MS::EG, MS::Args>>(&mut out);
 
     GMM::execute(
         GMM::init_lhs_loader(lhs, x_offset, k_range.0, batch_lhs, config),
--- a/crates/cubecl-linalg/src/matmul/components/global/args.rs
+++ b/crates/cubecl-linalg/src/matmul/components/global/args.rs
@@ -1,5 +1,6 @@
+use crate::tensor::{VirtualTensorOperations, VirtualTensorOperationsExpand};
 use cubecl::prelude::*;
-use cubecl_core as cubecl;
+use cubecl_core::{self as cubecl};
 
 #[cube]
 /// Arguments for the matrix multiplication algorithm.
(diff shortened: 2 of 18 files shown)


louisfd (approved):
LGTM

crates/cubecl-linalg/src/matmul/components/global/args.rs
```diff
@@ -1,5 +1,6 @@
+use crate::tensor::{VirtualTensorOperations, VirtualTensorOperationsExpand};
 use cubecl::prelude::*;
-use cubecl_core as cubecl;
+use cubecl_core::{self as cubecl};
```
maxtremblay, line 3: I don't know why you changed that, but the previous style is more consistent with the rest of the code base.

crates/cubecl-linalg/src/matmul/components/global/args.rs
```diff
@@ -1,5 +1,6 @@
...
 
 #[cube]
```
maxtremblay, line 5: Now that we have the new VirtualTensor trait, it looks like we can replace this with a struct. Something like 

```rust
pub struct MatmulArgs<L, R, O> 
where 
   L: VirtualTensor,
   R: VirtualTensor,
   O: VirtualTensor
{
   lhs: L,
   rhs: R,
   out: O
}
```

This seems even more relevant as there is only a single implementation of `MatmulArgs`

crates/cubecl-linalg/src/matmul/components/global/args.rs
```diff
@@ -60,6 +61,91 @@ pub struct TensorInput<EG: Numeric, GA: MatmulArgs<EG>> {
     ident: TensorInputIdent,
 }
 
+impl<EG: Numeric, MA: MatmulArgs<EG>> VirtualTensorOperations<EG> for TensorInput<EG, MA> {}
+impl<EG: Numeric, MA: MatmulArgs<EG>> VirtualTensorOperations<EG> for TensorOutput<EG, MA> {}
+
+impl<EG: Numeric, MA: MatmulArgs<EG>> VirtualTensorOperationsExpand<EG>
+    for TensorOutputExpand<EG, MA>
```
maxtremblay, line 68: The ordering of `impl` blocks in this file is really inconsistent. I suggest either implementing everything related `TensorInput` than everything related to `TensorOuput` or to implement similar functionalities for both consecutively for each functionality. But not intertwined like it currently is.

crates/cubecl-linalg/src/matmul/components/batch/shared.rs
```diff
@@ -28,6 +29,10 @@ pub(crate) fn gmm_execute<MS: MatmulSpec, GMM: global::Matmul<MS>>(
         batch_rhs += tmp % rhs.shape(b) * rhs.stride(b);
     }
 
+    let lhs = VirtualTensor::<MS::EG>::new::<TensorInput<MS::EG, MS::Args>>(&lhs);
+    let rhs = VirtualTensor::<MS::EG>::new::<TensorInput<MS::EG, MS::Args>>(&rhs);
+    let out = VirtualTensor::<MS::EG, ReadWrite>::new::<TensorOutput<MS::EG, MS::Args>>(&mut out);
```
maxtremblay, line 34: Could be a good idea to implement a trait like the following to simplify the above.

```rust
pub trait IntoVirtual<E: Numeric, IO> {
    fn into_virtual(self) -> VirtualTensor<E, IO>;
} 
```

crates/cubecl-linalg/src/matmul/components/global/tensor_view.rs
```diff
@@ -25,8 +24,8 @@ pub struct TensorReader<GA: MatmulArgs<E>, E: Numeric> {
 /// A view of a tensor that starts reading data from a specified offset.
 /// Ensures safe access by preventing out-of-bounds errors.
 /// Includes pre-fetched shapes and strides for optimized performance.
-pub struct TensorWriter<GA: MatmulArgs<E>, E: Numeric> {
-    pub tensor: TensorOutput<E, GA>,
+pub struct TensorWriter<E: Numeric> {
+    pub tensor: VirtualTensor<E, ReadWrite>,
```
maxtremblay, line 28: To save a little bit of duplication, we could use the newly introduced IO concept and define a struct

```rust
pub struct TensorIO<E, IO> { }
```

with possibly type aliases to avoid updating the rest of the code base.

```rust
pub type TensorReader<E> = TensorIO<E, Read>;
pub type TensorWriter<E> = TensorIO<E, ReadWrite>;
```

crates/cubecl-linalg/src/matmul/components/global/base.rs
```diff
@@ -70,7 +69,7 @@ pub trait Matmul<MS: MatmulSpec>: 'static + Send + Sync + MatmulKernel<Config: C
 
     /// Initialize the unloader at row m and column n
     fn init_unloader(
-        out: TensorOutput<MS::EG, MS::Args>,
+        out: VirtualTensor<MS::EG, ReadWrite>,
         m_offset: u32,
         n_offset: u32,
         batch_offset: u32,
```
maxtremblay, line 75: I really enjoy that the IO capacity is explicit. In comparison to the previous version, I immediately that `out` is mutable

maxtremblay (changes requested):
Should we rename `VirtualTensor` to `TensorVirtual` to match the other patterns `TensorInput`, `TensorReader`, etc.

Issue #386, opened 2024-12-19
Bug in lined f16 code


torsteingrindvik:
![image](https://github.com/user-attachments/assets/7c74444d-29bd-460b-97fb-f21fc1a69dd1)

For a test such as the above f32 works.

However f16 produces this code:

![image](https://github.com/user-attachments/assets/d4f0cbf6-e175-4997-83cf-6e5569ecc1b8)

PR #387, opened 2024-12-19
CUDA clamp and lined clamp tests

torsteingrindvik:
When #386 is fixed, this PR can re-generate the currently wrong f16 code to avoid regressions.


louisfd:
Outdated

PR #390, opened 2024-12-31
Clean up IR slightly by merging comments into NonSemantic IR type

wingertge:
Cleans things up slightly by renaming `DebugInfo` to `NonSemantic` and moving `Comment` into it.


crates/cubecl-core/src/ir/non_semantic.rs
```diff
@@ -0,0 +1,52 @@
+use std::fmt::Display;
+
+use serde::{Deserialize, Serialize};
+
+use crate::ir::fmt_vararg;
+
+use super::Variable;
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+pub enum NonSemantic {
```
nathanielsimard, line 10: I would add some comments on the struct. Also would it be better to rename it to `SourceInfo` or `SourceMetadata` ?

wingertge: The reason I named it that is because printlns aren't really metadata, and comments are only metadata in a loose sense. I'll add a comment there though, I added it to the enum but forgot it here. The thing that unifies these operations is that they're not semantically relevant, aka they have no effect on the result of the kernel.

PR #393, opened 2025-01-01
Refactor optimizer for maintainability/extensability

wingertge:
Refactors optimizer to split analyses and optimization passes so passes can be stateless and analyses can be lazily run when needed, and invalidated where necessary.


crates/cubecl-linalg/src/matmul/tests/cmma_matmul/matmul_test_launcher.rs
```diff
@@ -104,6 +104,7 @@ pub fn test_matmul_algorithm<A, EG, ES, R>(
     if A::check_availability::<R, (EG, ES, f32)>(&client, &config).is_err() {
         // Can't execute the test.
         println!("Skipped - not supported!");
+        client.flush();
```
nathanielsimard, line 107: shouldn't be needed anymore, but isn't a big deal.

crates/cubecl-opt/src/analyses/base.rs
```diff
@@ -0,0 +1,62 @@
...
+/// and persist until they're invalidated.
+pub trait Analysis {
+    /// Perform the analysis for the current optimizer state and return the persistent analysis state
+    fn init(opt: &mut Optimizer) -> Self;
+}
+
+#[derive(Default, Clone, Debug)]
+pub struct Analyses {
+    cache: Rc<RefCell<TypeMap>>,
+}
```
nathanielsimard, line 23: It was hard for me to separate `Analysis` and `Analyses` visually while scanning code 😅. Would we rename `Analyses` to something like `AnalysisCache` or `AnalysisStore`?

crates/cubecl-opt/src/analyses/base.rs
```diff
@@ -0,0 +1,62 @@
...
+    pub fn invalidate_analysis<A: Analysis + Any>(&self) {
+        self.analyses.invalidate::<A>();
+    }
+
+    pub fn invalidate_structure(&self) {
+        self.invalidate_analysis::<PostOrder>();
+        self.invalidate_analysis::<Dominators>();
+        self.invalidate_analysis::<PostDominators>();
+        self.invalidate_analysis::<Liveness>();
+    }
```
nathanielsimard, line 61: Could have some basic docs. Like is there a reason why we fetch the anlysis from the optimizer using it's type?

nathanielsimard: I mean I like it, just some clarifications would be great!

crates/cubecl-opt/src/analyses/const_len.rs
```diff
@@ -0,0 +1,96 @@
...
+
+use super::Analysis;
+
+#[derive(Debug, Clone)]
+pub struct Slice {
+    pub start: Variable,
+    pub end: Variable,
+    pub end_op: Option<Operation>,
+    pub const_len: Option<u32>,
+}
```
nathanielsimard, line 15: If I understand correctly, the pub analysis type is actually `Slices`, is there a reason to have `Slice` public?

crates/cubecl-opt/src/analyses/integer_range.rs
```diff
@@ -1,39 +1,59 @@
...
+use crate::{Optimizer, VarId};
+
+use super::Analysis;
 
-use super::OptimizerPass;
+#[derive(Default, Clone, Copy, PartialEq, Eq, Debug)]
+pub struct Range {
+    pub lower_bound: Option<u64>,
+    pub upper_bound: Option<u64>,
+}
```
nathanielsimard, line 18: Same comment as `Slices` vs `Slice`

crates/cubecl-opt/src/lib.rs
```diff
@@ -243,24 +222,7 @@ impl Optimizer {
...
-        self.post_order.push(block);
-    }
-
-    pub fn post_order(&self) -> Vec<NodeIndex> {
-        self.post_order.clone()
-    }
-
-    pub fn reverse_post_order(&self) -> Vec<NodeIndex> {
-        self.post_order.iter().rev().copied().collect()
+        self.invalidate_structure();
```
nathanielsimard, line 225: Not sure what invalidate means in this scenario, a comment would be great!

crates/cubecl-opt/src/passes/find_safe_indexes.rs
```diff
@@ -0,0 +1,65 @@
...
+
+use super::OptimizerPass;
+
+/// Use the results from integer range analysis to find indexes that are always in bounds, then
+/// transform them to unchecked indexes.
+pub struct InBoundsToUnchecked;
+
+impl OptimizerPass for InBoundsToUnchecked {
+    fn apply_post_ssa(&mut self, opt: &mut Optimizer, changes: AtomicCounter) {
+        let ranges = opt.analysis::<Ranges>();
```
nathanielsimard, line 16: Yes it makes so much sense now! Optimization passes trigger analyses, which are cached!

nathanielsimard (commented):
I really like the refactor. The only thing I'm not getting is when and why we have to invalidate an analysis? I guess it's when the state has been changed by an optimization pass, so the analysis must be recalculated to account for the state change.

If this is the case, I think it would be great to have an event based system where Analysis can device if they are invalidated after a state changes, not the other way around (if it makes sense).

wingertge:
> I really like the refactor. The only thing I'm not getting is when and why we have to invalidate an analysis? I guess it's when the state has been changed by an optimization pass, so the analysis must be recalculated to account for the state change.
> 
> If this is the case, I think it would be great to have an event based system where analyses can decide if they are invalidated after a state changes, not the other way around (if it makes sense).

It would add a lot of complexity and doesn't really reduce coupling since analyses are often invalidated by very specific things. I did think of a sort of dependency system, but decided against it. For reference: LLVM also has each pass invalidate the relevant analyses when the state they rely on changes.
The main thing that causes invalidation is a change in the structure of the control flow graph, since that invalidates all node mappings that have been established by analyses. The other one is things like range analysis which can change from constant propagation (more ranges might be possible to pin down once the constants have been inserted). But those are usually not breaking, you just lose out of optimization opportunity.

crates/cubecl-linalg/src/matmul/tests/cmma_matmul/matmul_test_launcher.rs
wingertge: The PR for that isn't merged yet so I'd leave it for now so tests can actually be run.

crates/cubecl-opt/src/analyses/base.rs
wingertge: I was thinking the same thing but couldn't come up with a better name, `AnalysisCache` is probably best.

nathanielsimard:
> It would add a lot of complexity and doesn't really reduce coupling since analyses are often invalidated by very specific things. I did think of a sort of dependency system, but decided against it.

Thinking more about this and compute pass depends on the analysis, so if we want to avoid a circular dependency, compute pass should trigger analysis AND invalidate analysis, so the oposite mentionned in my comment.

crates/cubecl-opt/src/analyses/base.rs
wingertge: I added some basic docs, but I figured the type was obvious - different analyses provide different data and methods, and they're not defined by a single kind of access. So they need to be stored by their type ID and cast to the specific type. Analyses are always singletons so the type itself can serve as the key.

crates/cubecl-opt/src/analyses/integer_range.rs
wingertge: See the reply on slices.

wingertge:
> Thinking more about this and compute pass depends on the analysis, so if we want to avoid a circular dependency, compute pass should trigger analysis AND invalidate analysis, so the oposite mentionned in my comment.

I was thinking something like `depends_on: structure, constant_prop`. And then an invalidate method that takes a type to invalidate. But I realized it would just be overengineering something that could be handled by a simple list in a function like `invalidate_structure`.

nathanielsimard (approved):
Fixing the conflicts, then I think we can merge!

PR #394, opened 2025-01-03
Check for CUDA distribution installation.

vaijira:
In case CUDA is installed by distro packages in the installation check the path of CUDA compiler and return /usr/include path.


kpwebb:
Just +1 on this -- my CUDA install (12.4 on Ubuntu 22.04) was in `/usr/include` not `/usr/local/cuda` so search for mma.h fell down.

nathanielsimard (approved):

PR #401, opened 2025-01-08
[Feat] Rework allocator

wingertge:
Reworks allocator to carry a global state, with a reference owned by each scope. This allows for globally incrementing indices, preventing any potential name collisions entirely.
Removes `depth` from variables now that indices are already globally unique, which simplifies a lot of code in the optimizer.
Changes IDs to `u32` because there's no actual performance benefit to using `u16`, and `u32` gives a lot more headroom now that `depth` is gone. `u16` + `u8` would've gotten aligned to 32 bits regardless with the old IDs.
Replaces all of the uses for `u16` for IDs with the `Id` type alias that I introduced in the `Variable` refactor to make potential refactors easier in the future.

I currently don't change how variables are declared,  which leads to some slightly awkward code in `CubeContext` and `Scope`. I think eventually declaration should move out of each scope to the global allocator, but that would require larger changes to the backends that I wanted to avoid  for now.

## Testing
All tests pass now (except for the 16 CUDA tests that are expected to fail).

### DIFF
--- a/crates/cubecl-core/src/ir/allocator.rs
+++ b/crates/cubecl-core/src/ir/allocator.rs
@@ -41,15 +41,15 @@ impl PartialEq for Allocator {
 }
 
 impl Allocator {
-    /// Create a new immutable local variable of type specified by `item` for the given `scope`.
+    /// Create a new immutable local variable of type specified by `item`.
     pub fn create_local(&self, item: Item) -> ExpandElement {
         let id = self.new_local_index();
         let local = VariableKind::LocalConst { id };
         ExpandElement::Plain(Variable::new(local, item))
     }
 
-    /// Create a new mutable local variable of type specified by `item` for the given `scope`.
-    /// Try to reuse a previously defined but unused mutable variable in the current scope if possible.
+    /// Create a new mutable local variable of type specified by `item`.
+    /// Try to reuse a previously defined but unused mutable variable if possible.
     /// Else, this define a new variable.
     pub fn create_local_mut(&self, item: Item) -> ExpandElement {
         if item.elem.is_atomic() {
@@ -60,7 +60,7 @@ impl Allocator {
         }
     }
 
-    /// Create a new mutable restricted local variable of type specified by `item` into the given `scope`.
+    /// Create a new mutable restricted local variable of type specified by `item`.
     pub fn create_local_restricted(&self, item: Item) -> ExpandElement {
         let id = self.new_local_index();
         let local = VariableKind::LocalMut { id };
@@ -122,14 +122,5 @@ impl Allocator {
 
     pub fn new_local_index(&self) -> u32 {
         self.next_id.fetch_add(1, Ordering::Release)
-    }
-
-    pub fn managed_locals(&self) -> Vec<Variable> {
-        self.local_mut_pool
-            .borrow()
-            .values()
-            .flatten()
-            .map(|it| **it)
-            .collect()
     }
 }
--- a/crates/cubecl-core/src/ir/allocator.rs
+++ b/crates/cubecl-core/src/ir/allocator.rs
@@ -1,8 +1,13 @@
-use std::{collections::HashMap, rc::Rc};
+use std::{
+    cell::RefCell,
+    collections::HashMap,
+    rc::Rc,
+    sync::atomic::{AtomicU32, Ordering},
+};
 
 use crate::prelude::ExpandElement;
 
-use super::{Item, Scope, Variable};
+use super::{Item, Matrix, Variable, VariableKind};
 
 /// An allocator for local variables of a kernel.
 ///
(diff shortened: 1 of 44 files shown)


crates/cubecl-core/src/ir/scope.rs
```diff
@@ -96,77 +98,58 @@ impl Scope {
 
     /// Create a matrix variable
     pub fn create_matrix(&mut self, matrix: Matrix) -> Variable {
-        let index = self.matrices.len() as u16;
```
maxtremblay, line 99: This functions is never used. It is safe to remove it?

crates/cubecl-core/src/ir/scope.rs
```diff
@@ -96,77 +98,58 @@ impl Scope {
...
     }
 
+    pub fn add_matrix(&mut self, variable: Variable) {
+        self.matrices.push(variable);
+    }
+
     /// Create a slice variable
     pub fn create_slice(&mut self, item: Item) -> Variable {
-        let id = self.slices.len() as u16;
-        let variable = Variable::new(
```
maxtremblay, line 115: Same. This functions is never used.

crates/cubecl-opt/src/block.rs
```diff
@@ -20,9 +20,9 @@ pub struct BasicBlock {
     /// The phi nodes that are required to be generated at the start of this block.
     pub phi_nodes: Rc<RefCell<Vec<PhiInstruction>>>,
     /// The variables written to by this block. Only set during the SSA transformation.
-    pub(crate) writes: HashSet<(u16, u8)>,
+    pub(crate) writes: HashSet<Id>,
     /// The live variables at the start of this block. Used for pruning phi nodes.
-    pub(crate) live_vars: HashSet<(u16, u8)>,
+    pub(crate) live_vars: HashSet<Id>,
     /// The dominance frontiers of this block (where phi nodes must be inserted).
```
maxtremblay, line 26: Maybe not for this PR, but could we replace the HashSet by a single Id / u32? Since we are allocating all Id in increasing order, we only need to know that how many there are.

crates/cubecl-opt/src/lib.rs
```diff
@@ -107,13 +107,11 @@ pub struct ConstArray {
 #[derive(Default, Debug, Clone)]
 struct Program {
     pub const_arrays: Vec<ConstArray>,
-    pub variables: HashMap<(u16, u8), Item>,
-    pub(crate) slices: HashMap<(u16, u8), Slice>,
+    pub variables: HashMap<Id, Item>,
+    pub(crate) slices: HashMap<Id, Slice>,
```
maxtremblay, line 111: Do we need an hash map or we could use a vec instead? (really minor)

maxtremblay (commented):
Thanks for the PR. I have some minor comments, but let's merge that soon :fire:

crates/cubecl-core/src/ir/allocator.rs
```diff
@@ -76,14 +109,27 @@ impl Allocator {
...
+    }
+
+    pub fn managed_locals(&self) -> Vec<Variable> {
+        self.local_mut_pool
+            .borrow()
+            .values()
+            .flatten()
+            .map(|it| **it)
+            .collect()
+    }
```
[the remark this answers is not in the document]
wingertge: You're right, that's left over from trying to refactor the declarations, but I'll do that in a separate PR later when more pressing things are resolved

crates/cubecl-core/src/ir/kernel.rs
```diff
@@ -4,7 +4,7 @@ use serde::{Deserialize, Serialize};
 use std::fmt::Display;
 use std::num::NonZero;
 
-#[derive(Debug, Clone, Serialize, Deserialize)]
+#[derive(Debug, Clone)]
 #[allow(missing_docs)]
```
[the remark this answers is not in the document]
wingertge: Meant to put that back, guess I missed a file when reverting that.

crates/cubecl-opt/src/block.rs
wingertge: no, this is a list of variables that are currently live (their value is still needed), not all variables. So it's not contiguous. Same for writes, it's only variables written to in that block.

crates/cubecl-opt/src/lib.rs
wingertge: I think it's better as a hashmap because it only tracks mutable variables/slices, so a vec would be extremely sparse. most variables will be immutable and not slices.

crates/cubecl-core/src/ir/scope.rs
wingertge: Probably, I think it's just there for legacy support with `cpa!`. Don't think there's really a point in keeping it though, since cpa kernels are deprecated and I doubt we'll need this for anything like bounds checking.

maxtremblay (approved):

PR #403, opened 2025-01-08
Add eye support for TensorHandle.

vaijira:


nathanielsimard (changes requested):
I would call the function `identify` instead of `eye`. Also, tensor handle shouldn't be a tensor object like the Tensor struct in Burn. So, for consistency, I would expose identify as a kernel with `launch` and `launch_ref`, where the output is passed as input.

```rust
pub mod identify {
    pub fn launch<E: Numeric>(output: TensorHandle<R, E>, dim: u32)  {
       // simply call the ref function.
    }
    pub fn launch_ref<E: Numeric>(output: TensorHandleRef<R>, dim: u32) {
       ...
    }
}
```

vaijira:
I rename it to identity because it makes more sense to me but i change it to identity. I don't know either if perhaps it should be out tensor module or it can belong there.

crates/cubecl-linalg/src/tensor/identity.rs
```diff
@@ -0,0 +1,79 @@
...
+                offset += gap - remainder;
+            }
+        }
+        output[ABSOLUTE_POS] = line;
+    }
+}
+
+/// Launch identity matrix kernel.
+/// Ensure output is a tensorhandle containing a square matrix with shape 2 x 2.
+/// output will contain the identity matrix.
```
nathanielsimard: The shape comment is wrong I think, it's not always 2x2?

crates/cubecl-linalg/src/tensor/identity.rs
```diff
@@ -0,0 +1,79 @@
...
+pub fn launch<R: Runtime, C: Numeric>(
+    client: &ComputeClient<R::Server, R::Channel>,
+    output: &TensorHandle<R, C>,
+) {
+    launch_ref::<R, C>(client, &output.as_ref());
+}
+
+/// Launch identity matrix kernel by ref.
+/// Ensure output is a tensorhandle containing a square matrix with shape 2 x 2.
+/// output will contain the identity matrix.
```
nathanielsimard: Same

nathanielsimard (commented):
Only a few remaining comments, but it looks good.

crates/cubecl-linalg/src/tensor/identity.rs
vaijira: It was already solved. I don't know if it is worthy too to return a Result<(), IdentityError> in case matrix is not square.

crates/cubecl-linalg/src/tensor/identity.rs
vaijira: Solved

nathanielsimard (approved):

PR #406, opened 2025-01-10
[Feat] Make atomics generic and add comprehensive support for different types

wingertge:
Makes `Atomic` generic over a `Numeric`, and adds support and feature checks for all supported types on each backend.
The exact features supported differ greatly, so splits float atomic support into `load`/`store`/`swap` (supported pretty much everywhere), `add` (supported on CUDA and SPIR-V with compute capability 6.x or higher), and `min`/`max` which I can't currently test because it's not supported on compute capability 8.x. The feature exists in Vulkan, so hardware support is presumably either planned or already exists somewhere.

Because of the new hardware support checks involved in SPIR-V support, refactors the way features are registered there to make it more maintainable.

## Testing

Adds new tests for atomic add, min and max. Actually testing the atomicity is difficult, so they just test basic functionality at the moment.

### DIFF
--- a/crates/cubecl-core/src/frontend/element/atomic.rs
+++ b/crates/cubecl-core/src/frontend/element/atomic.rs
@@ -15,6 +15,14 @@ use crate::{
     unexpanded,
 };
 
+/// An atomic numerical type wrapping a normal numeric primitive. Enables the use of atomic
+/// operations, while disabling normal operations. In WGSL, this is a separate type - on CUDA/SPIR-V
+/// it can theoretically be bitcast to a normal number, but this isn't recommended.
+#[derive(Clone, Copy, Hash, PartialEq, Eq)]
+pub struct Atomic<Inner: CubePrimitive> {
+    pub val: Inner,
+}
+
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 pub enum AtomicOp {
     Load(UnaryOperator),
@@ -214,7 +222,6 @@ impl<Inner: Numeric> Atomic<Inner> {
     }
 }
 
-/// An atomic type. Represents an shared value that can be operated on atomically.
 impl<Inner: Int> Atomic<Inner> {
     /// Compare the value at `pointer` to `cmp` and set it to `value` only if they are the same.
     /// Returns the old value of the pointer before the store.
@@ -318,11 +325,6 @@ impl<Inner: Int> Atomic<Inner> {
         ));
         new_var.into()
     }
-}
-
-#[derive(Clone, Copy, Hash, PartialEq, Eq)]
-pub struct Atomic<Inner: CubePrimitive> {
-    pub val: Inner,
 }
 
 impl<Inner: CubePrimitive> CubeType for Atomic<Inner> {
--- a/crates/cubecl-core/src/frontend/element/atomic.rs
+++ b/crates/cubecl-core/src/frontend/element/atomic.rs
@@ -3,8 +3,8 @@ use std::fmt::Display;
 use serde::{Deserialize, Serialize};
 
 use super::{
-    init_expand_element, ExpandElementBaseInit, ExpandElementTyped, IntoRuntime, LaunchArgExpand,
-    Numeric,
+    init_expand_element, ExpandElementBaseInit, ExpandElementTyped, Int, IntoRuntime,
+    LaunchArgExpand, Numeric,
 };
 use crate::{
     frontend::{CubeContext, CubePrimitive, CubeType, ExpandElement},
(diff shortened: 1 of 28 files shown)


crates/cubecl-wgpu/src/lib.rs
```diff
@@ -33,13 +34,14 @@ mod tests {
...
 #[cfg(all(test, feature = "spirv"))]
+#[allow(unexpected_cfgs)]
 mod tests_spirv {
     pub type TestRuntime = crate::WgpuRuntime<crate::spirv::VkSpirvCompiler>;
     use cubecl_core::flex32;
     use half::f16;
 
     cubecl_core::testgen_all!(f32: [f16, flex32, f32, f64], i32: [i8, i16, i32, i64], u32: [u8, u16, u32, u64]);
-    cubecl_linalg::testgen_matmul_plane!([f16, flex32, f32]);
+    //cubecl_linalg::testgen_matmul_plane!([f16, flex32, f32]);
```
vaijira, line 44: Was this commented out on purpose?

wingertge: yes, they're very slow (on the order of half an hour) and should only be run when necessary

crates/cubecl-core/src/frontend/element/atomic.rs
```diff
@@ -53,108 +52,68 @@ impl Display for AtomicOp {
...
-/// An atomic type. Represents an shared value that can be operated on atomically.
-pub trait Atomic: Sized + CubeType
-where
-    ExpandElement: From<<Self::Primitive as CubeType>::ExpandType>,
-    ExpandElement: From<<Self as CubeType>::ExpandType>,
-{
-    /// The numeric primitive represented by the atomic wrapper.
-    type Primitive: Numeric;
-
+impl<Inner: Numeric> Atomic<Inner> {
```
nathanielsimard, line 55: Much better UX imo

crates/cubecl-core/src/frontend/element/atomic.rs
```diff
@@ -239,50 +195,84 @@ where
...
                 lhs: *ptr,
                 rhs: *value,
             }),
             *new_var,
         ));
         new_var.into()
     }
+}
+
+/// An atomic type. Represents an shared value that can be operated on atomically.
```
nathanielsimard, line 217: That doc should be on the top of the struct, not the impl block I think.

crates/cubecl-core/src/runtime_tests/atomic.rs
```diff
@@ -0,0 +1,193 @@
...
+    float_feat: AtomicFeature,
+) -> bool {
+    match F::as_elem_native_unchecked() {
+        Elem::Float(kind) => {
+            client
+                .properties()
+                .feature_enabled(Feature::FloatAtomic(float_feat))
+                && client
+                    .properties()
+                    .feature_enabled(Feature::Type(Elem::AtomicFloat(kind)))
```
nathanielsimard, line 23: Feature::FloatAtomic => Feature::AtomicFloat to have the same pattern as the Elem::AtomicFloat. (Prefix instead of postfix)

crates/cubecl-cpp/src/shared/element.rs
```diff
@@ -25,21 +25,35 @@ pub enum Elem<D: Dialect> {
...
+pub enum AtomicKind<D: Dialect> {
     I32,
+    I64,
     U32,
+    U64,
+    F16,
+    BF16,
+    F32,
+    F64,
+    _Dialect(std::marker::PhantomData<D>),
```
nathanielsimard, line 42: Is there a reason for the dialect to be included here? Maybe add a comment.

crates/cubecl-hip/src/runtime.rs
```diff
@@ -112,6 +115,14 @@ fn create_client<M: WmmaCompiler<HipDialect<M>>>(
     );
     let mut device_props = DeviceProperties::new(&[Feature::Plane], mem_properties, topology);
     register_supported_types(&mut device_props);
+    // Not sure if there's a good way to check for support on HIP
```
nathanielsimard, line 118: ROCm is only supported by newer devices, so I expect the features to be available most of the time (but it's a big guess).

crates/cubecl-wgpu/src/compiler/wgsl/compiler.rs
```diff
@@ -324,6 +324,10 @@ impl WgslCompiler {
                 kind => panic!("{kind:?} is not a valid WgpuElement"),
             },
             cube::Elem::Bool => wgsl::Elem::Bool,
+            cube::Elem::AtomicFloat(i) => match i {
+                cube::FloatKind::F32 => wgsl::Elem::AtomicF32,
+                kind => panic!("atomic<{kind:?}> is not a valid WgpuElement"),
```
nathanielsimard, line 329: Double atomic 🤯 `HydrogenFloat` it is called I think!

crates/cubecl-core/src/frontend/element/atomic.rs
wingertge: That was an extra doc left over cause I just replaced the trait with the impl block. There's already documentation on the type, so I just removed it.

crates/cubecl-core/src/runtime_tests/atomic.rs
wingertge: guess that also makes sense. I just figured the feature is floating point atomics, hence `FloatAtomics`. But consistency is also good.

crates/cubecl-cpp/src/shared/element.rs
wingertge: I added it because I need it to construct the inner element. alternatively, I could directly print the type names, but then you have doubled up string literals and that seemed more error prone. I think the dialect is better overall, so I'll just add a comment.

nathanielsimard (approved):

PR #408, opened 2025-01-10
Reduce stride

maxtremblay:
Now the reduce kernel works when the output and input have different stride patterns. I added relevant tests.


louisfd (approved):
LGTM

Issue #270, opened 2024-11-18
`Line` Support for Bitwise Ops


quinton11:
Implementing Bitwise Ops for burn tensors [here](https://github.com/tracel-ai/burn/issues/2234).

For the Jit Backend to have an implementation, the Line struct will need to implement rust's bitwise operation traits,
`std::ops::BitAnd`
`std::ops::BitOr`
`std::ops::BitXor`
`std::ops::Not`
`std::ops::Shl - left shift`
`std::ops::Shr - right shift`

PR #411, opened 2025-01-12
patch: extending the int trait with not ops

quinton11:
Patch for #270. 

Necessary for implementing the Not operation in burn [here](https://github.com/tracel-ai/burn/issues/2234)

Somehow missed it in the initial PR


quinton11:
@laggui Minor but necessary patch

nathanielsimard (approved):

PR #412, opened 2025-01-12
feat: not-trait-impl-for-int-types

quinton11:
Patch for #270 

Unary bitwise `Not` trait implementation for int types [here](https://github.com/tracel-ai/burn/issues/2234)


quinton11:
@laggui

laggui (approved):

PR #414, opened 2025-01-13
Fix/feature flags

laggui:
Fixed some feature flags that were causing issues in Burn when building without default features.

- ~On wasm, `subgroups` are not always available for webgpu (dates back to [this discord discussion](https://discord.com/channels/1038839012602941528/1091796857996451942/1312046753826209895)) so I added a feature flag (currently not enabled by default).~ /edit: wrong approach, should check operations instead #417 
- `async-channel` is no longer optional (also used outside of mpsc)


louisfd (approved):
LGTM

PR #417, opened 2025-01-13
Fix webgpu subgroup import

nathanielsimard:


laggui (approved):
LGTM!

Tested for web examples with Burn ✅

PR #419, opened 2025-01-13
Streamline some memory handling

ArthurBrussee:
This PR streamlines some memory allocation problems, particularly when using ExclusivePages.

## WGPU create() mapping
Firstly, applicable to any memory allocator, is a different approach to efficiently use create(). Using `queue.write_buffer_with` is nice because we can keep one single ComputePass for a batch of operations, which has some benefits. However, the implicit ordering of `queue.write_buffer_with` has been a bit of a nightmare (#83, #156, #405). Additionally, it's still not optimal - `write_buffer_with` still needs to allocate staging_buffers on the fly, and doesn't re-use them. Lastly, wgpu ideally wants  global uniform data marked as `uniform` instead of `storage`. That helps uniformity analysis, and can be faster on some integrated GPU's with special fast memory for uniforms.

This PR instead splits out small, uniform, allocations and large create calls. For small uniforms we use a StagingBelt which efficiently re-uses staging buffers, and a seperate encoder to manually insert these copies before the compute work. Because we now only have a tiny subset of buffers to worry about, we can simplify the memory locking to just keeping a reference to it.

For this to work some more of the memory management had to be pushed to the wgpu stream. The good news is is that removes the brittle connection between `stream.flush()` and `server.on_flushed()`.

## ExclusivePages allocator

Secondly, this PR reword the ExclusivePages allocator:
- No more ring buffer. This caused the allocator to cycle between free pages which in turn meant they weren't deallocated properly. 
- Similair to the above, don't keep pages in a hashmap, as the random order causes dealloc behaviour.
- Use a more standard exponential distribution with _much fewer_ pools. The fragmentation between pools meant memory would often go un-used, whereas a more greedy strategy seems to better re-use memory.
- Allow allocating from N neighbour pools, if they happen to have a free buffer.
- Allow allocating pages smalller than the page size in the pools.
- Tweak the deallocation curve. Deallocating larger buffers more quickly actually doesn't really make sense. Since they are used less frequently, it takes _more_ allocations before we know they really are free!

## Results

Some results on a test of training my model (gaussian splats) to 5k steps. This has lots of dynamic memory usage patters, dynamic shapes, and spike-y workloads. Note the blue "memory used" line is much lower than reserved but that's expected - I'm measuring memory _after_ a training step when lots of buffers became free again.

Before:
image

time:  4m29s

After:
image

time: 4m17s

Interestingly, this workload (dynamic shapes, and spike-y memory usage) does really badly with sliced memory, that might have to be investigated in the future.

Slices (Same before and after this PR)
image

time: 4m34s


crates/cubecl-runtime/src/memory_management/memory_manage.rs
```diff
@@ -259,39 +244,51 @@ impl<Storage: ComputeStorage> MemoryManagement<Storage> {
...
-            panic!("Unable to find valid pool partition point: No memory pool big enough to reserve {size} bytes.");
+            panic!("No memory pool big enough to reserve {size} bytes.");
         }
 
-        let pool = &mut self.pools[pool_ind];
-        if pool.max_alloc_size() < size {
-            panic!("No memory pool big enough to reserve {size} bytes.");
+        // Try to find a free slot on the pool, or on its neighbours. The amount of neighbours considered
+        // is hardcoded atm.
+        for i in 0..self.n_pool_candidates {
```
nathanielsimard, line 278: Can this be detrimental to the slice strategy?

crates/cubecl-runtime/src/memory_management/memory_pool/exclusive_pool.rs
```diff
@@ -1,42 +1,47 @@
...
-    slice_id: SliceId,
-    dealloc_mark: bool,
+    slice: Slice,
+    alloc_size: u64,
 }
 
+// How much more to allocate (at most) than the requested allocation. This helps memory re-use, as a larger
+// future allocation might b able to re-use the previous allocation.
+//
+// The default is the golden-ratio. The science behind it is that it feels cooler than a random number,
```
nathanielsimard, line 33: Yes I agree !!

crates/cubecl-runtime/src/memory_management/memory_pool/exclusive_pool.rs
```diff
@@ -45,141 +50,123 @@ impl ExclusiveMemoryPool {
...
+            // Return a smaller part of the slice. By construction, we only ever
+            // get a page with a big enough size, so this is ok to do.
+            page.slice.storage.utilization = StorageUtilization { offset: 0, size };
+            page.slice.padding = padding;
+
+            let id = page.slice.storage.id;
+            let handle = page.slice.handle.clone();
+
+            // If this is the base allocation, mark page as used.
+            // Otherwise, let it go if noone else needs it.
```
nathanielsimard, line 118: typo: noone

crates/cubecl-runtime/src/memory_management/memory_pool/exclusive_pool.rs
```diff
@@ -45,141 +50,123 @@ impl ExclusiveMemoryPool {
...
-        self.pages.insert(
-            storage.id,
-            MemoryPage {
-                slice_id,
-                dealloc_mark: false,
-            },
-        );
-        self.slices.insert(slice_id, slice);
-        handle_slice
+        assert!(size <= self.page_size);
```
nathanielsimard, line 128: Should we keep that? I'm actually unsure, maybe with a message?

crates/cubecl-wgpu/src/compute/stream.rs
```diff
@@ -47,11 +66,49 @@ impl WgpuStream {
...
+            WgpuStorage::new(
+                device.clone(),
+                BufferUsages::STORAGE
+                    | BufferUsages::COPY_SRC
+                    | BufferUsages::COPY_DST
+                    | BufferUsages::INDIRECT,
+            ),
+            &memory_properties,
+            memory_config,
+        );
```
nathanielsimard, line 100: So this pool is only used for staging buffers?

crates/cubecl-wgpu/src/compute/stream.rs
```diff
@@ -329,19 +380,148 @@ impl WgpuStream {
...
+                .get_resource(binding.clone(), offset_start, offset_end)
+        {
+            res
+        } else if let Some(res) =
+            self.memory_management_uniforms
+                .get_resource(binding.clone(), offset_start, offset_end)
+        {
+            // This resource now might be used for computations, so we _cannot_ use it anymore for create() calls.
+            // This keeps the binding alive which means create() won't try to use it.
+            self.locked_copy_handles.push(binding);
```
nathanielsimard, line 400: Yeah that makes a lot of sense to handle that here.

crates/cubecl-wgpu/src/compute/stream.rs
```diff
@@ -329,19 +380,148 @@ impl WgpuStream {
...
+
+        // We'd like to keep operations as one long ComputePass. To do so, this tries to submit
+        // all copy operations & all execute operations in one batch. To do so, we do all copy operations
+        // at the start of the encoder, and all execute operations afterwards. For this re-ordering to be valid,
+        // a buffer we copy to MUST not have any outstanding compute work associated with it.
+        // - For small uniform buffers, any small handles with compute work are kept
+        //   in self.locked_copy_handles which means the allocation won't try to use them.
+        // - For bigger buffers, we do break the compute stream.
+
+        let allocator = if aligned_len < MAX_UNIFORM_BUFFER_SIZE {
```
nathanielsimard, line 427: I feel stupid a bit right now, since I kinda thought that all buffers were actually uniform by default when called with a single allocation! Obviously, we want that for performance! I'm wondering if it's not possible to 'defragment' the 'real' GPU memory allocated for all memory pages. When it gets quite stable in terms of utilization, we could perform one huge defragmentation of all our pages to speed up following work.

crates/cubecl-wgpu/src/compute/stream.rs
```diff
@@ -1,14 +1,32 @@
...
+use cubecl_runtime::{
+    memory_management::{
+        self, MemoryDeviceProperties, MemoryHandle, MemoryManagement, MemoryPoolOptions,
+        SliceBinding, SliceHandle,
+    },
+    TimestampsError, TimestampsResult,
+};
+use wgpu::{util::StagingBelt, BufferDescriptor, BufferUsages, ComputePipeline};
+
+const MAX_UNIFORM_BUFFER_SIZE: u64 = 8192;
```
nathanielsimard, line 15: Is it really the max?

crates/cubecl-wgpu/src/compute/stream.rs
```diff
@@ -329,19 +380,148 @@ impl WgpuStream {
...
+    pub fn memory_usage(&self) -> cubecl_runtime::memory_management::MemoryUsage {
+        self.memory_management
+            .memory_usage()
+            .combine(self.memory_management_uniforms.memory_usage())
+    }
+
+    fn flush_if_needed(&mut self) {
+        // For now we only consider the number of handles locked, but we may consider the amount in
+        // bytes at some point.
+        if self.tasks_count >= self.tasks_max || self.locked_copy_handles.len() >= 32 {
```
nathanielsimard, line 494: I would use a multiple of `tasks_max` for the maximum handles locked before we flush. Since it doesn't consider small handles, we could approximate with a factor of 2. `tasks_max` is the setting that should be used to reduce memory usage in general. Maybe we could introduce another setting and deriving a default from tasks_max.

crates/cubecl-wgpu/src/compute/stream.rs
```diff
@@ -329,19 +380,148 @@ impl WgpuStream {
...
+    }
+
     pub fn flush(&mut self) {
         // End the current compute pass.
-        self.pass = None;
+        self.compute_pass = None;
+
+        self.staging_belt.finish();
+        let copy_encoder = std::mem::replace(&mut self.copy_encoder, create_encoder(&self.device));
+        let encoder = std::mem::replace(&mut self.encoder, create_encoder(&self.device));
```
nathanielsimard, line 505: I was really wondering why we needed two encoders in the state. `copy_encoder` seems like a copy of the main `encoder` instead of the encoder used for copy. I would rename it to `encoder_memcopy`.

crates/cubecl-wgpu/src/compute/stream.rs
```diff
@@ -329,19 +380,148 @@ impl WgpuStream {
...
+        self.locked_copy_handles.clear();
+
+        // Cleanup allocations and deallocations.
+        self.memory_management.cleanup();
+        self.memory_management.storage().perform_deallocations();
+
+        self.memory_management_uniforms.cleanup();
+        self.memory_management_uniforms
+            .storage()
+            .perform_deallocations();
```
nathanielsimard, line 524: I think a bit more comments are necessary, it's hard to know why we need two memory management pools and a staging belt.

From what I see:

* StagingBelt => Serves as a pool of preallocated buffers to be used as staging buffers for small copies, such as kernel metadata.
* Memory Management Uniforms => Contains the buffers used in the kernels, copied from `StagingBelt`
*Memory Management => Everything else.

I'm still wondering is we can use uniform memory for big buffers.

crates/cubecl-runtime/src/memory_management/memory_manage.rs
ArthurBrussee: It's disabled by default (or well - by default it only considers 1 pool for subslices). I've also tried it with - and it was _maybe_ a bit better with it on, but since my test case doesn't work great with subslices anyway I didn't want to change behaviour here.

crates/cubecl-runtime/src/memory_management/memory_pool/exclusive_pool.rs
ArthurBrussee: No crucial but seemed nice as I got it wrong at some point. Added a message.

crates/cubecl-wgpu/src/compute/stream.rs
ArthurBrussee: No this is like before the managment that allocates nearly everything! Have clarified it a bit.

crates/cubecl-wgpu/src/compute/stream.rs
ArthurBrussee: I'm not 100% sure what you mean! Maybe some different terminology. In wgpu a "uniform" is something that is gauranteed to be:
- Read only
- The same value for every thread in a threadgroup

The compiler can largely infer this anyway, but, in wgsl you can explicatly mark a binding as

@group(0) @binding(0) var<uniforms> example_meta_data: array<u32>;

instead of 

@group(0) @binding(0) var<storage, read_write> example_meta_data: array<u32>;

I haven't actually changed the wgsl compiler yet to take advantage of this, but, this should be allowed now, as we know uniforms use an exclusive page and are allocated with BufferUsages::UNIFORMS. I'm not sure it'll make a ton of difference but yeah I think some drivers can take this as a hint to place that memory in some frequently accessed cache.

So, not sure how that relates to a single allocation or fragmentation!

crates/cubecl-wgpu/src/compute/stream.rs
ArthurBrussee: Maybe max is a bad name - renamed and added a comment.

crates/cubecl-wgpu/src/compute/stream.rs
ArthurBrussee: Ah yes basing it on tasks_max makes sense! Went with a factor 8 as really this should be an emergency brake, not a cause more flushing.

crates/cubecl-wgpu/src/compute/stream.rs
ArthurBrussee: Renamed this & added some comments

crates/cubecl-wgpu/src/compute/stream.rs
ArthurBrussee: Have added some more comments! Hopefully other comment clarifies that for big buffers BufferUsages::UNIFORMS doesn't make sense per se.

crates/cubecl-wgpu/src/compute/stream.rs
nathanielsimard: Ho I had in mind contiguous, so if you allocate 100 MiB, that's all contiguous in memory.

crates/cubecl-wgpu/src/compute/stream.rs
```diff
@@ -330,18 +402,175 @@ impl WgpuStream {
...
+        let copy_uniforms_encoder = std::mem::replace(&mut self.copy_uniforms_encoder, {
+            self.device
+                .create_command_encoder(&wgpu::CommandEncoderDescriptor {
+                    label: Some("CubeCL Tasks Encoder"),
+                })
+        });
+        let tasks_encoder = std::mem::replace(&mut self.tasks_encoder, {
+            self.device
+                .create_command_encoder(&wgpu::CommandEncoderDescriptor {
+                    label: Some("CubeCL Uniforms Copy Encoder"),
```
nathanielsimard, line 553: The labels are swapped.

crates/cubecl-runtime/src/memory_management/memory_pool/exclusive_pool.rs
```diff
@@ -1,185 +1,184 @@
...
 struct MemoryPage {
-    slice_id: SliceId,
-    dealloc_mark: bool,
+    slice: Slice,
+    alloc_size: u64,
+    free_count: u32,
 }
 
+// How much more to allocate (at most) than the requested allocation. This helps memory re-use, as a larger
+// future allocation might b able to re-use the previous allocation.
```
nathanielsimard, line 32: b => be

crates/cubecl-wgpu/src/compute/stream.rs
ArthurBrussee: Ahhh yes ok makes sense, no I think like you said all allocations are contigious by default

nathanielsimard (approved):

PR #421, opened 2025-01-14
fix: bitwise-not-unary-op-for-cubeprimitive

quinton11:
Patch for #270 
- Implement not unary op for CubePrimitive [here](https://github.com/tracel-ai/burn/issues/2234)


quinton11:
@laggui

nathanielsimard (approved):

PR #422, opened 2025-01-14
Pipeline API for Cuda

louisfd:
Addition of the Cuda Pipeline API ([see Cuda example](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html?highlight=pipeline#multi-stage-asynchronous-data-copies-using-cuda-pipeline)) for async memory copy. It should be used in matmul pipelined.


crates/cubecl-core/src/frontend/context.rs
```diff
@@ -100,6 +100,13 @@ impl CubeContext {
         matrix
     }
 
+    /// Create a new matrix element.
```
nathanielsimard, line 103: Wrong doc

nathanielsimard (approved):
LGTM. The only thing to do before merging is to fix the docs and maybe add a bit more.

nathanielsimard (approved):

PR #423, opened 2025-01-14
Fix reduce cube count too big

nathanielsimard:


laggui (approved):
Validated that it fixed the previous error I was getting ✅ 

```
=== PANIC ===
A fatal error happened, you can check the experiment logs here => '/tmp/text-classification-ag-news\experiment.log'
=============
thread 'main' panicked at .cargo\registry\src\index.crates.io-6f17d22bba15001f\wgpu-23.0.1\src\backend\wgpu_core.rs:2326:18:
wgpu error: Validation Error

Caused by:
  In ComputePass::end
    In a dispatch command, indirect:false
      Each current dispatch group size dimension ([65536, 1, 1]) must be less or equal to 65535


```

PR #425, opened 2025-01-15
Slice/try cast unchecked

nathanielsimard:
Will be needed for async memory copy at multiple places.


louisfd:
Merged in other PR

PR #428, opened 2025-01-15
Fix some path issues with arch linux

maxtremblay:
I had issue with the default cuda and cuda-tools installation on arch linux using the package manager (pacman). I made some path more flexible. Would be great if some users could try it on their own machine with a different setup.


nathanielsimard (approved):
Fix format and merge :)

PR #430, opened 2025-01-16
[Feat] Autotune rework, now with type magic

wingertge:
Removes the need for the tune macro with type magic and by converting `AutotuneOperationSet` to a struct instead of a trait. It never really needed to be a trait to begin with, so this simplifies things.

Draft PR for burn update: https://github.com/tracel-ai/burn/pull/2710


nathanielsimard (commented):
I've looked at the code, and I'm having trouble understanding why we need so many traits for the type magic to work. At some point, I think we should have a contributor book with all the design decisions and explanations behind a lot of stuff. It would have been a great place for the doc around how the type magic works. Maybe for now, just adding a README.md in the autotune folder is sufficient, but I think some docs are necessary to "get" what is going on!

Also not having the autotune operations implement in Burn doesn't help the review. It's hard to see the difference in usage without that, so maybe having the PR in Burn linked to this one would be great!

I think this is still a net improvement. Not more proc-macros and less state for each autotune operation.

wingertge:
> I've looked at the code, and I'm having trouble understanding why we need so many traits for the type magic to work. At some point, I think we should have a contributor book with all the design decisions and explanations behind a lot of stuff. It would have been a great place for the doc around how the type magic works. Maybe for now, just adding a README.md in the autotune folder is sufficient, but I think some docs are necessary to "get" what is going on!

Yeah I wanted to do that anyways, it was more about checking if it works for fusion for now. Basically the problem is this: `Tunable` can't have a marker, because different tunables of the same type may have different markers. So we need to `IntoTunable` trait with a marker, then the marker gets erased when it gets turned into an actual tunable.
The reason the marker needs to exist is because the trait resolver can't properly distinguish between trait implementations a lot of the time, even if in practice they're mutually exclusive. The marker makes it so that technically, these are completely different traits so there's no conflict.
The reason for the function traits are so we can have additional calls on  the functions for giving extra info, like maybe overriding the name, or for non-result tunables there's an `ok` function to wrap them in `Ok`.
The reason we need a trait for Result and normal output is because the type system can't distinguish the two, since `Result<Out, Err>` could also be `Out`, where `Out` just happens to be a `Result`. Specialization would remove the need for this extra trait, but that's probably not happening any time soon.
I'll write all this up in a nicer format and put it in a module level doc comment probably.

> Also having the autotune operations implemented in Burn doesn't help the review. It's hard to see the difference in usage without that, so maybe having the PR in Burn linked to this one would be great!
I forgot to link the branch where I updated it, I opened a draft PR and linked it now

nathanielsimard (approved):

PR #431, opened 2025-01-16
Add cubecl std

maxtremblay:
As we many things, the hardest step is the first. So, now that it exists, we can start using it.


louisfd (approved):
LGTM

PR #432, opened 2025-01-17
[Feat] Fast float math in SPIR-V

wingertge:
Adds a new option to the macro to set the floating point behaviour on a per-kernel basis. The default is the existing safe one, but there are options for various unsafe optimizations, currently only for SPIR-V beacuse as far as I can tell CUDA has no options like this.

Some of these optimizations allow reordering or fusing operations even if that changes precision, others require certain conditions to be met to be safe (NotNan, NotInf). Since the  safety of this can vary from kernel to kernel and not all options may be desirable, a bitfield can be set as a kernel option.

Also temporarily uses a more up to date branch of `rspirv` - if a new release isn't available in time for the next cubecl release, we could fall back to unsafe transmutes. I believe a new release is imminent though, since a PR is already approved.

# Quick explanation on these flags
Here's what the SPIR-V docs give as an example for how these flags impact float operations:

> For example, if the expression a + b + (-a) is decorated AllowReassoc then it may be implemented as b. This is valid whether or not it is also decorated NotInf even though the original expression may overflow to infinity when evaluated in floating-point.
>
> If the expression a + a + (-a) is not decorated AllowReassoc then it cannot, in general be rearranged. However, in this case, if it is decorated with NotInf then it may be implemented as a since the replacement is exact for all values that do not overflow to infinity and the value is undefined if one of the operands is infinity. If the expression is not decorated with either AllowReassoc or NotInf then the result must be infinity for sufficiently large but finite values of a.

# Testing
Optimizations by the driver can't really be tested, but the option is added to the binary tests to ensure it's emitted and there are no validation errors.


torsteingrindvik:
For CUDA: https://github.com/tracel-ai/cubecl/blob/main/crates/cubecl-cuda/src/compute/server.rs#L356

If that line has `--use_fast_math` it should opt-in to trading precision for performance: https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/#use-fast-math-use-fast-math

It would be nice to be able to do that on a per-kernel basis in order to be able to investigate how much it helps for math heavy kernels.


For more fine-grained control it would be nice to be able to use these: https://docs.nvidia.com/cuda/cuda-c-programming-guide/#intrinsic-functions


My understanding is that using `--use_fast_math` means opting in to all of less precise but faster versions.
In some situations that might be too heavy handed. 

It would be nice to be able to

```rust
#[cube(launch_unchecked)]
fn normal<F: Float>(input: &Array<F>, output: &mut Array<F>, foo: F) {
        // normal division, or fast if `--use_fast_math`
        output[ABSOLUTE_POS] = input[ABSOLUTE_POS] / foo;
    }
}

#[cube(launch_unchecked)]
fn fast<F: Float>(input: &Array<F>, output: &mut Array<F>) {
        // always `--use_fast_math` intrinsic version 
        output[ABSOLUTE_POS] = fast_math::div(input[ABSOLUTE_POS], foo);
    }
}
```

or something like that.

wingertge:
> For CUDA: https://github.com/tracel-ai/cubecl/blob/main/crates/cubecl-cuda/src/compute/server.rs#L356
> 
> If that line has `--use_fast_math` it should opt-in to trading precision for performance: https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/#use-fast-math-use-fast-math

That option is tangential and should maybe be included as well, but it's not the same thing as the SPIR-V flags. The SPIR-V options don't lower precision, they just make precision *non-deterministic*. Reordering operands or fusing operations can cause rounding to flip one way or the other, so it's banned by default. I believe CUDA enables it by default. For things like `NotNaN`, `NotInf` and `UnsignedZero`, I don't believe there's a CUDA equivalent. I think I'd add a `ReducedPrecision` flag that enables `--use_fast_math` on CUDA.

Another good question is actually the matter of defaults: should the reordering options be on by default? As far as I can tell they don't *reduce* precision, and we don't really care if the output is 100% deterministic across different driver versions.

crates/cubecl-core/src/codegen/integrator.rs
```diff
@@ -198,7 +207,19 @@ impl KernelSettings {
...
+    }
+
+    /// Activate debug symbols
+    pub fn debug_symbols(mut self) -> Self {
+        self.options.debug_symbols = true;
+        self
+    }
+
+    /// Set FP math mode
+    pub fn fp_math_mode(mut self, mode: FPMathMode) -> Self {
```
nathanielsimard, line 221: `fp_math_mode` is an unclear name, and `FPMathMode` doesn’t follow the Rust naming conventions. We could remove FP entirely and keep only `MathMode`.

wingertge: I can agree on the name being bad, but I don't think `MathMode` is very descriptive either. Maybe just call it `FastMath`, since it maps to that in CUDA and it's more descriptive?

nathanielsimard (approved):

PR #434, opened 2025-01-19
feat: bitwise not unary op

quinton11:
Patch for #270 

Rust uses `!` for both logical and bitwise not, where the operand differentiates the use; i.e bool for logical and int variations for bitwise

`wgsl` however uses `!` for logical not and `~` for bitwise not.

The `Not` op is still kept in tact as it was, handling boolean operands,

Tests now pass for the `bitwisenot` operation on `burn_jit` backends. When this is merged, I'll push the changes to `bitwise_ops` PR [here](https://github.com/tracel-ai/burn/pull/2498)

@laggui


nathanielsimard (approved):
LGTM

PR #436, opened 2025-01-19
Upgrade to wgpu 24

ArthurBrussee:
- Add atomic floats feature matrix to wgpu. In the future when f16 etc. are supported on wgpu this feature matrix might need to be more granular.
- Change checked execution - infinite loops are now again checked in checked mode as they can cause UB. It might make sense to make more granular options for this later.

### DIFF
--- a/crates/cubecl-wgpu/src/compiler/wgsl/compiler.rs
+++ b/crates/cubecl-wgpu/src/compiler/wgsl/compiler.rs
@@ -19,7 +19,7 @@ use cubecl_runtime::{DeviceProperties, ExecutionMode};
 use wgpu::{
     BindGroupLayoutDescriptor, BindGroupLayoutEntry, BindingType, BufferBindingType,
     ComputePipeline, DeviceDescriptor, PipelineLayoutDescriptor, ShaderModuleDescriptor,
-    ShaderStages,
+    ShaderRuntimeChecks, ShaderStages,
 };
 
 #[derive(Clone, Debug, Default)]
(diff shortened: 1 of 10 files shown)


crates/cubecl-wgpu/src/runtime.rs
```diff
@@ -212,6 +212,14 @@ pub(crate) fn create_client_on_setup<C: WgpuCompiler>(
     );
     let channel = MutexComputeChannel::new(server);
 
+    // The wgpu float32-atomic feature guarantees both add and min/max. It does only guarantee float32 support,
+    // but since f16 and others aren't supported anyway there is not much to it.
+    if features.contains(wgpu::Features::SHADER_FLOAT32_ATOMIC) {
+        device_props.register_feature(Feature::AtomicFloat(AtomicFeature::LoadStore));
+        device_props.register_feature(Feature::AtomicFloat(AtomicFeature::Add));
+        device_props.register_feature(Feature::AtomicFloat(AtomicFeature::MinMax));
```
wingertge, line 220: The way I've implemented the features for float atomics, `AtomicFeature` should be combined with registering the relevant atomic types. As far as I can tell, if an atomic type is supported at all then all operations are supported on it.
So this should also register `Feature::Type(Elem::AtomicFloat(FloatKind::F32))`

wingertge: Also, as far as I can tell `MinMax` isn't supported for float at least not universally, it isn't available in Vulkan on my card either.

wingertge: The PR feature matrix for float atomics says it's supported in Vulkan (but not on the 3080), but not implemented at all in wgpu. So it definitely should be disabled.

ArthurBrussee: Ah bad memory really thought the PR said it was supported, nvm.

And I see about the Elem::AtomicFloat feature, that makes sense! Thank you

crates/cubecl-wgpu/src/compiler/wgsl/compiler.rs
```diff
@@ -87,22 +87,29 @@ impl WgpuCompiler for WgslCompiler {
...
+        mode: ExecutionMode,
     ) -> Arc<ComputePipeline> {
         let source = &kernel.source;
         // Cube always in principle uses unchecked modules. Certain operations like
         // indexing are instead checked by cube. The WebGPU specification only makes
         // incredibly loose guarantees that Cube can't rely on. Additionally, kernels
         // can opt in/out per operation whether checks should be performed which can be faster.
-        //
+        let checks = ShaderRuntimeChecks {
+            bounds_checks: false,
```
nathanielsimard, line 98: I would add a comment that automatic bounds checks are done by the cubecl compiler.

ArthurBrussee: The comment above said as much but clarified this a bit

nathanielsimard: Yeah, the comment needed to be improved, especially with the new `bounds_check` option that needs to be set by us.

nathanielsimard (approved):

PR #438, opened 2025-01-20
Fix default hip feature flag

laggui:
Originates from [this discussion](https://github.com/tracel-ai/burn/discussions/2709), linked to [cubecl-hip-sys PR](https://github.com/tracel-ai/cubecl-hip/pull/13).

As it is, `cubecl-hip-sys` feature flags are not additive, so forcing the "rocm__6_3_1" version causes issues when trying to specify another version. I've removed it from the forced feature and moved it to the default for `cubecl-hip`.

I also removed the "channel-mpsc" check (previously required for the `async-channel` dep, which is no longer optional) and "std" check in the autotune module (which was causing issues for burn with no default features - seems to be unnecessary but correct me if I am wrong).

### DIFF
--- a/crates/cubecl-hip/Cargo.toml
+++ b/crates/cubecl-hip/Cargo.toml
@@ -15,6 +15,7 @@ default = [
   "cubecl-runtime/default",
   "cubecl-common/default",
   "cubecl-core/default",
+  "cubecl-hip-sys/rocm__6_3_1",
   "is_hip",
 ]
 std = ["cubecl-runtime/std", "cubecl-common/std", "cubecl-core/std"]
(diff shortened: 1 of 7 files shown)


syl20bnr (approved):
Excellent. That's how it should work.
PR incoming for the `cubecl-hip-sys` but this is not as trivial as I thought it would be.

nathanielsimard (approved):

syl20bnr:
@laggui version `6.3.1001` of `cubecl-hip-sys` published. 
If there is no `rocm__x_x_x` feature defined then the `build.rs` script tries to use the default ROCm version installed on the system.

PR #446, opened 2025-01-22
Add some missing ops: `leading_zeros`, `find_first_set`, `plane_ballot`

wingertge:
Adds support for a few missing ops required for certain operations. They seem to be almost always used together so I combined them into one.

Specifically:
`leading_zeros`, which counts the leading zeros of an integer (CUDA: `__clz`, wgsl: `countLeadingZeros`, SPIR-V: it's complicated).
`find_first_set` - this is different than `trailing_zeros`, because `ffs` returns 0 when no bits are set, and the 1-indexed bit number otherwise, starting from the least significant bit (CUDA: `__ffs`, wgsl: `firstLedingBit`, SPIR-V: `FindMsb (sort of)`).
`plane_ballot` - Takes a boolean from each invocation and collects them into a set of bitfields. Because plane size isn't known at compile time, a `Line` with 4 elements is always returned. All inactive/non-existent units will be set to 0 (I think, seems to be undefined on HIP). CUDA: `__ballot_sync`, wgsl: `subgroupBallot`, SPIR-V: `OpGroupNonUniformBallot`.

# Testing

Adds tests for the new operations. They're also used in the new vision kernel and the tests there pass so they seem to work.


nathanielsimard (approved):
Everything looks good !

PR #449, opened 2025-01-22
replace return with a terminate macro

### LINKED ISSUE #441: Returning from non-launch functions
In Cube, the `return` statement does not just exit a function—it terminates the entire kernel. This behavior is only appropriate for launch functions, i.e., those annotated with `#[cube(launch)]`. When used in kernel subroutines, `return` will not behave as expected in typical Rust, making it highly error-prone.

Furthermore, since returning non-void types is already disabled, the `return` statement rarely functions in the familiar Rust-like way.

Given these limitations, and until we fully support the `return` statement, we recommend removing it altogether. Instead, we propose introducing a more explicit `exit!()` macro, which clearly indicates that it will exit the entire kernel.

maxtremblay:
Closes #441

In the future, it would be nice to introduce back the return keyword with the expected behavior. However, I believe this is currently better than the weird behavior of the previous implementation.


louisfd (changes requested):
I have one change request, AND since it's a breaking change you will need to update Burn ASAP

crates/cubecl-wgpu/src/lib.rs
```diff
@@ -26,7 +26,7 @@ mod tests {
     pub type TestRuntime = crate::WgpuRuntime<crate::WgslCompiler>;
 
     cubecl_core::testgen_all!();
-    cubecl_linalg::testgen_matmul_plane!([f32]);
+    //cubecl_linalg::testgen_matmul_plane!([f32]);
```
louisfd, line 29: you should reactivate this

louisfd (approved):
LGTM, only one small comment

PR #450, opened 2025-01-22
[hip] Fix shuffle intrinsic names for __shfl_xor and __shfl_down

syl20bnr:
Fixed a bug identified in discussion: https://github.com/tracel-ai/burn/discussions/2709

### DIFF
--- a/crates/cubecl-cpp/src/hip/dialect.rs
+++ b/crates/cubecl-cpp/src/hip/dialect.rs
@@ -79,10 +79,10 @@ impl<M: WmmaCompiler<Self>> Dialect for HipDialect<M> {
         format!("__shfl({var}, {source})")
     }
     fn warp_shuffle_xor(var: &str, offset: &str) -> String {
-        format!("__shfl_xor_sync({var}, {offset})")
+        format!("__shfl_xor({var}, {offset})")
     }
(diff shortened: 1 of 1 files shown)


laggui (approved):
LGTM!

Went looking through the docs, maybe the `_sync` variant is not enabled by default? 

Taken from the [ROCm docs](https://rocm.docs.amd.com/projects/HIP/en/develop/reference/kernel_language.html#warp-cross-lane-functions):

> Note that the __sync variants are made available in ROCm 6.2, but disabled by default to help with the transition to 64-bit masks. They can be enabled by setting the preprocessor macro HIP_ENABLE_WARP_SYNC_BUILTINS. These builtins will be enabled unconditionally in ROCm 6.3. Wherever possible, the implementation includes a static assert to check that the program source uses the correct type for the mask.

Technically, it should be available for 6.3 no? Not familiar with the HIP functions though so I'll leave this up to you - just wanted to point this out.

syl20bnr:
Great find ! 
Not sure we need them yet because in Cuda dialect we force the mask to `-1` which is what the non sync version of the shuffle functions do in HIP.
 
I created an issue to add support the mask parameter: https://github.com/tracel-ai/cubecl/issues/451

PR #453, opened 2025-01-23
Merge all error types in a single one

maxtremblay:
It is easier to use in burn when there is a single error type.


louisfd (approved):
LGTM

PR #455, opened 2025-01-25
Move to our own typehash implementation

wingertge:
`type_hash` has been abandoned for years and uses ancient dependencies. Since the code is small enough, I just added it to `cuebcl-macros-internal` and migrated it to a cleaner `darling` implementation. This way we can control it more and potentially adjust it to the needs of cubecl IR.


crates/cubecl-ir/src/type_hash.rs
```diff
@@ -0,0 +1,202 @@
...
+
+/// A hash of a type's structure
+pub trait TypeHash {
+    /// Generate a hash of the type structure.
+    #[allow(unused)]
+    fn type_hash() -> u64 {
+        let mut hasher = fnv::FnvHasher::default();
+        Self::write_hash(&mut hasher);
+        hasher.finish()
+    }
```
nathanielsimard, line 12: What's the difference between a type hash and the `core::any::TypeId` ? Would hashing `core::any::type_name` would have the same effect?

wingertge: `TypeId` isn't stable across compilations, the point of a type hash is to version the structure so you can ensure a cache with binary serialization is valid for the current version of the IR. Otherwise you might get runtime errors or incorrect data.

wingertge: I added more documentation to make this clearer.

Cargo.toml
```diff
@@ -19,17 +19,18 @@ derive-new = { version = "0.6.0", default-features = false }
 derive_more = { version = "1", default-features = false }
 log = { default-features = false, version = "0.4.22" }
 
+fnv = "1"
```
nathanielsimard, line 22: Should we use the hash function from std? Also we might want to add a link to the original type hash repo in a Notice file similar to how Burn handles that: https://github.com/tracel-ai/burn/blob/main/NOTICES.md

wingertge: I'll add the notice, as for the hasher, there's two problems.
One, the default hasher isn't stable and isn't meant to be - it's meant to prevent key collisions and that involves random seeds.
The other problem is `no_std`. Rust's `DefaultHasher` is only available in `std`, not `core` or `alloc`. So we need `fnv` for no_std compatibility.

nathanielsimard (approved):

PR #458, opened 2025-01-30
Some cleaning in matmul tiles

maxtremblay:
- Rename `init_lhs` to `allocate_lhs` and similar for `rhs` and `accumulator`.
- Remove simple helper functions in accelerated that were only wrapper in top level functions.


louisfd (approved):
LGTM

PR #459, opened 2025-01-30
Replace all Float bound with Numeric for matmul

maxtremblay:
I replaced most bound from `<F: Float>` to `<N: Numeric>`. I believe this will be useful for quantization where we want to support matmul between `u8` elements. Since `Float: Numeric`, one can still perform matmul with float as before. So I don't see a reason not to do it anyway as it won't break any of the existing code.


louisfd (approved):
LGTM

PR #460, opened 2025-01-31
[Feat] Uniformity analysis for SPIR-V

wingertge:
Adds a new uniformity analysis for the SPIR-V backend, and applies the `Uniform` decoration to all outputs that are plane uniform. This can potentially allow more optimizations to be applied by the driver, especially in cases where we know a variable is uniform but the lower level driver wouldn't.

# Testing
All tests pass


nathanielsimard (approved):
LGTM

PR #461, opened 2025-02-01
Add inclusive/exclusive sum and prod

wingertge:
Adds plane inclusive/exclusive sum and prod.

# Testing

Adds new tests for the new ops.

### DIFF
--- a/crates/cubecl-core/src/frontend/element/base.rs
+++ b/crates/cubecl-core/src/frontend/element/base.rs
@@ -382,5 +382,7 @@ pub(crate) fn __expand_new<C: Numeric, Out: Numeric>(
     val: C,
 ) -> ExpandElementTyped<Out> {
     let input: ExpandElementTyped<C> = val.into();
-    <Out as super::Cast>::__expand_cast_from(scope, input)
+    let const_val = input.expand.as_const().unwrap();
+    let var = Variable::constant(const_val.cast_to(Out::as_elem(scope)));
+    ExpandElement::Plain(var).into()
 }
--- a/crates/cubecl-core/src/frontend/plane.rs
+++ b/crates/cubecl-core/src/frontend/plane.rs
@@ -88,6 +88,10 @@ pub mod plane_sum {
 }
 
 /// Perform an inclusive sum operation across all units in a plane.
+/// This sums all values to the "left" of the unit, including this unit's value.
+///
+/// # Example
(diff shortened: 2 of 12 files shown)


crates/cubecl-core/src/frontend/element/base.rs
```diff
@@ -382,5 +382,7 @@ pub(crate) fn __expand_new<C: Numeric, Out: Numeric>(
     val: C,
 ) -> ExpandElementTyped<Out> {
     let input: ExpandElementTyped<C> = val.into();
-    <Out as super::Cast>::__expand_cast_from(scope, input)
+    let const_val = input.expand.as_const().unwrap();
+    let var = Variable::constant(const_val.cast_to(Out::as_elem(scope)));
+    ExpandElement::Plain(var).into()
```
nathanielsimard, line 387: Nice fix! Forcing the constant is the right move here.

crates/cubecl-core/src/frontend/plane.rs
```diff
@@ -87,6 +87,66 @@ pub mod plane_sum {
     }
 }
 
+/// Perform an inclusive sum operation across all units in a plane.
+#[allow(unused_variables)]
+pub fn plane_inclusive_sum<E: CubePrimitive>(value: E) -> E {
+    unexpanded!()
+}
```
nathanielsimard, line 94: What's the difference with plane_sum? I would add that in the docs just so it's clearer which one is needed.

crates/cubecl-core/src/frontend/plane.rs
```diff
@@ -87,6 +87,66 @@ pub mod plane_sum {
...
+            out,
+        ));
+
+        output.into()
+    }
+}
+
+/// Perform an exclusive sum operation across all units in a plane.
+#[allow(unused_variables)]
+pub fn plane_exclusive_sum<E: CubePrimitive>(value: E) -> E {
```
nathanielsimard, line 122: same

nathanielsimard (commented):
Just a few comments.

crates/cubecl-core/src/frontend/element/base.rs
wingertge: It was causing issues because I implemented the `IntExpand` dynamic element, and that was causing this code to emit `i64` constants in WGPU (which isn't supported).

nathanielsimard (approved):

PR #466, opened 2025-02-06
Fix `spirv-dump` and feature registration for mixed kernels

wingertge:
Enabling `spirv-dump` would crash when it encounters a custom wgsl shader, and atomics or f16 would crash because they're removed from the WGPU device features.
This ensures the kernel only gets saved if `repr` is `Some`, and ensures the WGPU device is created with the full feature set, not the trimmed down one that is used for enabling Vulkan extensions.
Also reexports `WgpuCompiler`, which should've been done before.


laggui (approved):
LGTM!

Unrelated CI failure was fixed on main

PR #468, opened 2025-02-06
[Fix] Ensure volatile values propagate during forward pass in optimizer

wingertge:
Volatile values should kill everything that depends on them in GVN, but the volatility previously didn't propagate through the graph. Ensure they do so all dependents get killed.


nathanielsimard (approved):
LGTM

Issue #355, opened 2024-12-06
Add ability to get hold of ComputeClient when using existing wgpu setup


torsteingrindvik:
The wgpu device sharing example shows a possible way to integrate a wgpu device with other programs.

However if `R::client(&shared_device)` is called there is a panic:

>             unreachable!("Cannot select an adapter for an existing device.")

(from `crates/cubecl-wgpu/src/runtime.rs`  request_adapter)

As such there isn't a nice way (or any way?) to get a compute client.

Locally I modified `crates/cubecl-wgpu/src/runtime.rs` in `init_device` to return the compute client it creates, but a proper fix would be nice.

wingertge:
This should be relatively simple, but one thing to note is that SPIR-V relies on setting some features that aren't set by wgpu (like the Vulkan memory model), so the device still needs to be created on the cubecl side if you want to support SPIR-V. Or if both things need custom features I suppose you could copy paste what we do in SPIR-V and add whatever you need. Either way the device needs custom creation logic, a default wgpu device from another library wouldn't work.
I think we should document that somewhere for the existing device variant, maybe even check for the required features and error if they're not present (so you don't get an ugly segfault later).

PR #470, opened 2025-02-06
Make wgpu compiler dynamic and determined by adapter

wingertge:
Makes the `cubecl-wgpu` compiler dynamic and automatically checks the backend on initialization.

# Breaking changes

Manually compiling kernels requires a signature change from `Compiler::compile(...)` to `Compiler::default().compile(...)`.


crates/cubecl-hip/src/runtime.rs
```diff
@@ -107,6 +107,8 @@ fn create_client<M: WmmaCompiler<HipDialect<M>>>(
         // This is a guess - not clear if ROCM has a limit on the number of bindings,
         // but it's dubious it's more than this.
         max_bindings: 1024,
+        // Minimum supported
+        max_shared_memory_size: 49152,
```
nathanielsimard, line 111: Is the comment right?

crates/cubecl-wgpu/src/backend/base.rs
```diff
@@ -0,0 +1,157 @@
...
+}
+
+#[cfg(feature = "spirv")]
+pub async fn request_device(adapter: &Adapter) -> (Device, Queue) {
+    if is_vulkan(adapter) {
+        vulkan::request_vulkan_device(adapter).await
+    } else {
+        wgsl::request_device(adapter).await
+    }
+}
```
nathanielsimard, line 130: Yup, makes so much sense!

crates/cubecl-wgpu/src/backend/vulkan.rs
```diff
@@ -0,0 +1,275 @@
...
+    };
+
+    let descriptor = DeviceDescriptor {
+        label: None,
+        required_features: full_feat,
+        required_limits: limits,
+        // The default is MemoryHints::Performance, which tries to do some bigger
+        // block allocations. However, we already batch allocations, so we
+        // can use MemoryHints::MemoryUsage to lower memory usage.
+        memory_hints: wgpu::MemoryHints::MemoryUsage,
```
nathanielsimard, line 128: The `memory_hints` value should probably be extracted into a variable, since it is also used in the `adapter.create_from_raw` function call. I guess it is important for both to be the same

crates/cubecl-wgpu/src/backend/vulkan.rs
```diff
@@ -0,0 +1,275 @@
...
+}
+
+fn register_types(props: &mut DeviceProperties<Feature>, ext_feat: &ExtendedFeatures<'_>) {
+    use cubecl_core::ir::{Elem, FloatKind, IntKind};
+
+    let mut register = |elem| {
+        props.register_feature(Feature::Type(elem));
+    };
+
+    let supported_types = [
```
nathanielsimard, line 184: This is more like `default_types`

crates/cubecl-wgpu/src/backend/vulkan.rs
```diff
@@ -0,0 +1,275 @@
...
+            })
+        })
+        .collect::<Vec<_>>();
+    log::debug!("Supported CMMA sizes: {sizes:#?}");
+    for size in sizes {
+        props.register_feature(size);
+    }
+}
+
+fn conv_type(vk_ty: ComponentTypeKHR) -> Option<Elem> {
```
nathanielsimard, line 259: conv => comp ?

crates/cubecl-wgpu/src/compiler/base.rs
```diff
@@ -1,31 +1,113 @@
...
+};
+use derive_more::derive::From;
 
-use crate::WgpuServer;
+use crate::{WgpuServer, WgslCompiler};
+
+use super::wgsl::ComputeShader;
+
+#[derive(Debug, Clone)]
+pub enum DynCompiler {
```
nathanielsimard, line 16: It's weird to have the prefix `Dyn` when we use enum (static dispatch in Rust), but I get what it does. Maybe `CombinedCompiler`, `AutoCompiler` or `WgpuCompiler` would be better names.

crates/cubecl-wgpu/src/compiler/base.rs
```diff
@@ -1,31 +1,113 @@
...
+            DynRepresentation::SpirV(spirv_kernel) => spirv_kernel.fmt(f),
+        }
+    }
+}
+
+impl CompilerRepresentation for DynRepresentation {
+    fn shared_memory_size(&self) -> usize {
+        0
+    }
+}
```
nathanielsimard, line 54: Is this a problem? I guess we should do a match and call `shared_memory_size` on the inner compiler?

crates/cubecl-wgpu/src/compiler/base.rs
```diff
@@ -1,31 +1,113 @@
...
-pub trait WgpuCompiler: Compiler {
     fn compile(
-        server: &mut WgpuServer<Self>,
-        kernel: <WgpuServer<Self> as ComputeServer>::Kernel,
+        &mut self,
+        kernel: KernelDefinition,
+        compilation_options: &Self::CompilationOptions,
+        mode: ExecutionMode,
+    ) -> Self::Representation {
+        match self {
```
nathanielsimard, line 67: I guess you'll need an extra `allow` for clippy when `spirv` isn't enabled.

crates/cubecl-wgpu/src/compiler/base.rs
```diff
@@ -1,31 +1,113 @@
...
-
-    #[allow(async_fn_in_trait)]
-    async fn request_device(adapter: &Adapter) -> (Device, Queue);
-    fn register_features(
-        adapter: &Adapter,
-        device: &Device,
-        props: &mut DeviceProperties<Feature>,
-        comp_options: &mut Self::CompilationOptions,
-    );
+    ) -> CompiledKernel<DynCompiler>;
```
nathanielsimard, line 112: Is this needed anymore? Could we just implement the function for the `DynCompiler` ?

nathanielsimard (changes requested):

crates/cubecl-hip/src/runtime.rs
wingertge: it used to just be hardcoded to that for both CUDA and HIP, so I assume it was the minimum supported? I now check the actual minimum for CUDA (since it's almost always higher than 48k), but I don't fully understand the HIP code and can't really test if my code works so I used the hardcoded constant from before.

crates/cubecl-wgpu/src/backend/vulkan.rs
wingertge: conv as in convert, I should spell it out

crates/cubecl-wgpu/src/compiler/base.rs
wingertge: It was never used in anything other than CUDA, I think it should actually just be on the CUDA repr itself now that we give access to that

crates/cubecl-wgpu/src/compiler/base.rs
wingertge: It was to allow for robustness checks and spirv dumping that are specifically done for SPIR-V, but that really belongs in just a function in the backend so I moved it.

crates/cubecl-wgpu/src/compiler/base.rs
wingertge: not sure what you mean, I don't see any clippy warnings on either CI or locally with SPIR-V disabled

torsteingrindvik:
Since this touches relevant code parts I'll mention this issue: https://github.com/tracel-ai/cubecl/issues/355

Would be nice if device sharing could be possible to get hold of a client when doing device sharing.
Feel free to ignore if out of scope/not relevant enough :heart:

wingertge:
> Since this touches relevant code parts I'll mention this issue: #355
> 
> Would be nice if possible to get hold of a client when doing device sharing. Feel free to ignore if out of scope/not relevant enough ❤️

This PR doesn't touch the part that selects an adapter (which is where it errors), so I'd put that in a separate PR.

nathanielsimard (approved):

PR #471, opened 2025-02-07
Clean stage again

maxtremblay:
- Uniformize many names and clarify some short (or  wrong) names.
  - Use `tile_count` with either the `m`, `n` and `k` axes or the `row` and `col` axes
    for the number of tiles within a stage.
  - Use `tile_shape` for the shape of a tile. 
  - Use `tile_size` for the total number of elements in a tile. The product of all shape of a time.
- Introduce a new concept of `StageTiling` that replace many previously used traits and structs related to tiling. A `CompleteStageTiling` is able to work with both the mnk axes and the row-col axes.
- Remove a few now unused traits, structs and functions.

### DIFF
--- a/crates/cubecl-linalg/src/matmul/components/stage/base.rs
+++ b/crates/cubecl-linalg/src/matmul/components/stage/base.rs
@@ -5,7 +5,7 @@ use crate::matmul::components::tile::TileConfig;
 use crate::matmul::components::{config::MatmulConfig, global::AccumulatorLoader};
 use crate::matmul::components::{global, MatmulConfigFactory};
 use crate::matmul::components::{Ident, MatrixLayout};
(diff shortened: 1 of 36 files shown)


crates/cubecl-linalg/src/matmul/components/stage/base.rs
```diff
@@ -137,7 +139,7 @@ pub trait StageConfig: MatmulConfig {
     fn line_size(&self, ident: Ident) -> u32;
 
     /// Returns the [StageDim] for the given ident
```
laggui, line 141:
```suggestion
    /// Returns the [StageTiling] for the given ident
```

laggui (approved):
LGTM!

Missed a docstring, but approving in advance.

PR #476, opened 2025-02-11
More flexible matmul test

maxtremblay:
Prepare the matmul tests to be ready for quantization. However, this currently skip all tests related to quantization as it is not implemented yet. The strategy will be to partially enable quantization tests during development using this method:

https://github.com/tracel-ai/cubecl/blob/c1c333c56461aeea2bba6aac9f72c897ad725d01/crates/cubecl-linalg/src/matmul/tests/test_utils.rs#L148C1-L152C6

### DIFF
--- a/crates/cubecl-linalg/src/matmul/tests/test_utils.rs
+++ b/crates/cubecl-linalg/src/matmul/tests/test_utils.rs
@@ -10,7 +10,8 @@ use cubecl_core::{
 
 use crate::{
     matmul::{
-        components::{Ident, MatmulProblem},
+        components::{Ident, MatmulProblem, MatmulSelection, MatrixLayout},
+        kernels::matmul::Algorithm,
         tests::cmma_matmul::matmul_test_launcher::strides,
     },
     tensor::TensorHandle,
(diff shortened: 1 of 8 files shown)


crates/cubecl-linalg/src/matmul/tests/test_utils.rs
```diff
@@ -196,6 +343,91 @@ where
...
+                out[batch_out + row * n + col] -= sum;
+            }
+        }
+
+        // Add final constant term
+        for row in 0..m {
+            for col in 0..n {
+                out[batch_out + row * n + col] += (k as i32) * lhs_zero_offset * rhs_zero_offset;
+            }
+        }
```
maxtremblay, line 425: Here you can see the flow for the quantized matmul.

nathanielsimard: Another way to test it is to dequantize a tensor, perform matmul, requantize the tensor and see if the quantize matmul works.

nathanielsimard (approved):

Issue #85, opened 2024-08-28
Have you considered a minimal-effort CPU runtime?


PoignardAzur:
In your README, you mention wanting to build [a JIT cranelift backend for the CPU](https://github.com/tracel-ai/cubecl/issues/26).

I can see the appeal of such a backend, but at the same time, there are use-cases where users may really want a CPU runtime for their shaders and don't care that much about performance.

For instance, in Vello, we end up maintaining CPU pseudo-shaders in parallel of our actual WGSL shaders, mostly for testing and as a fallback. Personally, I'd like to push the fallback case even further so we can run Vello on machines without GPUs; in those cases, being able to run anything at all is a win, even with degraded performance. If we could achieve that *and* get rid of our duplicate CPU shaders, that would be a massive win for us.

Have you considered making a best-effort CPU runtime? One where annotated rust functions are simply lowered to regular rust functions, and you leave auto-vectorization to the rustc backend? How much effort do you think it would take to implement that runtime?

nathanielsimard:
Making a low-effort CPU runtime would probably be as hard as making a proper CPU runtime. To speed things up, we might generalize our CUDA compiler to a C++ compiler and compile it using gcc or llvm. The compiler wouldn't be embedded, but it would be faster to develop.

PR #486, opened 2025-02-16
WIP cranelift runtime

### LINKED ISSUE #26: CPU Runtime: Cranelift
Make a Cranelift runtime for CPU. 
This will need experimentations, as it will probably need a different parallelization approach than GPU kernels (less threads, but better SIMD), so we will need other algorithms like matmul. 
- [ ] Leverage AVX2, AVX512

skewballfox:
Will eventually close #26 #85 

Adds support for a CPU runtime via cranelift. There seems to be quite a few necessary components to have a minimal running runtime that can compile and pass a single test. I'm open to either keeping this PR open/unmerged until it's fully complete, or until it reaches some threshold (compiles, passes X test set) so that full-support can be completed incrementally.

### DIFF
--- a/crates/cubecl-cranelift/Cargo.toml
+++ b/crates/cubecl-cranelift/Cargo.toml
@@ -1,0 +1,50 @@
+[package]
+authors = ["nathanielsimard <nathaniel.simard.42@gmail.com>"]
+categories = ["science"]
+description = "CPU runtime for CubeCL"
+edition.workspace = true
+keywords = ["simd", "cranelift"]
+license.workspace = true
+name = "cubecl-cranelift"
+readme.workspace = true
+repository = "https://github.com/tracel-ai/cubecl/tree/main/crates/cubecl-cuda"
+version.workspace = true
+
+[features]
+default = [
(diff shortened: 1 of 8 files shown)


skewballfox:
One thing I'm having trouble with is with Cranelift ([JIT](https://docs.wasmtime.dev/api/cranelift_jit/struct.JITModule.html))[Module](https://docs.rs/cranelift-module/0.116.1/cranelift_module/trait.Module.html)s, it seems memory and functions are bundled, and memory is managed from within the module. This seems to complicate making a compute storage struct for the cranelift runtime, and I'm trying to work out how to keep data/memory separate from the compiled code. I'm having trouble finding it again, but one option I came across was passing the memory as an ffi
- It might be possible to just make the storage struct a thin wrapper around the system allocator. Though admittedly I'm having trouble figuring out how to return a pointer to untyped uninitialized memory in a way that's safe to send.
- If we want to have a dedicated allocator for the cranelift runtime, would this be a good usecase for an arena allocator?

louisfd:
Outdated

PR #488, opened 2025-02-19
Clean some matmul

maxtremblay:
Remove the Matmul Selector since this was offering no benefit over the Algorithm and it was a lot of code for no real purpose.


louisfd (approved):
LGTM

PR #491, opened 2025-02-21
Strided Tile + Layout and Loader refactoring

louisfd:
- Added the SimpleStrided strategy [warning: uses a dummy loader for now, so it's very slow. Fix in the upcoming PRs]
- Upgraded tiling order to tiling layout, which can be contiguous (like before, and uses an underlying tiling order), or strided. The strided one gives strided tiles, and reflects the global memory layout exactly. 
- Refactored loaders into sync and async, as they do not have the same interface. Now a global matmul can be generic over all sync loaders, while being unable to use an async one, for instance.
- Remove some config and advanced config, like the tiling order/layout and the load mode, which are now encoded within the type system. 

Correctness and performance:
- All tests still pass. 
- No performance regression on my end.

### DIFF
--- a/crates/cubecl-linalg/src/matmul/components/global/single_stage/cooperative_loading.rs
+++ b/crates/cubecl-linalg/src/matmul/components/global/single_stage/cooperative_loading.rs
@@ -1,0 +1,85 @@
+use crate::matmul::components::global::tensor_view::{TensorReader, Window};
+use crate::matmul::components::global::{GlobalConfig, LoadMode, LoadingValidation};
+use crate::matmul::components::stage::TilingLayout;
+use crate::matmul::components::{Ident, InvalidConfigError, MatrixLayout};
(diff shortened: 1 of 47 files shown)


crates/cubecl-linalg/src/matmul/components/global/single_stage/cooperative_loading.rs
```diff
@@ -0,0 +1,85 @@
...
+            // If padding needed: TODO comptime conditional
+            for i in window.size..expected_window_size {
+                destination[i] = Line::cast_from(0);
+            }
+        }
+    }
+}
+
+#[cube]
+fn memcpy_slow<ES: Numeric>(source: Slice<Line<ES>>, destination: &mut SliceMut<Line<ES>>) {
```
nathanielsimard, line 81: I like the name

nathanielsimard (approved):

PR #494, opened 2025-02-24
Cube option

maxtremblay:
This make `Option<T>` where `T: LaunchArg` launchable in a cubecl kernel. 
Combined with 
```rust
match comptime!(option) { \* ... *\ }
```
it allows to match at comptime data that are runtime known. For example, for the first test named `kernel_option_scalar`, we can generate two different kernels based on an argument
of type `Option<i32>`.

```cuda

// When option is None
extern "C" __global__ void kernel_option_scalar(int output_0[], uint info[]) {

  int threadIdxGlobal = threadIdx.x + threadIdx.y * blockDim.x +
                        threadIdx.z * (blockDim.x * blockDim.y);
  const bool l_0 = threadIdxGlobal == uint(0);
  if (l_0) {
  }
}

// When option is Some(value: i32)
extern "C" __global__ void kernel_option_scalar(int output_0[], uint info[],
                                                int scalars_i32[]) {

  int threadIdxGlobal = threadIdx.x + threadIdx.y * blockDim.x +
                        threadIdx.z * (blockDim.x * blockDim.y);
  const bool l_0 = threadIdxGlobal == uint(0);
  if (l_0) {
    const uint l_1 = info[uint(0)];
    const bool l_2 = uint(0) < l_1;
    if (l_2) {
      output_0[uint(0)] = scalars_i32[0];
    }
  }
}
```

## Limitations

First, we can't use 
```rust
let x = Some(y);
``` 
directly in a kernel. Thus, I introduce a new function named `some` that does exactly. However, it has some issue with type inference. A proper fix of the way we expand enum variant should solve the issue.

Second, we can't use stuff like `option.map(|x| 2 * x)` to generate code at comptime. Currently, we have to use a match statement to achieve this behavior. 

## Design choices

I hesitated between introducing a new enum 
```rust
pub enum CubeOption<T> {
    Some(T),
    None
}
```
for which we could implement more utilities similar to `std::Option`. However, I decided to go with `std::Option` directly for now as a first step. 

## Justification

I believe this will quite useful to implement quantization in the matmul without doubling all implementations.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,6 +25,7 @@ serde = { version = "1.0.204", default-features = false, features = [
     "alloc",
 ] } # alloc is for no_std, derive is needed
 serde_json = { version = "1.0.119", default-features = false }
+ron = { version = "0.8", default-features = false }
 variadics_please = "1"
 
 dashmap = "5.5.3"
--- a/crates/cubecl-core/Cargo.toml
+++ b/crates/cubecl-core/Cargo.toml
@@ -44,6 +44,7 @@ num-traits = { workspace = true }
 paste = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
+ron = { workspace = true }
 variadics_please = { workspace = true }
 
 [dev-dependencies]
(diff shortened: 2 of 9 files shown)


louisfd (approved):
LGTM

PR #496, opened 2025-02-26
Feat/matmul/simple barrier

louisfd:
- Add cooperative barrier option, in which all the cube must perform all memcpy_async calls. 
- Add 5 loaders working with the strided layout. Some of them are experimental, will probably be deleted at some point
- Make loaders generic over the copy mechanism (pipeline vs barrier)
- Since i'm experimenting, i fork simple_matmul a bit. This will be cleaned up once it's more clear what we want to keep
- Fix memcpy to use references


louisfd:
I'll make another PR containing this one

PR #497, opened 2025-02-28
Barrier variant + memcpy_async benchmarks + Matmul with barrier

louisfd:
Barrier variant
- CubeCoop: the memcpy_async call is cooperative, i.e. should not depend on UNIT_POS
- CubeManual: the memcpy_async call is per unit, can be dispatched to any unit

memcpy_async benchmarks 
- Many variants on how to dispatch memcpy_async. Can beat sync loading on double buffering setup on some architecture, but otherwise the async has a lot of overhead it seems. 

Matmul with barrier
- Refactoring in matmul for a copy mechanism, that works with pipeline as well as barriers for async loading. There's also a sync dummy one for testing correctness if you don't have cuda, but it's not well coalesced so very slow. 
- WindowCooperativeLoading: loads all the needed slices cooperatively
- Other loadings using CubeManual are coming soon. 
- Async loading do not support check bounds anymore, because the way it was done was naive and slow. A better solution coming soon.


crates/cubecl-linalg/src/matmul/components/global/single_stage/simple/matmul_barrier_dummy.rs
```diff
@@ -0,0 +1,237 @@
...
+> {
+    _ms: PhantomData<MP>,
+    _stage_matmul: PhantomData<SMM>,
+    _lhs_loading: PhantomData<LL>,
+    _rhs_loading: PhantomData<RL>,
+}
+
+#[cube]
+impl<MP: MatmulPrecision, SMM, LL, RL> GlobalMatmul<MP>
+    for SimpleBarrierDummyMatmul<MP, SMM, LL, RL>
```
nathanielsimard, line 132: If it's the same as the SimpleBarrierMatmul, I would simply make it generic on the copy mechanism instead of forking the first one.

louisfd: I actually tried making it generic in my subsequent upcoming PR, but it turned out to be a pain. In the end I deleted the dummy one, it's not that useful.

PR #499, opened 2025-03-03
Feat/reduce fuse

nathanielsimard:
Stuff needed for fusing reduce kernels in burn


crates/cubecl-reduce/src/args.rs
```diff
@@ -0,0 +1,332 @@
...
+
+pub trait ReducePrecision {
+    type In: Numeric;
+    type Out: Numeric;
+}
+
+impl<In: Numeric, Out: Numeric> ReducePrecision for (In, Out) {
+    type In = In;
+    type Out = Out;
+}
```
maxtremblay, line 17: I am not a big fan of the name `Precision`. Doesn't really make sense when `In` or `Out` are integers. I suggest `ReduceDType` or `ReduceIO` instead?

maxtremblay (approved):

PR #506, opened 2025-03-06
Derive enum

maxtremblay:
Finally, here is the PR for the improved enum bahavior. They are now usable in launch function and we can perform comptime match with runtime variant. 

The only unpolished part is that we need to use `MyEnum::new_VariantName(x, y, z)` instead of `MyEnum::VariantName { x, y, z }` with runtime values. This is a limitation because they are no syntactic difference between `MyEnum::Variant` and `some_path::MyStruct`. However, for the former, we need to call the expand type of `MyEnum` (the first segment of the path) and, for the later, we need to call the expand type of `MyStruct` (the last segment of the path). I decided to keep the original behavior to keep struct working as they are today by introducing the `new_Variant` pattern

### DIFF
--- a/crates/cubecl-core/src/runtime_tests/enums.rs
+++ b/crates/cubecl-core/src/runtime_tests/enums.rs
@@ -1,0 +1,153 @@
+use crate as cubecl;
+use cubecl::prelude::*;
+
+#[derive(CubeLaunch, Clone, Debug, Hash, PartialEq, Eq, Copy)]
+pub enum TestEnum<
+    T: std::fmt::Debug + CubeType + Clone + IntoRuntime + LaunchArg + std::cmp::Eq + std::hash::Hash,
+> {
+    A(i32, u32),
+    B(BStruct),
+    C(T),
+    D,
+    E { x: i32 },
+    F { x: i32, y: u32 },
+}
+
+#[derive(CubeLaunch, Clone, Debug, Hash, PartialEq, Eq, Copy)]
+pub struct BStruct {
+    x: i32,
+    y: u32,
+}
+
+// We just check that it compiles for the syntax.
+#[allow(unused_variables)]
+#[allow(clippy::needless_match)]
+#[cube(launch)]
+pub fn kernel_comptime_variants(#[comptime] test: TestEnum<i32>) {
+    let test2 = comptime! {
+        match test {
+            TestEnum::A(x, y) => TestEnum::A(x, y),
+            TestEnum::B(x) => TestEnum::B(x),
+            TestEnum::C(x) => TestEnum::C(x),
+            TestEnum::D => TestEnum::D,
+            TestEnum::E {x} => TestEnum::E { x },
+            TestEnum::F {x, ..} => TestEnum::F { x, y: 2 }
+        }
+    };
+}
+
+// We just check that it compiles for the syntax.
+#[allow(unused_variables)]
+#[allow(clippy::needless_match)]
+#[allow(clippy::clone_on_copy)]
+#[cube(launch)]
+pub fn kernel_runtime_variants(test: TestEnum<i32>) {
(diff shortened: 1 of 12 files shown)


crates/cubecl-core/src/runtime_tests/enums.rs
```diff
@@ -0,0 +1,153 @@
...
+            #[test]
+            fn scalar_enum() {
+                let client = TestRuntime::client(&Default::default());
+                cubecl_core::runtime_tests::enums::test_scalar_enum::<TestRuntime>(client);
+            }
+
+            #[test]
+            fn array_float_int() {
+                let client = TestRuntime::client(&Default::default());
+                cubecl_core::runtime_tests::enums::test_array_float_int::<TestRuntime, f32>(
```
nathanielsimard, line 144: What does float_int mean?

crates/cubecl-macros/src/generate/cube_type/generate_enum.rs
```diff
@@ -89,6 +117,388 @@ impl CubeTypeEnum {
...
+        let launch_arg_expand = prelude_type("LaunchArgExpand");
+        let cube_type = prelude_type("CubeType");
+        let kernel_builder = prelude_type("KernelBuilder");
+
+        let name = &self.ident;
+        let name_expand = Ident::new(&format!("{}Expand", self.ident), Span::call_site());
+        let compilation_arg =
+            Ident::new(&format!("{}CompilationArg", self.ident), Span::call_site());
+
+        // let generics = self.expanded_generics();
```
nathanielsimard, line 382: dead code

crates/cubecl-macros/src/parse/expression.rs
```diff
@@ -32,6 +32,7 @@ impl Expression {
                 let left = Self::from_expr(*binary.left, context)?;
                 let right = Self::from_expr(*binary.right, context)?;
                 if left.is_const() && right.is_const() {
+                    // panic!("left -> {left:?}");
```
nathanielsimard, line 35: dead code

crates/cubecl-macros/src/parse/expression.rs
```diff
@@ -66,12 +67,14 @@ impl Expression {
             Expr::Path(path) => {
                 let name = path.path.get_ident();
                 if let Some(var) = name.and_then(|ident| context.variable(ident)) {
+                    // panic!("HELLO NATH {var:?};");
```
nathanielsimard, line 70: Hello !!

crates/cubecl-macros/src/parse/expression.rs
```diff
@@ -66,12 +67,14 @@ impl Expression {
...
                     if var.is_keyword {
                         Expression::Keyword { name: var.name }
                     } else {
                         Expression::Variable(var)
                     }
                 } else {
+                    // panic!("NAME  {name:?}\n CONTEXT {context:?}");
```
nathanielsimard, line 77: dead code

crates/cubecl-macros/src/generate/expression.rs
```diff
@@ -519,15 +525,71 @@ impl Expression {
...
+            }
+            // Match Pat1 | Pat2 | ...
+            Pat::Or(ref mut pat) => {
+                pat.cases.iter_mut().for_each(Self::expand_pat);
+            }
+            // Match the underscore pattern _
+            Pat::Wild(_) => {}
+            _ => {
+                panic!("unsupported pattern in match");
+                // panic!("{pat:?}");
```
nathanielsimard, line 575: dead code

nathanielsimard (approved):

crates/cubecl-core/src/runtime_tests/enums.rs
maxtremblay: The enum as two variants `Float` and `Int`

PR #507, opened 2025-03-06
Matmul: sync strided loader

louisfd:
A new loader that is sync but loads into the newer tiling layout. The goal is to have fairer comparison between sync and async.


crates/cubecl-linalg/src/matmul/components/global/single_stage/loader/sync/strided.rs
```diff
@@ -0,0 +1,71 @@
...
+        // 4
+        let line_size = config.global_line_size(ident);
+        // 256 / 4 = 64
+        let num_stage_lines = tiling.total_size() / line_size;
+        // 1*32= 32
+        let unit_count = config.num_planes() * config.plane_dim();
+        // 64 / 32 = 2
+        let num_loads_per_unit = comptime!(num_stage_lines / unit_count);
+
+        // 0 * 32 + 0..32 = 0..32
```
nathanielsimard, line 59: Are the comments really useful?

nathanielsimard (approved):

PR #512, opened 2025-03-11
Add memory_cleanup function, rework some dealloc functions

ArthurBrussee:
Add a function to explicitly cleanup memory when needed.

My model has a seperate 'refine' stage that allocates a bunch of memory every N steps. This sometimes pushes things over GPU limits. Explicatly calling memory cleanup helps a lot with this.

Some other bits:
- Let wgpu manage destroying the buffer. Rather than explicatly calling destroy(), dropping the last reference will just destroy the buffer. This fixes and issue where buffers in async futures ARE still needed by wgpu (even if not by burn) and are deallocated. The current code can crash here. This might also be more efficient as wgpu can decide when/how to release things.

- It's important to have bindings for memory up until the copy is actually in the wgpu queue. The rules around this were a bit tricky before. I've made it harder to get a wgpu buffer while dropping the resource to prevent this. This includes making read_buffers take Vec<Binding> instead of a wgpu buffer. That means using "managed" buffers for the sync buffer and query buffers. Bit annoying perhaps!

- Set the dealloc_period higher. Think I had pushed it a bit low to compensate for some things, but now that I can memory_cleanup manually a higher dealloc period is fine and is noticably faster. That's only for exclusive pages, sliced pages never deallocate atm anyway.

### DIFF
--- a/crates/cubecl-wgpu/src/compute/stream.rs
+++ b/crates/cubecl-wgpu/src/compute/stream.rs
@@ -523,6 +523,11 @@ impl WgpuStream {
             .combine(self.memory_uniforms.memory_usage())
     }
 
+    pub fn memory_cleanup(&mut self) {
+        self.memory_main.cleanup(true);
+        // No need to cleanup uniform memory as that isn't deallocated.
+    }
+
     fn flush_if_needed(&mut self) {
         // Flush when there are too many tasks, or when too many handles are locked.
         // Locked handles should only accumulate in rare circumstances (where uniforms
@@ -570,7 +575,7 @@ impl WgpuStream {
             .regulate(&self.device, self.tasks_count, index);
 
         // Cleanup allocations and deallocations.
-        self.memory_main.cleanup();
+        self.memory_main.cleanup(false);
         self.memory_main.storage().perform_deallocations();
 
         self.tasks_count = 0;
--- a/crates/cubecl-wgpu/src/compute/stream.rs
+++ b/crates/cubecl-wgpu/src/compute/stream.rs
@@ -20,8 +20,14 @@ const SMALL_UNIFORMS_BUFFER_SIZE: u64 = 8192;
 
 #[derive(Debug)]
 pub struct WgpuStream {
+    // Main memory pool to allocate buffers from.
     memory_main: MemoryManagement<WgpuStorage>,
+    // Memory pool for small uniform buffers (under [`SMALL_UNIFORMS_BUFFER_SIZE`]).
+    // Allocations in this pool are specially marked as for `uniform` usage.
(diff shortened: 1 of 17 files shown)


crates/cubecl-wgpu/src/compute/stream.rs
```diff
@@ -19,11 +21,12 @@ const SMALL_UNIFORMS_BUFFER_SIZE: u64 = 8192;
 #[derive(Debug)]
 pub struct WgpuStream {
     memory_main: MemoryManagement<WgpuStorage>,
-
-    uniforms_staging_pool: StagingBelt,
     memory_uniforms: MemoryManagement<WgpuStorage>,
+    memory_queries: MemoryManagement<WgpuStorage>,
```
nathanielsimard, line 25: I think it would make sense to add documentations on what each memory pool is used for.

ArthurBrussee: Added some comments to the pools! The queries pool is a bit unfortunate but just so that read_buffers can operate over a Binding

crates/cubecl-wgpu/src/compute/stream.rs
```diff
@@ -20,8 +20,14 @@ const SMALL_UNIFORMS_BUFFER_SIZE: u64 = 8192;
 
 #[derive(Debug)]
 pub struct WgpuStream {
+    // Main memory pool to allocate buffers from.
     memory_main: MemoryManagement<WgpuStorage>,
+    // Memory pool for small uniform buffers (under [`SMALL_UNIFORMS_BUFFER_SIZE`]).
+    // Allocations in this pool are specially marked as for `uniform` usage.
```
nathanielsimard, line 26: I would add that this is used for synchronization at the moment.

nathanielsimard: Also do you think we should have a staging memory pool?

ArthurBrussee: Oh mh actually switched over the sync buffer now to the main memory - it's just incidental I used the uniforms buffer, but given that the uniforms buffer has more special requirements using the main buffer is simpler.


And interesting about the stagin pool - the StagingBelt acts like a pool already for small allocs. For big allocs we indeed just allocate a buffer and throw it away immediatly. Using a pool would fix this and maybe speedup dataset creation. At the same time, you'd end up with all this memory left in the pool that's not really needed, and there would have to be a way to deallocate them again.

I think with slightly better cleanup logic (that doesn't rely on continously allocating) this could work and maybe be a nice speedup, but, will leave that for a seperate PR for now!

nathanielsimard (approved):

PR #513, opened 2025-03-12
File cache utility that syncs data on the disk.

nathanielsimard:
1. Added a CUDA PTX cache.  
2. Refactored the autotune cache to use the newer one.  
    - Now multiple processes can safely use the autotune cache.  
    - Fixed an issue where we used to recompute autotune benchmarks when we called the same kernel many times while the autotune was running.  
    - Now saves the autotune results as well. Useful for debugging, since we now have access to the timings and can see if any kernel encountered an error, along with the reason.  
    - Removed catch unwind in the cache; proper error handling is now required to work well with autotune. This makes the behavior more consistent across all platforms.


crates/cubecl-common/src/cache.rs
```diff
@@ -0,0 +1,315 @@
...
+/// The biggest constraint for the cache is that values should never change for a given key.
+/// There is no update possible; if a value is reinserted a second time with a different value but
+/// the same key, an error will arise.
+///
+/// This is important to keep the file format simple: there is no metadata, no headers, just a plain
+/// separator between each cache entry. Therefore, it isn’t possible to edit previously saved content.
+///
+/// ## No Big Files
+///
+/// The cache isn’t optimized for space; use it for small caches.
```
laggui, line 33: Yeah the hybrid cache you linked offline was a bit overkill for now 😄 this is much more limited but fits the scope pretty well

laggui (approved):

PR #514, opened 2025-03-12
Add bound checks inner for reduce kernels

nathanielsimard:
The `BoundChecksInner::Branch` fix memory error in the cuda backend when fuse on read is activated for the reduce kernels using the plane and plane shared strategy.

### DIFF
--- a/crates/cubecl-reduce/src/config.rs
+++ b/crates/cubecl-reduce/src/config.rs
@@ -15,6 +15,19 @@ pub enum LineMode {
     Perpendicular,
 }
 
+#[derive(Debug, PartialEq, Eq, Clone, Copy, Hash)]
+/// How bound checks is handled for inner reductions.
+pub enum BoundChecksInner {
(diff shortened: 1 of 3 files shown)


crates/cubecl-reduce/src/config.rs
```diff
@@ -15,13 +15,27 @@ pub enum LineMode {
     Perpendicular,
 }
 
+#[derive(Debug, PartialEq, Eq, Clone, Copy, Hash)]
+/// How bound checks is handled for inner reductions.
+pub enum BoundChecksInner {
+    /// No bound check is necessary.
+    None,
+    /// Using a mask is enough for bound checks.
+    Mask,
```
maxtremblay, line 24:
```suggestion
    /// Using a mask is enough for bound checks.
    /// This will still read the memory in an out-of-bound location,
    /// but will replace the value by the null value.
    Mask,
```

maxtremblay (approved):

PR #516, opened 2025-03-13
Migrate matmul-based convolution to `cubecl-linalg`

wingertge:
This was discussed ages ago in the Discord, but I forgot about it until now. Migrating the GEMM convolution to `cubecl-linalg` will make it much easier to keep it in sync with changes to the underlying matmul components, since it's so tightly coupled.


louisfd (approved):
LGTM, thank you btw 🙏 
Will you be removing the code from Burn as well?

wingertge:
> Will you be removing the code from Burn as well?

I have a PR ready but there's a bug with the new autotune keys that causes tests to hang, so I'd like to get that fixed and run the suite just in case. Everything passed before I merged the tuning changes, but just in case.

PR #527, opened 2025-03-18
Conditional read

louisfd:
Instruction similar to select, but where the read is not performed beforehand; only if the condition is true, but still branchless.


crates/cubecl-spirv/src/instruction.rs
```diff
@@ -323,6 +323,9 @@ impl<T: SpirvTarget> SpirvCompiler<T> {
                 }
             }
             Operator::Select(op) => self.compile_select(op.cond, op.then, op.or_else, out, uniform),
+            Operator::ConditionalRead(op) => {
+                self.compile_select(op.cond, op.slice, op.fallback, out, uniform)
```
nathanielsimard, line 327: Is it correct? I think this is more than a select, since the select doesn't have a read position? I guess it should do the same thing as the index operation.

PR #529, opened 2025-03-18
Simpler memory management in wgpu stream

ArthurBrussee:
This might be the 5th flip-flip on this piece of code, but, hopefully 5th time is the charm!

The main goal of this PR is to swap some custom GPU uploading for `queue.write_buffer`. This function is part of the [WebGPU specification ](https://developer.mozilla.org/en-US/docs/Web/API/GPUQueue/writeBuffer), and can be implemented more efficiently on the web. In particular, it saves a copy from WASM memory -> JS memory. See [this thread](https://github.com/gpuweb/gpuweb/discussions/1428) for more info. The second advantage is that on unified memory architectures (eg. macbooks), this implementation CAN skip using a staging buffer all together. I _think_ Chrome actually does this, but `wgpu` doesn't currently.

To do this, this goes back to how we did things a few PR's back: Memory is "excluded" or "locked" from being used for the copy queue, if it has any preceding compute work. This means "re-ordering" the copy operation to the start of the submission is allowed, so we can just use `write_buffer` to write data at the start of the next queue.submit().

While going over this code, I also noticed a particularly gnarly potential bug - Bindings have NO notion of what memory pool they came from. Their ID is just an atomically incremented counter. As such, the current `get_resource` function is wrong - it might find the ID in a different memory pool than it was allocated in. I'm not sure if this really happens in practice, but it seems very sketchy.

This PR merges the memory pools:
- Query resolves don't need a seperate pool, marking buffers as compatible with a QUERY_RESOLVE doesn't seem to have downsides / performance loss.
- The small uniforms buffer is no longer needed, all memory is done just with write_buffer. This does lose the ability to specially mark this memory as usable for 'uniform' buffers, but Cube didn't use this anyway, nor does my code which did use it get any faster from it.

Testing on desktop, I see no performance difference before/after this PR. On WASM, loading a particularly big model that used to OOM now works.

### DIFF
--- a/crates/cubecl-cuda/src/compute/server.rs
+++ b/crates/cubecl-cuda/src/compute/server.rs
@@ -182,7 +182,7 @@ impl ComputeServer for CudaServer {
 
     fn empty(&mut self, size: usize) -> server::Handle {
         let ctx = self.get_context();
-        let handle = ctx.memory_management.reserve(size as u64);
+        let handle = ctx.memory_management.reserve(size as u64, None);
         server::Handle::new(handle, None, None, size as u64)
     }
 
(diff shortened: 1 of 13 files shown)


nathanielsimard (approved):
Looks very clean, great job!

PR #531, opened 2025-03-19
Add availability check for conv

laggui:
Follow-up on #516 and https://github.com/tracel-ai/burn/pull/2908.

Turns out hanging macos tests were caused by an actual issue (had to debug this through the CI in https://github.com/tracel-ai/burn/pull/2922)

```
thread 'tests::cube::autodiff::f32_ty::ad_conv1d::tests::test_conv1d_groups' panicked at /Users/admin/.cargo/git/checkouts/cubecl-058c47895211d464/1a59333/crates/cubecl-wgpu/src/compiler/wgsl/compiler.rs:325:17:
Cooperative matrix-multiply and accumulate not supported.
```

@wingertge looks like the check was in Burn before but removed in the migration to the cubecl repo

/edit: tests pass locally but we need to either fix the dependency issue, change MSRV or move to edition 2024 😅

### DIFF
--- a/crates/cubecl-linalg/src/convolution/base.rs
+++ b/crates/cubecl-linalg/src/convolution/base.rs
@@ -1,7 +1,10 @@
-use crate::matmul::components::{
-    global::{AccumulatorLoader, OutputLoader},
-    stage::{StageMatmul, StageMatmulFamily},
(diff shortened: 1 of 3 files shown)


wingertge:
1.85 stable was released yesterday so it's time to bump prev anyways

nathanielsimard (approved):

PR #534, opened 2025-03-20
Perf/reduce

nathanielsimard:


crates/cubecl-reduce/src/config.rs
```diff
@@ -149,6 +152,17 @@ impl ReduceConfig {
                 ) as u32
             }
         };
+
+        if self.line_size_input > 1 && self.line_mode == LineMode::Perpendicular {
+            let rank = output.strides.len();
```
maxtremblay, line 157: Add a TODO that this can be improved

maxtremblay (approved):

PR #538, opened 2025-03-20
Refactor matmul quantized should run to use the same guard system

louisfd:


crates/cubecl-linalg/src/matmul/kernels/error.rs
```diff
@@ -82,6 +89,13 @@ impl Debug for MatmulLaunchError {
                     err.to_string()
                 )
             }
+            MatmulLaunchError::Unimplemented(err) => {
+                writeln!(
+                    f,
+                    "Unable to launch matmul because the a feature is not ready: {:?}",
```
maxtremblay, line 95:
```suggestion
                    "Unable to launch matmul because the feature is not ready: {:?}",
```

crates/cubecl-linalg/src/matmul/kernels/error.rs
```diff
@@ -165,3 +179,17 @@ impl Debug for MatmulAvailabilityError {
...
+
+pub enum MatmulUnimplementedError {
+    Quantization,
+}
+
+impl Debug for MatmulUnimplementedError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            MatmulUnimplementedError::Quantization => {
+                writeln!(f, "Quantization is not yet available.")
```
maxtremblay, line 191: When combined with `MatmulLaunchError`, the message will read as 

```
Unable to launch matmul because the feature is not ready: Quantization is not yet available.
```

This is a bit redundant.

Maybe we don't need this one and simply pass a string to `MatmulLaunchError::Unimplemented`.

maxtremblay (approved):
This is a little bit nitpicking, but I have a few comments on this PR.

PR #542, opened 2025-03-25
Update tune reduce

maxtremblay:
I try a new autotune key for reduce. Let's try it with burn also


ArthurBrussee:
Currently, reduce seems to autotune everytime a shape changes, but for my usecase where my sizes continuously vary a bit, autotune keeps firing. Is it viable to put the shape into buckets like matmul/conv have?

nathanielsimard:
> Currently, reduce seems to autotune everytime a shape changes, but for my usecase where my sizes continuously vary a bit, autotune keeps firing. Is it viable to put the shape into buckets like matmul/conv have?

We already put things into buckets like matmul and conv, we just have to made those buckets to not be too sensitive on shapes.

louisfd (approved):

PR #543, opened 2025-03-26
Refactor the way bindings are passed to the runtime

wingertge:
The current way bindings are passed to the runtime is extremely fragile and inflexible. This refactors the `Vec<Binding>` to instead use a new struct called `Bindings` that provides a much more flexible binding representation. It splits inputs from outputs so the ordering is a bit less fragile, and also splits scalars, metadata and tensormaps into their own representations.

* Scalars and metadata are now passed as host-side buffers and the runtime can choose how to handle them (allocate in the memory pool, use uniform parameters, use `__grid_constant__` kernel params, etc).
* The scalar order has been removed in favor of sorting by the `Ord` implementation on `Elem`, to provide a more robust mapping
* TensorMaps are passed as before, but are now just a field on `Bindings` instead of a separate runtime argument.
* Removes some legacy code left over from the `cpa` days

# Testing

All tests pass on all backends. I can't test HIP, so some testing for that would be welcome.


crates/cubecl-wgpu/src/compute/stream.rs
```diff
@@ -363,8 +375,20 @@ impl WgpuStream {
...
+        if data.len() < aligned_len as usize {
+            let mut buffer_view = self
+                .queue
+                .write_buffer_with(
+                    resource.buffer(),
+                    resource.offset(),
+                    BufferSize::new(aligned_len).unwrap(),
+                )
+                .unwrap();
+            buffer_view[0..data.len()].copy_from_slice(data);
```
ArthurBrussee, line 387: Hmm annoying that this is needed! Might as well just do this unconditionally I think

crates/cubecl-wgpu/src/compute/stream.rs
```diff
@@ -70,31 +70,43 @@ impl WgpuStream {
...
+        bindings: Bindings,
         dispatch: &CubeCount,
     ) {
         let dispatch_resource = match dispatch.clone() {
             CubeCount::Static(_, _, _) => None,
             CubeCount::Dynamic(binding) => Some(self.mem_manage.get_resource(binding)),
         };
 
+        let info = self.create(bytemuck::cast_slice(&bindings.metadata));
+        let scalars = bindings
```
ArthurBrussee, line 82: This might be more work, but, would it be possible to pack all info & scalars into one buffer? That has some small locality benefits, but also, on WebGPU there is a limit of 10 buffers. The more buffers leftover the more can be fused.

crates/cubecl-wgpu/src/compute/stream.rs
wingertge: makes sense. It probably didn't appear in the tests since wgpu still doesn't have support for types that have less than 4 byte alignment rn. But it's needed for SPIR-V.

crates/cubecl-wgpu/src/compute/stream.rs
wingertge: At most we could pack all the ones of the same size I think. We can bitcast between types but there's no reinterpreting of pointers in wgsl. Think if we do this it should be a separate PR since it requires a bunch of changes to the compiler too.

ArthurBrussee: Fwiw the WGSL alignment rules are well defined. So, it's possible to generate a giant a struct in wgsl with all the fields and set this as an opaque u8[]. Maybe f16 and other future types complicate this though.

anyway yeah just like marking things as `uniform` now it's not really related to the PR, mainly just much easier to do now.

crates/cubecl-wgpu/src/compute/stream.rs
ArthurBrussee: Sighhh I actually just read the wgpu source code - only write_buffer gets the special case fast path on WebGPU. write_buffer_with does not.

So, for WebGPU's sake, you'd still want to use write_buffer if possible. Sorry about that!

crates/cubecl-wgpu/src/compute/stream.rs
wingertge: oh yeah a struct would definitely work. Would just need to copy the individual scalar buffers into a packed buffer and generate a struct for it in the compiler for both wgsl and SPIR-V.

wingertge: You can sort by type size and make it work with alignment even with f16, that's what I did with CUDA. The only challenge is that you'd need to interleave the info tensor in between the 64-bit types and the remaining types and handle that properly. But not that hard.

crates/cubecl-runtime/src/server.rs
```diff
@@ -149,6 +148,90 @@ impl Handle {
...
+    /// Add a buffer binding
+    pub fn with_buffer(mut self, binding: Binding) -> Self {
+        self.buffers.push(binding);
+        self
+    }
+
+    /// Set the buffers to `bindings`
+    pub fn with_buffers(mut self, bindings: Vec<Binding>) -> Self {
+        self.buffers = bindings;
+        self
```
nathanielsimard, line 180: That's probably dangerous, I would expect with_buffers == with_buffer multiple times, but it's not since it removes all previously added values.

crates/cubecl-runtime/src/server.rs
```diff
@@ -149,6 +148,90 @@ impl Handle {
...
+    /// Add a scalar parameter
+    pub fn with_scalar(mut self, elem: Elem, length: usize, data: Vec<u64>) -> Self {
+        self.scalars.push(ScalarBinding::new(elem, length, data));
+        self
+    }
+
+    /// Set the scalars to `bindings`
+    pub fn with_scalars(mut self, bindings: Vec<ScalarBinding>) -> Self {
+        self.scalars = bindings;
+        self
```
nathanielsimard, line 192: same

crates/cubecl-cpp/src/shared/instruction.rs
```diff
@@ -409,17 +412,35 @@ for ({i_ty} {i} = {start}; {i} {cmp} {end}; {increment}) {{
...
+            Instruction::Metadata {
+                info_offset,
+                split_meta,
+                out,
+            } => {
                 let out = out.fmt_left();
-                writeln!(f, "{out} = info[{info_offset}];")
+                match *split_meta {
+                    true => writeln!(f, "{out} = static_info.x[{info_offset}];"),
+                    false => writeln!(f, "{out} = __ldg(&info[{info_offset}]);"),
```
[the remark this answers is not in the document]
wingertge: Load global, specifically through the readonly cache. It basically tells the compiler that it can safely assume it's always immutable data and it doesn't need to account for other things mutating it anywhere (which would require using the normal, non-immutable cache). The compiler does insert these automatically *usually* if the pointer is `const __restrict__`, but it's better to just use it explicitly I think.

crates/cubecl-runtime/src/server.rs
wingertge: Yeah I wasn't sure, I guess the field is public so you could just set it directly if you intend to override it. I'll change it.

crates/cubecl-core/src/compute/builder.rs
```diff
@@ -1,148 +1,127 @@
...
     pub context: Scope,
-    constants: Vec<ConstantInfo>,
-    inputs: Vec<InputInfo>,
-    outputs: Vec<OutputInfo>,
-    indices: HashMap<Elem, usize>,
-    num_constant: Id,
-    num_input: Id,
-    num_output: Id,
+    buffers: Vec<BufferInfo>,
+    scalars: HashMap<Elem, usize>,
```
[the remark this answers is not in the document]
wingertge: Good idea, didn't think of that

crates/cubecl-cpp/src/shared/kernel.rs
```diff
@@ -170,63 +189,89 @@ extern \"C\" __global__ void {}(
...
-            binding_index += 1;
 
-            match binding.vis {
-                Visibility::Read => {
-                    write!(f, "{} {}[]", binding.item, name)?;
-                    // TODO: It breaks slices, because we can't easily create pointer to __restrict__,
-                    // we should have multiple pointer types to enable that optimization.
-                    //
-                    // write!(f, "const {}* __restrict__ {}", binding.item, name)?;
+        if self.grid_constant {
```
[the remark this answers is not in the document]
wingertge: I figured it's an implementation detail, but I'll add a quick comment. The main thing is just that it's constant memory directly next to each SM so it's fast.

crates/cubecl-cpp/src/shared/kernel.rs
```diff
@@ -170,63 +189,89 @@ extern \"C\" __global__ void {}(
...
+                if binding_index < num_bindings {
+                    f.write_str(",")?;
                 }
             }
 
-            if binding_index < num_bindings {
-                f.write_str(",")?;
+            for size in [4, 2, 1] {
+                scalars_of_size(f, size, &mut binding_index)?;
+            }
```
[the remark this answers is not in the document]
wingertge: I also don't like this, but basically sizes are always powers of two and we measure in bytes so it can't be less than 1 byte. This is because to guarantee proper alignment of everything, we need to put 64 byte aligned (tensormap) types first, then 8 byte aligned types (pointers, 64-bit scalars), then the metadata (which is 4 byte aligned), then the remaining scalars in reverse size order. Hence 4-byte aligned, then 2-byte aligned, then 1-byte aligned. The length of each buffer is always a multiple of alignment, so as long as they're sorted, the alignment will always be correct.
This is a problem that arises from the fact that CUDA tightly packs its kernel params with no padding, at least as far as I've been able to find online. So putting, for example, a single u8 in there, followed by a u32, would cause the u32 to be misaligned and throw a runtime error.
It's a bit dodgy to have these sizes hardcoded, but I couldn't come up with a better way.

nathanielsimard (approved):

crates/cubecl-cpp/src/shared/instruction.rs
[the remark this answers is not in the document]
wingertge: Since it's CUDA/HIP exlcusive I've removed it now

PR #544, opened 2025-03-27
fix select in cond read

maxtremblay:


louisfd (approved):
LGTM

nathanielsimard:
The fix is strange, not sure it makes sense to have a checked read with line size > 1

PR #546, opened 2025-03-27
Fixes for naive matmul kernel

wingertge:
Fixes two bugs in naive_matmul:
One was introduced by the recent TensorMap PR, it used `rhs.stride(rank - 2)`, but the correct stride to use is actually `rhs.stride(rank - 1)`.
This was hidden by the second bug:
The kernel ensures the rhs tensor is always transposed, so k is the contiguous dimension. However, it never swaps back the strides. So for contiguous rhs matrices, strides are passed as `NK`, but for matrices that are already transposed and don't need to be corrected the strides are `KN`. This breaks every assumption about strides we make, so I fixed it by swapping the strides back to the normal order after the correction.

Also includes a fix for anchoring that prevents it from rounding down the value. This was breaking some of the convolution tests by producing invalid convolution sizes.

# Testing

This fixes the burn tests.

### DIFF
--- a/crates/cubecl-linalg/src/matmul/kernels/naive.rs
+++ b/crates/cubecl-linalg/src/matmul/kernels/naive.rs
@@ -55,7 +55,7 @@ fn matmul_kernel<N: Numeric>(
 
     for i in 0..k {
         let lhs_index = row * lhs.stride(rank - 2) / line_size + i + offset_lhs;
-        let rhs_index = col * rhs.stride(rank - 2) / line_size + i + offset_rhs;
+        let rhs_index = col * rhs.stride(rank - 1) / line_size + i + offset_rhs;
 
         sum += lhs[lhs_index] * rhs[rhs_index];
     }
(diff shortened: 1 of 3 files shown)


maxtremblay:
> Also includes a fix for anchoring that prevents it from rounding down the value. This was breaking some of the convolution tests by producing invalid convolution sizes.

Did you try using a min in the anchor instead?

wingertge:
> Did you try using a min in the anchor instead?

The min is dependent on the parameters of the convolution. I don't think we should ever shrink the tune input, that can very easily lead to using a kernel that's actually broken for a larger size (or in this case, is broken for the smaller size).

maxtremblay:
Ok, this is maybe out-of-scope for this PR, but @nathanielsimard should we consider having the option to ceil or floor with the anchor to avoid issues because of a hardcoded choice?

wingertge:
> Ok, this is maybe out-of-scope for this PR, but @nathanielsimard should we consider having the option to ceil or floor with the anchor to avoid issues because of a hardcoded choice?

I think when adding that it might also make sense to have a different kind of `max`, where it doesn't cap the size but caps the power padding. So instead  of just `next_power_of_2` it would do `min(next_power_of_2, next_multiple_of(max))`.

nathanielsimard (approved):

PR #560, opened 2025-04-01
[Feat]Implement basic clustering support and metadata

wingertge:
Adds basic support for clusters, including a kernel param (via the macro right now, since it's the only way we have to pass optional params), as well as globals that can read the cluster size and cube index within a cluster.

# Testing
Adds tests for the cluster metadata that are skipped when clusters are unsupported


crates/cubecl-core/src/frontend/topology.rs
```diff
@@ -71,6 +71,38 @@ The position of the working unit inside the cube along the Z axis.
...
+);
+
+constant!(
+    CLUSTER_DIM_Z,
+    crate::ir::Builtin::ClusterDimZ,
+    r"
+The dimension of the cluster along the Z axis.
+"
+);
+
```
nathanielsimard, line 105: I would add the prefix `CUBE_CLUSTER` otherwise it can be very confusing.

crates/cubecl-core/src/frontend/topology.rs
```diff
@@ -134,6 +166,39 @@ constant!(
...
+constant!(
+    CUBE_POS_CLUSTER_Y,
+    crate::ir::Builtin::CubePosClusterY,
+    r"
+The cube position in the cluster along the Y axis.
+"
+);
+
+constant!(
+    CUBE_POS_CLUSTER_Z,
```
nathanielsimard, line 195: I would use the prefix, so `CUBE_CLUSTER_POS_X`

crates/cubecl-core/src/runtime.rs
```diff
@@ -72,6 +72,8 @@ pub enum Feature {
     Barrier,
     /// Tensor Memory Accelerator features. Minimum H100/RTX 5000 series for base set
     Tma(TmaFeature),
+    /// Clustered launches and intra-cluster operations like cluster shared memory
+    Cluster,
```
nathanielsimard, line 76: Prefix

nathanielsimard (changes requested):
I would really prefer to prefix "cluster" with "cube." It's very confusing to have "cluster" without context. Maybe we could come up with a name that works in the hierarchy of "cube," but for now, I would use the prefix to clarify what it's a cluster of.

crates/cubecl-core/src/frontend/topology.rs
wingertge: I'm actually not sure that's the best option, it kind of makes it sound like it's the position of the `cubecluster` within the grid, not the position of the cube in the cluster.

wingertge: This naming is also consistent with `UNIT_POS_PLANE`.

nathanielsimard (approved):

PR #561, opened 2025-04-01
Reinterpret read

maxtremblay:
I introduce a new data type in `cubecl-std`. A `ReinterpretList` allows to read a list of some input type as if it was a list of some other type. This is semantically equivalent to bitcasting the original list, but the bitcast is done while reading. This is particularly useful to read metadata at the end of a Tensor or Array buffer. For example, this is done for quantization.

I also added the support for common rust macros such as `panic`, `todo`, `assert` and it is now trivial to add more macros that can be evaluated at comptime. (simply add its name to a list somewhere in `cubecl-macros`)

However, note that since the macros are evaluated at comptime, we have to be careful about their behavior with control flow.

For example, the following will always panic as the control flow is evaluated at runtime, but the panic is generated at comptime.

```rust
#[cube]
fn something(array: Array<i32>) {
   if array[0] < 0 { // you need a comptime condition for panic to work as expected.
     panic!("first element must be positive");
   }
   // rest of the kernel
}
```

However, we can now do the following without being yelled at by the compiler and using flaky comptime tricks.

```rust
#[cube]
fn something(#[comptime] fancy_feature: bool) {
  if comptime!(fancy_feature) {
    todo!("the fancy feature is not yet supported") 
  } else {
    // normal execution
  } 
}
``` 

Finally, I fixed some issues with vectorization of lines created by `Line::empty`.

## burn

There is a branch in burn to add those changes, but since burn is currently broken with cuda, I can't test. But it compiles.

https://github.com/tracel-ai/burn/tree/update-cubecl.

### DIFF
--- a/crates/cubecl-std/src/reinterpret_list.rs
+++ b/crates/cubecl-std/src/reinterpret_list.rs
@@ -3,6 +3,9 @@ use core::{cmp::Ordering, marker::PhantomData};
 use cubecl::prelude::*;
 use cubecl_core as cubecl;
 
+/// This struct allows to take a list of `Line<From>` and reinterpret it
+/// as a list of `To`. Semantically, this is equivalent to bitcasting the list of `Line<From>`
+/// to a list of `To`. When indexing, the index is valid in the casted list.
 #[derive(CubeType)]
 pub struct ReinterpretList<From: CubePrimitive, To: CubePrimitive, L: List<Line<From>>> {
     list: L,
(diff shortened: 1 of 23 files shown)


crates/cubecl-std/src/reinterpret_list.rs
```diff
@@ -0,0 +1,118 @@
...
+    #[cube(comptime)]
+    _phantom: PhantomData<(From, To)>,
+}
+
+#[cube]
+impl<From: CubePrimitive, To: CubePrimitive, L: List<Line<From>>> ReinterpretList<From, To, L> {
+    pub fn new(list: L, #[comptime] line_size: u32) -> ReinterpretList<From, To, L> {
+        let num_bytes_line_from = comptime!(core::mem::size_of::<From>() as u32 * line_size);
+        let num_bytes_to = comptime!(core::mem::size_of::<To>() as u32);
+        // panic!("FROM {num_bytes_line_from} TO {num_bytes_to}");
```
louisfd: dead code

louisfd (approved):
LGTM

PR #562, opened 2025-04-01
Gate debug symbols behind feature/env override

wingertge:
Gates all debug symbols behind a feature flag/environment variable override/kernel option (the already existing `debug_symbols`).


crates/cubecl-macros/build.rs
```diff
@@ -3,22 +3,24 @@ use std::env;
...
 
-    if nightly_feature_enabled || nightly_override_enabled {
-        println!("cargo:rustc-cfg=nightly");
-    }
+    let nightly_feature_enabled = env::var("CARGO_FEATURE_NIGHTLY").is_ok();
+    let nightly_override_enabled = env::var("CUBECL_DEBUG_NIGHTLY").is_ok();
+    let nightly_enabled = nightly_feature_enabled || nightly_override_enabled;
 
-    if debug_feature_enabled || debug_override_enabled {
+    if debug_enabled || nightly_enabled {
```
ArthurBrussee, line 19: Idk about having it always on with nightly! I like to dev in nightly sometimes, and still would want to have fast compile times

wingertge: That's not nightly, that's the nightly *feature*. So it's only set when `CUBECL_DEBUG_NIGHTLY` is set during compilation, or the nightly feature is explicitly enabled on cubecl-macros.

nathanielsimard (approved):

PR #564, opened 2025-04-02
Fix buffer write when < aligned_len

laggui:
Noticed a bunch of SPIR-V tests were failing when running the checks in Burn.

```
In Queue::write_buffer
    Copy size 25 does not respect `COPY_BUFFER_ALIGNMENT`
```

Some recent changes in wgpu stream assume 4-byte alignment which is always valid on wgpu but not necessarily SPIR-V. For example, the failing tests in `burn-vision` tried to allocated a 5x5 u8 tensor (bool).


nathanielsimard:
@laggui I think the bug was already addressed.

laggui:
> @laggui I think the bug was already addressed.

It was not when I opened the PR 😄 but you're right, looks like a similar fix was included as part of #543

Closing this

PR #566, opened 2025-04-02
[metal] Some fixes

syl20bnr:


crates/cubecl-cpp/src/shared/binary.rs
```diff
@@ -306,7 +311,8 @@ impl<D: Dialect> Binary<D> for Index {
 
         let item_out = out.item();
         if let Elem::Atomic(inner) = item_out.elem {
-            write!(f, "{inner}* {out} = &{lhs}[{rhs}];")
+            let addr_space = D::address_space_for_variable(lhs);
+            writeln!(f, "{addr_space}{inner}* {out} = &{lhs}[{rhs}];")
```
nathanielsimard, line 315: Does it work on cuda?

syl20bnr: yes, we have that already on other places with pointers
in CUDA and HIP the string is empty

nathanielsimard (approved):

PR #571, opened 2025-04-03
fix col major line size

louisfd:


crates/cubecl-linalg/src/matmul/kernels/matmul/base.rs
```diff
@@ -52,12 +52,12 @@ pub fn launch_ref<R: Runtime, MP: MatmulPrecision, A: Algorithm>(
     // }
 
     let check_layout = |tensor: &TensorHandleRef<'_, R>| match matrix_layout(tensor.strides) {
-        MatrixLayout::Contiguous => (false, false),
-        MatrixLayout::MildlyPermuted {
+        MatrixTensorLayout::Contiguous => (false, false),
```
maxtremblay, line 55: Maybe `TensorLayout` instead of `MatrixTensorLayout`

maxtremblay (approved):

PR #572, opened 2025-04-03
Add TMA tests and fix TMA loaders

wingertge:
Adds tests for TMA matmul and fixes some bugs. I'd like to try more things to make it faster, but want to get it to a working state first.


crates/cubecl-linalg/src/matmul/tests/cmma_matmul/tma_test_launcher.rs
```diff
@@ -0,0 +1,231 @@
...
+};
+use crate::matmul::kernels::matmul::Algorithm;
+use crate::matmul::tests::test_utils::Sample;
+use crate::matmul::tests::test_utils::TestPrecision;
+
+use super::matmul_test_launcher::{TensorRawParts, shape, tensor_size, transpose};
+
+/// Test the correctness of the specified Matmul on the given device,
+/// against a naive CPU implementation over the given problem
+pub fn test_tma_matmul_algorithm<A, P, R>(
```
louisfd, line 24: Seems like a lot of duplication. Is it only lhs and rhs line sizes that change? If so I think it should be refactored

crates/cubecl-linalg/src/matmul/tests/test_macros/cmma/suite.rs
```diff
@@ -432,3 +482,247 @@ macro_rules! matmul_standard_tests {
         }
     };
 }
+
+#[allow(missing_docs)]
+#[macro_export]
+macro_rules! matmul_tma_tests {
```
louisfd, line 488: Is this duplication necessary?

louisfd (commented):
Thanks. Seems good in general. I think there is a bit of duplication that could be avoided, and I have a few little comments.

crates/cubecl-core/src/frontend/container/slice.rs
```diff
@@ -251,6 +252,15 @@ mod metadata {
     }
 }
 
+pub(crate) fn is_tf32<C: CubePrimitive, T: CubePrimitive>(scope: &mut Scope) -> bool {
+    let ty_c = C::as_elem(scope);
+    let ty_t = T::as_elem(scope);
+    let ty_f32 = f32::as_elem(scope);
+    let ty_tf32 = tf32::as_elem(scope);
+
+    (ty_c == ty_f32 && ty_t == ty_tf32) || (ty_c == ty_tf32 && ty_t == ty_f32)
```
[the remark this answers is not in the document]
wingertge: that would trigger the initial check for equality. This is just to ensure tf32 and f32 are treated as equivalent when casting.

crates/cubecl-linalg/src/matmul/components/global/copy_mechanism.rs
```diff
@@ -6,18 +6,52 @@ use cubecl_core::prelude::*;
...
+
+    fn memcpy_async_tensor_to_shared_3d(
+        _this: &Self,
+        _source: &TensorMap<ES>,
+        _destination: &mut SliceMut<Line<ES>>,
+        _b: u32,
+        _x: u32,
+        _y: u32,
+    ) {
+        comptime!(unimplemented!(
```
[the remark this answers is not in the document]
wingertge: Can we leave that for now? I removed it in the im2col branch since the traits no longer exist and I can just use barrier directly now.

crates/cubecl-linalg/src/matmul/tests/cmma_matmul/tma_test_launcher.rs
wingertge: The main change is with the way input is generated, TMA needs pitched tensors but there's no way to make them work with how quantization works rn. So we need to split the input generation. I tried merging them at one point but quantization is blocking that.

crates/cubecl-linalg/src/matmul/tests/test_macros/cmma/suite.rs
wingertge: I was able to merge the test macros at least

crates/cubecl-linalg/src/matmul/components/global/copy_mechanism.rs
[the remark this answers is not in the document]
louisfd: Yaeh it was just a heads-up

PR #573, opened 2025-04-04
Matmul: stage buffering const

louisfd:
just a trick to avoid testing on single buffering then benchmarking on double, like it happened to me

### DIFF
--- a/crates/cubecl-linalg/src/convolution/selection.rs
+++ b/crates/cubecl-linalg/src/convolution/selection.rs
@@ -78,8 +78,7 @@ impl ConvSelector<ImplicitCmmaConv> for Balanced {
         };
 
         let selection = ConvSelection { matmul: selection };
-
-        // TODO support selection of double buffering
+        
         (selection, (config_input, STAGE_BUFFERING))
     }
 }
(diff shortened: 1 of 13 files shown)


crates/cubecl-linalg/src/convolution/selection.rs
```diff
@@ -81,7 +80,7 @@ impl ConvSelector<ImplicitCmmaConv> for Balanced {
         let selection = ConvSelection { matmul: selection };
 
         // TODO support selection of double buffering
```
maxtremblay, line 82: Remove or update the comment

crates/cubecl-linalg/src/matmul/components/stage/base.rs
```diff
@@ -180,11 +180,14 @@ pub trait StageConfig: MatmulConfig {
...
 
 #[derive(Clone, Copy, PartialEq, Eq, Hash, Debug)]
-pub enum Buffering {
+pub enum StageBuffering {
     Single,
     Double,
 }
+
+// TODO support selecting double buffering
+pub const STAGE_BUFFERING: StageBuffering = StageBuffering::Double;
```
maxtremblay, line 193: I confirmed by benchmarking that double is the best default choice.

maxtremblay (approved):

crates/cubecl-linalg/src/matmul/components/stage/base.rs
maxtremblay: also update the comment

PR #579, opened 2025-04-07
Get number of SMs and tensor cores from device

louisfd:
Feature available for CUDA only, otherwise falls back to the approx consts we had. Not benchmarked yet.

### DIFF
--- a/crates/cubecl-cuda/src/runtime.rs
+++ b/crates/cubecl-cuda/src/runtime.rs
@@ -99,6 +99,10 @@ fn create_client(device: &CudaDevice, options: RuntimeOptions) -> ComputeClient<
         let max_cube_count =
             CubeDim::new_3d(grid_dim_x as u32, grid_dim_y as u32, grid_dim_z as u32);
 
+        let num_streaming_multiprocessors =
+            Some(get_attribute(device_ptr, CU_DEVICE_ATTRIBUTE_MULTIPROCESSOR_COUNT).unwrap() as u32);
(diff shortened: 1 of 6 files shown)


crates/cubecl-cuda/src/runtime.rs
```diff
@@ -151,6 +158,21 @@ fn create_client(device: &CudaDevice, options: RuntimeOptions) -> ComputeClient<
...
+        70 => Some(8),   // Volta (SM 7.0)
+        75 => Some(8),   // Turing (SM 7.5)
+        80 => Some(4),   // Ampere A100 (SM 8.0)
+        86 => Some(16),  // Ampere GA10x (SM 8.6)
+        89 => Some(16),  // Ampere GA10B (SM 8.9, Jetson Orin)
+        90 => Some(64),  // Hopper H100 (SM 9.0)
+        91 => Some(64),  // Hopper L40S (SM 9.1)
+        92 => Some(64),  // Hopper H200 (SM 9.2)
+        100 => Some(64), // Blackwell (SM 10.0, est.)
+        _ => None,       // Unknown or unsupported architecture
```
nathanielsimard, line 172: Is there a reference for that? For H100 I think there is 4 tensor cores per SM, not 64.

nathanielsimard: From what I'm seeing, only versions 70 & 75 have 8 tensor cores per sm, all others have 4.

nathanielsimard (approved):

PR #590, opened 2025-04-10
Fix kernel selection

nathanielsimard:
https://github.com/tracel-ai/burn/pull/3006

### DIFF
--- a/crates/cubecl-linalg/src/matmul/components/global/args.rs
+++ b/crates/cubecl-linalg/src/matmul/components/global/args.rs
@@ -11,7 +11,9 @@ use crate::matmul::components::{self, MatmulPrecision, MatmulProblem, MatmulSele
 
 use super::Quantization;
 
-pub trait InputsLaunch: LaunchArg {
+/// Create the input runtime arguments for a matmul kernel that works on concrete inputs and
+/// output (not fused).
+pub trait ConcreteInputsFactory: LaunchArg {
     fn create<'a, R: Runtime>(
         lhs: &'a TensorHandleRef<'a, R>,
         rhs: &'a TensorHandleRef<'a, R>,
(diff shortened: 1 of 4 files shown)


crates/cubecl-linalg/src/matmul/components/global/args.rs
```diff
@@ -11,7 +11,9 @@ use crate::matmul::components::{self, MatmulPrecision, MatmulProblem, MatmulSele
 
 use super::Quantization;
 
-pub trait InputsLaunch: LaunchArg {
+/// Create the input runtime arguments for a matmul kernel that works on concrete inputs and
+/// output (not fused).
+pub trait ConcreteInputsFactory: LaunchArg {
```
maxtremblay, line 16: It is weird that a trait is named `ConcreteSomething` as trait are by nature abstract.

Maybe simply remove the prefix and use `InputsFactory` instead?

crates/cubecl-linalg/src/matmul/kernels/matmul/algorithm/selector.rs
```diff
@@ -22,14 +23,44 @@ const NUM_TENSOR_CORES_APPROX: u32 = 4;
 const NUM_PLANES_PER_TENSOR_CORES: u32 = 2;
 
 /// Select which kernel to launch for the given Algorithm.
+///
+/// Only works for concrete tensor inputs and output.
 #[allow(clippy::result_large_err)]
-pub fn select_kernel<MS: MatmulSpec, R: Runtime, A: Algorithm>(
+pub fn select_kernel_concrete<MS: MatmulSpec, R: Runtime, A: Algorithm>(
```
maxtremblay, line 29: Same here. Not a fan of suffix `concrete`

maxtremblay (approved):

PR #594, opened 2025-04-11
Added min max reduction

nathanielsimard:


crates/cubecl-reduce/src/instructions/max.rs
```diff
@@ -0,0 +1,94 @@
...
+    type Config = ();
+
+    fn requirements(_this: &Self) -> ReduceRequirements {
+        ReduceRequirements { coordinates: false }
+    }
+
+    fn from_config(_config: Self::Config) -> Self {
+        Max {}
+    }
+    fn null_input(_this: &Self, #[comptime] line_size: u32) -> Line<In> {
```
maxtremblay, line 31: Really picky, but can you add a newline here?

crates/cubecl-reduce/src/instructions/max.rs
```diff
@@ -0,0 +1,94 @@
...
+        _this: &Self,
+        accumulator: Self::AccumulatorItem,
+        _shape_axis_reduce: u32,
+    ) -> Out {
+        let mut sum = In::from_int(0);
+        #[unroll]
+        for k in 0..accumulator.size() {
+            sum += accumulator[k];
+        }
+        Out::cast_from(sum)
```
maxtremblay, line 84: This is wrong. If you copied from MaxAbs it could be wrong in both. You want something like 

```rust
let mut max = In::min_value();
#[unroll] 
for k in 0..accumulator.size() {
    let candidate = accumulator[k];
    max = select(candidate > max, candidate, max);
}
Out::cast_from(max)
```

crates/cubecl-reduce/src/instructions/min.rs
```diff
@@ -0,0 +1,94 @@
...
+        accumulator: Self::AccumulatorItem,
+        _shape_axis_reduce: u32,
+    ) -> Out {
+        let mut sum = In::from_int(0);
+        #[unroll]
+        for k in 0..accumulator.size() {
+            sum += accumulator[k];
+        }
+        Out::cast_from(sum)
+    }
```
maxtremblay, line 85: See comment in Max implementation. This is wrong

maxtremblay (changes requested):

crates/cubecl-reduce/src/instructions/max.rs
maxtremblay: It is indeed broken in MaxAbs 

https://github.com/tracel-ai/cubecl/blob/09a65feace88167a8ec02348d7e538c4b821e103/crates/cubecl-reduce/src/instructions/maxabs.rs#L80C1-L85C28

PR #596, opened 2025-04-13
Move SPIR-V extension method definitions to separate `rspirv-ext` crate

wingertge:
Migrates the debug and `GLSL.std.450` extensions to an autogenerated [`rspirv-ext`](https://github.com/wingertge/rspirv-ext.git) crate to clean up the generated code and handle things like deduplication automatically.

`rspirv` has merged the newer version we need into main, but still hasn't published. So both crates are still on git for now. Hopefully they can publish soon and I can then publish `rspirv-ext`.

# Testing
Debug info still works as expected and there's no panics or anything like that. It's not possible to unit test debug symbols unfortunately, but I checked several kernels and they appear to be correct.


nathanielsimard (approved):
We only need the dependencies `rspirv` and `rspirv-ext` to be released on `crates.io` to not be blocking on a `cubecl` release.

wingertge:
> We only need the dependencies `rspirv` and `rspirv-ext` to be released on `crates.io` to not be blocking on a `cubecl` release.

Done 👍

PR #602, opened 2025-04-15
[Feat] Add multi stage convolution

wingertge:
Adds a new convolution algorithm with multiple stages. Also updates cudarc since I did that anyways while debugging.

# Testing
All tests pass with the new algorithm.


crates/cubecl-linalg/src/convolution/homogeneous/multi_stage_tma.rs
```diff
@@ -0,0 +1,404 @@
...
+
+use super::base::{
+    config::{self, ConvolutionConfig},
+    implicit_conv,
+};
+
+/// Performs matrix multiplication at the global level, with each plane sharing the same responsibilities
+/// - All planes load data to the stage
+/// - All planes are used in the stage matmul computation
+pub struct MultiStageTmaConvolution<MP: MatmulPrecision, SMM: StageMatmul<MP>> {
```
louisfd, line 47: The doc is wrong

louisfd (approved):
LGTM, I'm just unsure what goes on in MultiStageTmaConvolution::execute and I'd appreciate a doc explaining how the algorithm works exactly 😄

crates/cubecl-linalg/src/convolution/homogeneous/multi_stage_tma.rs
wingertge: I guess it should say `convolution` instead, or did you mean something else? Cause this is not specialized or anything, so all planes are still involved in loading and computation

wingertge:
> LGTM, I'm just unsure what goes on in MultiStageTmaConvolution::execute and I'd appreciate a doc explaining how the algorithm works exactly 😄

I was going to say it's pretty simple, it just uses more stages to prefetch as much data as can fit into shared memory to reduce latency dependency, but then I realized I actually made a mistake and don't need to duplicate the fragments for each stage, only the stage itself and the barrier. So I guess trying to explain it was good

wingertge:
Is there anything else that needs to be done with  it for the merge? I'd like to avoid further merge conflicts.

nathanielsimard:
@wingertge I think we can merge this after the merge conflicts, the release is done :)

wingertge:
@nathanielsimard Done

PR #620, opened 2025-04-16
[Feat] Dynamic shared memory on CUDA

wingertge:
Changes shared memory on CUDA to use a single dynamic buffer with offset pointers, instead of separate declarations. This allows us to fully utilize the shared memory available, instead of being limited to 48k. I did not change HIP, because I can't test it. It would be very easy to add it to HIP though, just need to change two dialect functions.

### Instructions

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -12,11 +12,36 @@ jobs:
       crate: cubecl-common
     secrets: inherit
 
+  publish-cubecl-macros-internal:
+    uses: tracel-ai/cubecl/.github/workflows/publish-template.yml@main
+    with:
+      crate: cubecl-macros-internal
+    secrets: inherit
+
+  publish-cubecl-ir:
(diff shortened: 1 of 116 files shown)


nathanielsimard (commented):
The CubeCL server has access to the amount of shared memory used by a kernel during compilation. I would opt-in dynamic shared memory only if the maximum of static shared memory is reached, leaving kernels with less shared memory to use static shared memory.

wingertge:
> The CubeCL server has access to the amount of shared memory used by a kernel during compilation. I would opt-in dynamic shared memory only if the maximum of static shared memory is reached, leaving kernels with less shared memory to use static shared memory.

That would unnecessarily complicate things by adding two separate code paths in CUDA depending on the size of the shared memory, and it doesn't seem like it has any performance impact whatsoever for smaller sizes since all offsets are constant.

nathanielsimard (approved):

PR #621, opened 2025-04-16
[Perf] Rework `into_contiguous` to improve performance

wingertge:
Adds several new optimizations to significantly speed up `into_contiguous` while also reducing its compute load.
Improves perfomance by approximately 35% while lowering compute load between 20% and 80% depending on how many dimensions are powers of two (which have a special fast path in `FastDivmod`).

### Instructions

### DIFF
--- a/crates/cubecl-linalg/src/tensor/contiguous.rs
+++ b/crates/cubecl-linalg/src/tensor/contiguous.rs
@@ -74,7 +74,7 @@ pub fn index_offset_contiguous_fastdivmod<N: CubePrimitive>(
         offset += ogwl * stride.index(dim);
         remainder = rem;
 
-        comptime![dim -= 1;]
+        comptime![dim = dim.saturating_sub(1);]
     }
 
     offset / tensor.line_size()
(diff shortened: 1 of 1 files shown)


crates/cubecl-linalg/src/tensor/contiguous.rs
```diff
@@ -114,6 +145,45 @@ fn into_contiguous_kernel<N: CubePrimitive>(
...
+    #[unroll]
+    for i in 0..lines_per_thread {
+        let offset = i * line_size;
+        let mut reg = Line::<N>::empty(line_size);
+        #[unroll]
+        for k in 0..line_size {
+            let offset_input = offset_input + offset + k;
+            let offset_input =
+                index_offset_contiguous_fastdivmod::<N>(input, offset_input, &shape, &stride);
+            reg[k] = input[offset_input][0];
```
nathanielsimard, line 174: I guess it's better to not vectorize the input?

wingertge: If the input can be vectorized it's still better to just do 1 to 1 mapping but it's rare. the `pack` kernel is specifically for the case where the input can't be vectorized, to ensure writes are more efficient.

PR #622, opened 2025-04-16
change null value

maxtremblay:
There is a bug with https://github.com/tracel-ai/burn/pull/3033.

This was cause because `min_value` was used instead of `0`. But `abs(min_value)` is quite large was always returned by the algorithm.


antimora:
@nathanielsimard @louisfd why are skipping proper reviews?

louisfd:
For this specific PR I was physically with Max and approved it verbally, but you're right that sometimes maybe we merge a bit too fast 😅

PR #625, opened 2025-04-16
[cpp] Explicitly cast to out elem type in binary ops

syl20bnr:
This prevent the C++ compiler from auto promoting the result type. For instance adding two shorts would be promoted to int which breaks some assumptions in Burn.

### DIFF
--- a/crates/cubecl-cpp/src/shared/binary.rs
+++ b/crates/cubecl-cpp/src/shared/binary.rs
@@ -93,9 +93,18 @@ macro_rules! operator {
                 f: &mut std::fmt::Formatter<'_>,
                 lhs: Lhs,
                 rhs: Rhs,
-                _item: Item<D>,
+                out_item: Item<D>,
             ) -> std::fmt::Result {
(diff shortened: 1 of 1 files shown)


crates/cubecl-cpp/src/shared/binary.rs
```diff
@@ -93,9 +93,10 @@ macro_rules! operator {
                 f: &mut std::fmt::Formatter<'_>,
                 lhs: Lhs,
                 rhs: Rhs,
-                _item: Item<D>,
+                out_item: Item<D>,
             ) -> std::fmt::Result {
-                write!(f, "{lhs} {} {rhs}", $op)
+                let out_elem = out_item.elem();
+                write!(f, "{out_elem}({lhs} {} {rhs})", $op)
```
nathanielsimard, line 99: Only if necessary

nathanielsimard (changes requested):

nathanielsimard (approved):

PR #649, opened 2025-04-28
[Feat] n-dimensional convolution

wingertge:
Makes convolution mostly generic over spatial dimensionality, with a current soft limit of 3D (since TMA instructions change based on dimensionality, so variants need to be bounded). Theoretically, cyclic should work with any dimensionality, but only 1D 2D and 3D are used in reality anyways.

### Instructions

### DIFF
--- a/crates/cubecl-linalg/src/convolution/algorithm/simple_tma.rs
+++ b/crates/cubecl-linalg/src/convolution/algorithm/simple_tma.rs
@@ -8,7 +8,7 @@ use cubecl_core::{
 
 use crate::{
     convolution::{
-        base::{ConvolutionConfigFactory, ConvolutionProblem},
+        base::{ConvolutionConfigFactory, ConvolutionProblem, Dimensionality},
         homogeneous::simple_tma::SimpleTmaConvolutionFamily,
     },
     matmul::components::{
@@ -86,6 +86,7 @@ impl<TMM: TileMatmulFamily> Algorithm for SimpleTmaConvAlgorithm<TMM> {
         handle: &TensorHandleRef<'_, R>,
         ident: InputIdent,
     ) -> TensorHandle<R, E> {
+        let rank = handle.shape.len();
         let mut handle = if has_valid_layout(handle, ident) {
             TensorHandle::from_ref(handle)
         } else {
@@ -94,12 +95,13 @@ impl<TMM: TileMatmulFamily> Algorithm for SimpleTmaConvAlgorithm<TMM> {
         match ident {
             InputIdent::Lhs => handle,
             InputIdent::Rhs => {
-                handle.shape = vec![
-                    handle.shape[0],
-                    handle.shape[1] * handle.shape[2],
-                    handle.shape[3],
+                let k_size = handle.shape[1..rank - 1].iter().product();
+                handle.shape = vec![handle.shape[0], k_size, handle.shape[rank - 1]];
+                handle.strides = vec![
+                    handle.strides[0],
+                    handle.strides[rank - 2],
+                    handle.strides[rank - 1],
                 ];
-                handle.strides = vec![handle.strides[0], handle.strides[2], handle.strides[3]];
                 handle
             }
         }
@@ -108,16 +110,20 @@ impl<TMM: TileMatmulFamily> Algorithm for SimpleTmaConvAlgorithm<TMM> {
 
 fn has_valid_layout<R: Runtime>(handle: &TensorHandleRef<'_, R>, ident: InputIdent) -> bool {
     let stride_align = TMA_STRIDE_ALIGN / handle.elem_size;
+    let rank = handle.shape.len();
 
-    let aligned = handle.strides[..3]
+    let aligned = handle.strides[..rank - 1]
         .iter()
         .all(|stride| stride % stride_align == 0);
 
     let valid_layout = match ident {
-        InputIdent::Lhs => handle.strides[3] == 1,
+        InputIdent::Lhs => handle.strides[rank - 1] == 1,
         InputIdent::Rhs => {
-            let c_major = handle.strides[3] == 1;
-            let kernel_contig = handle.strides[2] * handle.shape[2] == handle.strides[1];
+            let c_major = handle.strides[rank - 1] == 1;
+            let mut kernel_contig = true;
+            for i in 1..rank - 2 {
+                kernel_contig &= handle.strides[i] == handle.strides[i + 1] * handle.shape[i + 1];
+            }
             c_major && kernel_contig
         }
     };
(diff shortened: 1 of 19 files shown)


crates/cubecl-linalg/src/convolution/algorithm/simple_tma.rs
```diff
@@ -152,28 +168,48 @@ pub(crate) fn check_problem_tma(problem: &ConvolutionProblem) -> Result<(), Inva
...
+        Dimensionality::Dim1 => (-(2isize.pow(15)), 2isize.pow(15) - 1),
+        Dimensionality::Dim2 => (-(2isize.pow(7)), 2isize.pow(7) - 1),
+        Dimensionality::Dim3 => (-(2isize.pow(4)), 2isize.pow(4) - 1),
+    };
+
+    let corner = calculate_upper_corner(&problem.padding, &problem.kernel_size, &problem.dilation);
+    for (i, offs) in corner.iter().enumerate() {
+        check_range(
+            *offs as isize,
+            &format!("corner[{i}]"),
```
nathanielsimard, line 181: You could avoid allocating a string just for the checks I guess. Using a function `|| format!("corner[{i}]")`

crates/cubecl-linalg/src/convolution/base.rs
```diff
@@ -172,3 +170,11 @@ impl ConvolutionProblem {
...
     }
 }
+
+/// Spatial dimensionality of an operation
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Debug)]
+pub enum Dimensionality {
+    Dim1,
+    Dim2,
+    Dim3,
+}
```
nathanielsimard, line 180: So Conv1d => Dimensionality::Dim1, Conv2d = Dimensionality::Dim2 ?

crates/cubecl-linalg/src/convolution/launch.rs
```diff
@@ -25,28 +25,28 @@ type Output<Alg, MP> =
...
-pub fn launch_conv2d_nhwc<R: Runtime, MP: MatmulPrecision, Alg: Algorithm>(
+pub fn launch_conv<R: Runtime, MP: MatmulPrecision, Alg: Algorithm, const N_SPATIAL: usize>(
     client: &ComputeClient<R::Server, R::Channel>,
     input: &TensorHandleRef<'_, R>,
     weight: &TensorHandleRef<'_, R>,
     bias: &Option<TensorHandleRef<'_, R>>,
     out: &TensorHandleRef<'_, R>,
-    args: ConvolutionArgs,
+    args: ConvolutionArgs<N_SPATIAL>,
 ) -> Result<(), ConvLaunchError>
```
nathanielsimard, line 50: Not sure how much the const generic affects compilation speed, but maybe creating a function without it and redispatching to it would help.

crates/cubecl-linalg/src/convolution/loader/im2col_tma.rs
```diff
@@ -80,30 +81,78 @@ impl<MP: MatmulPrecision, G: ConvGemmConfig> TmaIm2colLoader<MP, G> {
...
 
-            let in_h = (this.map.h_offset * config.stride(0)) as i32 - config.padding(0);
-            let in_w = (this.map.w_offset * config.stride(1)) as i32 - config.padding(1);
+            let spatial_dims = comptime![this.map.spatial_offsets.len()];
+            let mut in_offs = Sequence::<i32>::new();
+
+            #[unroll]
+            for dim in 0..spatial_dims {
+                let dim = unwrap(dim);
+                let offs = this.map.spatial_offsets.index(dim) * comptime![config.stride(dim)];
```
nathanielsimard, line 90: Having the strides as comptime value seems a bit much. Maybe that's not the tensor strides though and that would be fine.

crates/cubecl-linalg/src/convolution/loader/im2col_tma.rs
```diff
@@ -80,30 +81,78 @@ impl<MP: MatmulPrecision, G: ConvGemmConfig> TmaIm2colLoader<MP, G> {
...
-                    in_w,
-                    channel_start as i32,
-                    offset_y as u16,
-                    offset_x as u16,
-                );
+                match config.dimensionality() {
+                    Dimensionality::Dim1 => {
+                        let offset = k_idx * config.dilation(0);
+
+                        bar.tma_load_im2col_3d(
```
nathanielsimard, line 106: 3d means rank 3 (with batch size and channel I guess?) Maybe adding a comment would be great.

crates/cubecl-linalg/src/convolution/loader/im2col_tma.rs
```diff
@@ -116,3 +165,27 @@ impl<MP: MatmulPrecision, G: ConvGemmConfig> TmaIm2colLoader<MP, G> {
         TmaIm2colReader::<MP>::new(*this.stages.index(stage_idx), InputIdent::Lhs)
     }
 }
+
+#[cube]
+pub(crate) fn div_mod_seq(pos: u32, shape: &Sequence<FastDivmod>) -> (u32, Sequence<u32>) {
```
nathanielsimard, line 170: What is the output sequence?

crates/cubecl-linalg/src/convolution/reader/im2col.rs
```diff
@@ -109,31 +112,63 @@ impl<E: Numeric> Im2colReader<E> {
...
-        let rem = view_k / self.shape_channel;
-        let kernel_x = rem % kernel_w;
-        let kernel_y = rem / kernel_w;
+        let mut rem = view_k / self.shape_channel;
+
+        let spatial_dims = comptime![self.shapes_spatial.len()];
+        let mut in_pos = Sequence::<i32>::new();
+
+        #[unroll]
+        for i in 0..spatial_dims {
```
nathanielsimard, line 124: I would document a bit more what is the spatial dims.

crates/cubecl-linalg/src/convolution/reader/im2col.rs
```diff
@@ -149,3 +184,9 @@ impl<E: Numeric> Im2colReader<E> {
         self.tensor.read(position)
     }
 }
+
+#[allow(unused_variables)]
+#[cube]
+fn unwrap(v: u32) -> comptime_type!(u32) {
+    intrinsic!(|_| v.constant().expect("Must be constant").as_u32())
+}
```
nathanielsimard, line 192: We could have a function for index (u32) which is `u32.as_const_unchecked() -> u32`

crates/cubecl-linalg/src/convolution/base.rs
wingertge: Yes

crates/cubecl-linalg/src/convolution/loader/im2col_tma.rs
wingertge: The config stride is the convolution stride, and the config is always comptime

crates/cubecl-linalg/src/convolution/loader/im2col_tma.rs
wingertge: Didn't think it would need a comment because it's for internal use in the convolution kernels only. It decomposes the linear `m` index into the spatial (depth, height, width) indices and a remainder (the batch). So the sequence is the coordinate along each of the dimensions in the `Sequence`. It's just a generic version of
```rust
let (rem, out_x) = (pos / width, pos % width);
let (rem, out_y) = (rem / height, rem % height);
(rem, [out_y, out_x])
```
In that example, the sequence would contain `[height, width]`.

crates/cubecl-linalg/src/convolution/reader/im2col.rs
wingertge: I think this might just be a language barrier. It's spatial, as in space, as in depth, height, width or z, y, x. As opposed to the batch and channels, which don't represent a position in space. It should be pretty self-explanatory.

crates/cubecl-linalg/src/convolution/reader/im2col.rs
wingertge: I think so too, but I'd do it in a different PR probably. Alternatively, we might want to just take an expanded u32 and unwrap like we do with some other things rn.

crates/cubecl-linalg/src/convolution/loader/im2col_tma.rs
nathanielsimard: I would add your comment in the code, just for clarity if we have to update/change the code later as context.

crates/cubecl-linalg/src/convolution/reader/im2col.rs
nathanielsimard: So anything not batch related

nathanielsimard (approved):

crates/cubecl-linalg/src/convolution/reader/im2col.rs
wingertge: Not batch and not channels, yes.

PR #654, opened 2025-04-29
Sanitize init element const

nathanielsimard:
It makes it easier to read the generated C++ & WGSL code.


nathanielsimard:
Burn commit: https://github.com/tracel-ai/burn/pull/3116

PR #657, opened 2025-04-30
Fix deadlock by really properly not capturing locks

ArthurBrussee:
I was experiencing a deadlock on WASM. I'm guessing other platforms might hit this but usually don't use an actual async runtime  or if they do a multi threaded one.

The problem is mixing async code and regular locks. It's important to _not_ hold a lock across await points. We tried doing this by marking futures as 'static but that's not quite what you want. This would still allow capturing &self as long as &self was 'static (eg. by cloning an arc). The fusion client was incorrectly capturing things leading to a deadlock.

The proper thing to do is `use<>` bounds. I've annotated futures where possible. However, `use<>` bounds in traits are still unstable. There I've gone back to a boxed future, which also doesn't capture parameters by default. Rust 1.88 stabilizes this so maybe can then remove the unneeeded boxing again.

This is only the cubecl changes, burn has the more changes, see https://github.com/tracel-ai/burn/pull/3123

### DIFF
--- a/crates/cubecl-cuda/src/compute/server.rs
+++ b/crates/cubecl-cuda/src/compute/server.rs
@@ -1,4 +1,5 @@
 use cubecl_core::benchmark::ProfileDuration;
+use cubecl_core::future::{self, DynFut};
 use cubecl_cpp::{
     CudaCompiler, cuda::arch::CudaArchitecture, formatter::format_cpp, shared::CompilationOptions,
 };
@@ -36,7 +37,6 @@ use cudarc::driver::sys::{
 };
 use cudarc::driver::sys::{CUfunc_st, CUtensorMapInterleave};
 use std::collections::HashMap;
-use std::future::Future;
 use std::path::PathBuf;
 use std::{ffi::CStr, os::raw::c_void};
 use std::{ffi::CString, mem::MaybeUninit};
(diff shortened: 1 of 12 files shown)


crates/cubecl-cuda/src/compute/server.rs
```diff
@@ -335,8 +311,8 @@ impl ComputeServer for CudaServer {
             // One option is to create a dummy kernel with 1 thread that launches the real kernel with the dynamic dispatch settings.
             // For now, just read the dispatch settings from the buffer.
             CubeCount::Dynamic(binding) => {
-                let data = self.read_sync(binding);
-                let data = bytemuck::cast_slice(&data);
+                let data = future::block_on(self.read_async(vec![binding]));
+                let data = bytemuck::cast_slice(&data[0]);
```
nathanielsimard, line 315: Why we need to access [0]?

ArthurBrussee: read_async reads multiple bindings. read_sync was only doing one. I don't think it's worth keeping both around as these functions are tricky to get right.

(related - merging read and read_tensor would be good but not sure how viable that is)

nathanielsimard (approved):

PR #670, opened 2025-05-06
Fix wasm by always using cfg(std_io)

ArthurBrussee:
WASM is currently broken, as some code assumes feature = "std" means there is a filesystem and env variables are available. This is not true on WASM and on mobile platforms.

This fixes things by standardizing on a std_desktop_platform cfg already in use for autotune caching.

I've left logging with stdout and stderr as cfg(std) as that _does_ have functionality on WASM/mobile platforms. It might be easier to put things through the `log` crate though as that has cross platform configuration options of it's own

### DIFF
--- a/crates/cubecl-runtime/Cargo.toml
+++ b/crates/cubecl-runtime/Cargo.toml
@@ -41,7 +41,7 @@ serde = { workspace = true }
 toml = { workspace = true, optional = true }
 dirs = { workspace = true, optional = true }
 
-# Persistent cache deps - has to match the autotune_persistent_cache cfg.
+# Persistent cache deps - has to match the cfg(std_desktop_platform) cfg.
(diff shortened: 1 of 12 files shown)


nathanielsimard (commented):
`desktop` isn't really the best word, since that's probably the same feature flag you would want for a server too. Maybe a more general feature flag name like `std_file_system` or `std_fs` would be more accurate.

ArthurBrussee:
True but then it's not about just filesystems, it's also about environment variables. Maybe `std_io`?

nathanielsimard:
> True but then it's not about just filesystems, it's also about environment variables. Maybe `std_io`?

Yup good point, `std_io` is perfect!

nathanielsimard (approved):

PR #674, opened 2025-05-06
Feat/recursive profiling

nathanielsimard:

### Instructions

https://github.com/tracel-ai/burn/pull/3161


crates/cubecl-runtime/src/client.rs
```diff
@@ -227,10 +233,21 @@ where
     /// Nb: this function will only allow one function at a time to be submitted when multithrading.
     /// Recursive measurements are not allowed and will deadlock.
     pub fn profile(&self, func: impl FnOnce()) -> ProfileDuration {
-        let guard = self.state.profile_lock.lock();
-        self.channel.start_profile();
+        #[cfg(feature = "std")]
+        let (stream_id, should_init) = self.state.queue.aquire_profile_priority();
```
ArthurBrussee, line 237: This will break on WASM - thread::sleep panics. Since it's guaranteed to be single threaded maybe can skip this entirely, not sure.

crates/cubecl-wgpu/src/compute/stream.rs
```diff
@@ -121,25 +126,23 @@ impl WgpuStream {
...
+                    timestamp_writes: None,
                 })
-                .forget_lifetime()
+                .forget_lifetime();
+
+            for token in self.timestamps_device_init.drain(..) {
+                if let KernelTimestamps::Device { query_set } =
+                    &mut self.timestamps.get_mut(&token).unwrap()
+                {
+                    pass.write_timestamp(query_set, 0);
```
ArthurBrussee, line 141: write_timestamp is native only unlike timestamp_writes in the compute pass. This was done as write_timestamp without context is too ambigious, GPU's might re-order things, but a pass has a beginning and end. See https://github.com/gpuweb/gpuweb/pull/4370

crates/cubecl-wgpu/src/compute/server.rs
```diff
@@ -164,18 +165,19 @@ impl ComputeServer for WgpuServer {
...
         };
 
-        let currently_profiling = self.stream.is_profiling();
+        let profiling_token = self.stream.get_profiling_token();
 
         if profile_level.is_some() {
             // Add in current time if currently profiling.
-            if currently_profiling {
-                let profile = self.stream.stop_profile();
+            if let Some(token) = profiling_token {
```
ArthurBrussee, line 172: Thnis all seems unnecesarry now that you can profile recursively? If that works properly, you should be able to just start/stop a new profile in the execute call right? Ideally that would even be moved out of each individual server and just work by something wrapping the execute() calls.

crates/cubecl-runtime/src/client.rs
```diff
@@ -227,10 +233,21 @@ where
...
-        let result = self.channel.end_profile();
+
+        let result = self.channel.end_profile(token);
+
+        #[cfg(feature = "std")]
+        if should_init {
+            self.state.queue.set_profile_priotity_init_token(token);
+        }
+
         let result = match self.properties().time_measurement() {
```
ArthurBrussee, line 251: I don't quite understood this when this was added. So this is basically turning a Device measurement into a System measurement if needed? The autotune code will already block on the future on non-wasm platforms, so this just seems like a redundant thing to do?

nathanielsimard: It's mostly to avoid having other threads register more tasks into the same device while performing autotuning on another thread.

ArthurBrussee: Sorry thought I left a comment range but bit unclear - for reference I mean the future::block_on and the seperate properties().time_measurement. 

For a server doing a system profile, this will do nothing, as that's already sync, no work at all is happening in the block_on.

For a server doing a device profile, this _shouldn't_ do anything, if submitting other things messes with the timing it's just broken? It's only measuring actual GPU time within a scope. If properly scoped no other GPU work is running at that time. 

Sneakily blocking an async function doesn't seem great 😅It's also incorrect when properties say to use Device measurement but the server returns a System measurement, as it just passes through the system measurement. It should be up to the server what kind of time measurement is happening.

If it really messes things up to submit more work while waiting for device measurements, then maybe the lock should be kept until the ProfileDuration is dropped?

crates/cubecl-wgpu/src/compute/stream.rs
```diff
@@ -310,47 +373,71 @@ impl WgpuStream {
...
+                    query_set_start.num_ref -= 1;
+
+                    self.encoder.resolve_query_set(
+                        &query_set_start.query_set,
+                        0..1,
+                        resource.buffer(),
+                        resource.offset(),
+                    );
+
+                    let query_set_end = self.timings.query_sets.get_mut(&start).unwrap();
```
ArthurBrussee, line 418: Doesn't this overwrite the start timestamp? Both are written to resource[offset()] it seems?

nathanielsimard: Fixed

ArthurBrussee (commented):
Hell yeah thanks for adding this, just tested it on WASM and it works! (wih #670 also patched in anyway).

Few minor things but looks good

ArthurBrussee (approved):

PR #675, opened 2025-05-07
[Feat] Add fp8, fp6 and theoretically fp4

wingertge:
Adds basic support for fp8, fp6 and fp4 formats suported by CUDA. fp4 isn't supported properly, because the compiler makes many assumptions about types that don't work for sub-byte types like fp4. I'll make sub-byte types work in a separate PR.


crates/cubecl-core/src/runtime_tests/minifloat.rs
```diff
@@ -0,0 +1,212 @@
...
+    expected.extend(expect_1[..num_out].iter().copied());
+
+    // TODO: Eventually add approx comparison that can deal with arbitrary floats. Manually
+    // double check for now
+    let actual_2 = client.read_one(handle1.binding());
+    let actual_2 = F::from_bytes(&actual_2);
+    println!("actual_2: {actual_2:?}");
+
+    assert_eq!(actual, &expected);
+    //assert_eq!(&actual_2[..num_out], &data[..num_out]);
```
nathanielsimard, line 77: Is there a reason to not keep the assertion?

crates/cubecl-core/src/runtime_tests/minifloat.rs
```diff
@@ -0,0 +1,212 @@
...
+                4,
+            );
+        }
+
+        // #[test]
+        // fn test_fp4() {
+        //     let client = TestRuntime::client(&Default::default());
+        //     cubecl_core::runtime_tests::minifloat::test_fp4::<TestRuntime>(client.clone(), 2);
+        //     cubecl_core::runtime_tests::minifloat::test_fp4::<TestRuntime>(client.clone(), 4);
+        // }
```
nathanielsimard, line 202: Dead code?

crates/cubecl-cpp/src/cuda/dialect.rs
```diff
@@ -89,8 +98,37 @@ impl DialectTypes<Self> for CudaDialect {
...
+        let mut items_deduplicated = HashSet::new();
+
+        for item in items {
+            let mut item = *item;
+            match item.elem() {
+                Elem::FP4(_) => {
+                    item.elem = Elem::FP4(FP4Kind::E2M1);
+                }
+                Elem::FP4x2(_) => {
+                    item.elem = Elem::FP4x2(FP4Kind::E2M1);
```
nathanielsimard, line 111: I'm unsure if we want to have `fp4x2` or any `x2` types. For the same reason, we don't have `f16x2` and `bf16x2`. We have those in the C++ compiler, of course, but not on the frontend side since vectorization is programmed with the line size instead. I guess the reason you included it was to read and write values with 8 bits, but maybe we could use something else instead? The concept of a minimum line size, to which we have to pad if it's uneven, would make sense for reading data back from the GPU.

crates/cubecl-cpp/src/shared/element.rs
```diff
@@ -63,6 +124,13 @@ impl<D: Dialect> Elem<D> {
             Elem::_Dialect(_) => 0,
         }
     }
+
+    pub const fn size_bits(&self) -> usize {
+        match self {
+            Elem::FP4(_) => 4,
+            other => other.size() * 8,
+        }
+    }
```
nathanielsimard, line 133: Is FP6 encoded on 1 full byte?

crates/cubecl-cpp/src/shared/instruction.rs
```diff
@@ -651,8 +653,334 @@ for ({i_ty} {i} = {start}; {i} {cmp} {end}; {increment}) {{
...
+            Instruction::SpecialCast(UnaryInstruction { input, out }) => {
+                // Only supported in CUDA so I'm putting it here. Move to dialect if necessary.
+                special_cast::<D>(f, input, out)
+            }
+        }
+    }
+}
+
+/// special cast function for recursive conversion in the case of minifloat to minifloat conversion
+fn special_cast<D: Dialect>(
```
nathanielsimard, line 665: An example would help, what input becomes what output, maybe in a table.

crates/cubecl-cpp/src/shared/instruction.rs
```diff
@@ -651,8 +653,334 @@ for ({i_ty} {i} = {start}; {i} {cmp} {end}; {increment}) {{
...
+            out_opt.elem()
+        )
+    })
+}
+
+fn handle_unroll<D: Dialect>(
+    f: &mut fmt::Formatter,
+    out: Variable<D>,
+    mut op: impl FnMut(&mut fmt::Formatter, usize) -> fmt::Result,
+) -> fmt::Result {
```
nathanielsimard, line 952: All of those cast functions could be moved into another module just for that.

crates/cubecl-cpp/src/shared/item.rs
```diff
@@ -54,12 +57,27 @@ impl<D: Dialect> Item<D> {
...
                 native: self.native,
             },
             Elem::BF16 => Item {
-                elem: Elem::BF162,
+                elem: Elem::BF16x2,
+                vectorization: self.vectorization / 2,
+                native: self.native,
+            },
+            Elem::FP4(kind) => Item {
+                elem: Elem::FP4x2(kind),
```
nathanielsimard, line 70: Is there versions that work on 32 bits?

crates/cubecl-cuda/src/compute/server.rs
```diff
@@ -644,7 +644,7 @@ impl CudaContext {
         let compute_kernel = kernel_compiled.repr.as_ref().unwrap();
         let cube_dim = kernel_compiled.cube_dim;
         let fast_math = compute_kernel.flags.inst_fast_math;
-        let arch = format!("--gpu-architecture=sm_{}", self.arch);
+        let arch = format!("--gpu-architecture=sm_{}a", self.arch);
```
nathanielsimard, line 647: Why a ?

crates/cubecl-core/src/runtime_tests/minifloat.rs
antimora: My ignorant question. Will `println` be eventually removed?

wingertge: It's only for manually checking when running the test with `--nocapture`. I'd like to add a way to check float equality with arbitrary precision but we don't current have that and it would be rather complicated. The equality would fail the test immediately because `-2.1` is rounded to `-2.0` for example, since the next interval for fp8 is `2.25` which is further away. It's just to ensure the value correctly converts back as a sanity check.

crates/cubecl-cuda/src/compute/server.rs
wingertge: So we can use architecture specific features like fp6. You did remind me to test on 3080 though and it seems like that's only supported as a tag for 90+. So I'll have to gate it. I was hoping they just added it as an alias to older arches for simplicity.

wingertge: They did add an alias for the SASS target even though it technically only means something for the PTX generation, just so you can use the same for both. But I guess that didn't apply for previous arches.

crates/cubecl-core/src/runtime_tests/minifloat.rs
wingertge: Sort of, I'd leave it there as a reminder to add tests for fp4 once sub-byte types are properly supported

crates/cubecl-cpp/src/cuda/dialect.rs
wingertge: This comment is on the C++ dialect, which is where we do have `F16x2` 😅 I'm assuming you're talking about the type though. I only added that to allow `bytemuck` to properly convert buffers without losing semantics like would if we just have it represented as u8. It doesn't implement `CubePrimitive` so it's not used on the GPU side, only on CPU for conversion.

crates/cubecl-cpp/src/shared/element.rs
wingertge: yep, the top two bits are just `0`. Similar to `tf32`, except the conversion actually does zero out the bits while tf32 just ignores them (since `fp6` is actually a size change and `tf32` is just a reinterpet).

crates/cubecl-cpp/src/shared/instruction.rs
wingertge: Agreed, it really balooned to a much larger amount of code than expected once I realized how annoying the CUDA APIs are

crates/cubecl-cpp/src/cuda/dialect.rs
nathanielsimard: Yeah I was refering to the type declaration in `cubecl_common`. Do you think we could make a type like `Packed<e2m1, 2>` to be used with all subtypes ?

```rust
pub struct Packed<E: SubByteElement, const FACTOR: usize> {
    data: [E; FACTOR],
}
```
where `core::mem::size_of(data) >= 1`

wingertge: The problem is that we can't really declare something like that in current Rust. You can't declare sub-byte types at all, so `fp4x2` would be 2 bytes with a type like that. I'll look into this more when I work on the sub-byte migration PR.

wingertge: I might look into https://crates.io/crates/packed_struct to see if we can do something with that.

PR #676, opened 2025-05-07
Feat: Add sync_plane instruction

louisfd:
- [Breaking] Rename sync_units to sync_cube, to communicate better the scope of the synchronization
- Added sync_plane to synchronize within a plane. Since this is not supported everywhere, I made two versions:
  - One that can fallback to sync_cube if sync_plane is not supported. In some cases this may simply incur a minor performance cost. 
  - One that will panic if sync_plane is not supported. In some algorithms, like those that have plane specialization, falling back to sync_cube would simply be invalid. 

### Instructions

- [x] Submit a PR in burn with your fixes and link it here: https://github.com/tracel-ai/burn/pull/3163


crates/cubecl-core/src/frontend/synchronization.rs
```diff
@@ -13,16 +14,47 @@ use crate::{
...
+/// when native support is unavailable.
+///
+/// In some cases, falling back may only incur a minor performance cost.
+/// In others, it can break correctness or even cause deadlocks —
+/// especially when warp specialization is involved.
+pub enum SyncPlaneMode {
+    Strict,   // Requires native warp sync, or fails at kernel compilation
+    Fallback, // Allows fallback to Cube-wide sync, may break correctness
+}
+
```
nathanielsimard, line 57: Not sure I would allow a fallback in the `sync_plane` function. If you need some in your context. You may create a wrapper for your use case with a comptime value passed as input to determine which sync function to use.

crates/cubecl-cpp/src/metal/dialect.rs
```diff
@@ -835,6 +835,17 @@ impl DialectInstructions<Self> for MslDialect {
...
     }
+
+    fn compile_instruction_sync_warp(
+        f: &mut std::fmt::Formatter<'_>,
+        fallback_allowed: bool,
+    ) -> std::fmt::Result {
+        if fallback_allowed {
+            Self::compile_instruction_sync_threads(f)
+        } else {
+            panic!("Synchronization within a plane is not supported on Metal")
```
nathanielsimard, line 846: Why `simdgroup_barrier(mem_flags::mem_none);` can't be used? It seems to sync the execution of 32 units within a plane

nathanielsimard (changes requested):

crates/cubecl-spirv/src/sync.rs
```diff
@@ -6,14 +6,22 @@ use crate::{SpirvCompiler, SpirvTarget};
...
                 let semantics =
                     MemorySemantics::ACQUIRE_RELEASE | MemorySemantics::WORKGROUP_MEMORY;
                 let semantics = self.const_u32(semantics.bits());
                 self.control_barrier(scope, scope, semantics).unwrap();
             }
+            Synchronization::SyncPlaneFallback | Synchronization::SyncPlaneStrict => {
+                // Adopting wgpu semantics
+                let scope = self.const_u32(Scope::Subgroup as u32);
+                let semantics =
+                    MemorySemantics::ACQUIRE_RELEASE | MemorySemantics::WORKGROUP_MEMORY;
```
nathanielsimard, line 21: Could you make the real one by changing the `MemorySemantics`?

nathanielsimard (commented):
You would also need a device flag to say if `sync_plane` is available, like atomics, wmma and other instructions.

nathanielsimard (approved):

PR #679, opened 2025-05-07
Vectorized random kernels

Cielbird:
Calculate random tensors with vectorization

When vectorization is used, it adds a small amount correlation between some adjacent elements. This is very minor and can be mostly ignored. The wald_wolfowitz_runs_test has been adjusted because of this.

### DIFF
--- a/crates/cubecl-random/src/bernoulli.rs
+++ b/crates/cubecl-random/src/bernoulli.rs
@@ -23,26 +23,34 @@ impl<E: CubeElement + Numeric> PrngRuntime<E> for Bernoulli<E> {
         write_index_base: u32,
         n_invocations: u32,
         #[comptime] n_values_per_thread: u32,
+        #[comptime] line_size: u32,
         state_0: &mut u32,
         state_1: &mut u32,
         state_2: &mut u32,
         state_3: &mut u32,
-        output: &mut Tensor<E>,
+        output: &mut Tensor<Line<E>>,
     ) {
         let prob = args.probability;
 
+        let mut output_line = Line::empty(line_size);
+
         #[unroll(n_values_per_thread <=8)]
-        for i in 0..n_values_per_thread {
-            *state_0 = taus_step_0(*state_0);
-            *state_1 = taus_step_1(*state_1);
-            *state_2 = taus_step_2(*state_2);
-            *state_3 = lcg_step(*state_3);
+        for line_index in 0..n_values_per_thread / line_size {
+            // vectorization
+            #[unroll]
+            for i in 0..line_size {
+                *state_0 = taus_step_0(*state_0);
+                *state_1 = taus_step_1(*state_1);
+                *state_2 = taus_step_2(*state_2);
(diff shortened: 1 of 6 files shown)


crates/cubecl-random/src/bernoulli.rs
```diff
@@ -23,26 +23,34 @@ impl<E: CubeElement + Numeric> PrngRuntime<E> for Bernoulli<E> {
...
         state_2: &mut u32,
         state_3: &mut u32,
-        output: &mut Tensor<E>,
+        output: &mut Tensor<Line<E>>,
     ) {
         let prob = args.probability;
 
+        let mut output_line = Line::empty(line_size);
+
         #[unroll(n_values_per_thread <=8)]
```
louisfd, line 37: The condition for unrolling did not follow the vectorization change. The goal is simply to control the number of iterations, so if you reduce the number of iteration the condition should also be reduced. How about:
```rust
let num_iterations = n_values_per_thread / line_size;
#[unroll(num_iterations <= 8)]
for line_index in 0..num_iterations {...}
```

crates/cubecl-random/src/normal.rs
```diff
@@ -19,51 +19,62 @@ impl<E: CubeElement + Numeric> PrngRuntime<E> for Normal<E> {
...
     ) {
         let mean = f32::cast_from(args.mean);
         let std = f32::cast_from(args.std);
 
         let should_unroll = n_values_per_thread <= 16;
 
+        let mut output_line_0 = Line::empty(line_size);
+        let mut output_line_1 = Line::empty(line_size);
+
         #[unroll(should_unroll)]
```
louisfd, line 37: same, but adjust with the factor 2 to get the same threshold

crates/cubecl-random/src/tests_utils.rs
```diff
@@ -76,6 +76,9 @@ pub fn assert_normal_respects_68_95_99_rule<E: Numeric>(data: &[E], mu: f32, s:
         let expected = percent * data.len() as f32 / 100.;
         assert!(f32::abs(count as f32 - expected) < 2000.);
     };
+    for s in &stats {
```
louisfd, line 79: I don't think we want to merge a print

crates/cubecl-random/src/uniform.rs
```diff
@@ -18,35 +18,44 @@ impl<E: CubeElement + Numeric> PrngRuntime<E> for Uniform<E> {
...
     ) {
         let lower_bound = args.lower_bound;
         let upper_bound = args.upper_bound;
 
         let should_unroll = n_values_per_thread <= 8;
         let scale = upper_bound - lower_bound;
 
+        let mut output_line = Line::empty(line_size);
+
         #[unroll(should_unroll)]
```
louisfd, line 36: same

louisfd (changes requested):
Thanks! There is mainly the should_unroll to adjust

PR #680, opened 2025-05-08
Feat/matmul/ordered double buffering

louisfd:
The "ordered" matmul variant uses the LHS fragments as one of the two buffers in its double buffering scheme. This helps reduce shared memory usage. It relies on a loader that ensures each plane only accesses the data it has loaded itself. 

This approach uses the new "Ordered Tiling Order", which guarantees that a plane loads all the tiles for a given k iteration before moving to the next one. This ordering makes it possible to start overwriting shared memory as soon as a k iteration is done, without waiting for the full computation to finish.

Also in this PR: 
- Fix the mut variables created by advance_view (RuntimeCell)
- Metal: use simdgroup_barrier instead of threadgroup_barrier after store and before cast
- Support having different number of stages for Lhs and Rhs
- Don't uselessly use two sets of tiles
- Fix underflow in events with saturating_sub
- Fix double buffering tests by not running multi-row if m != 2*n
- Add doc to tiling order

### Instructions


raiderss:
Fantastic work on this project! This could integrate well with our FiveM resources at Eyes Store Tebex.

crates/cubecl-linalg/src/matmul/components/stage/layout.rs
```diff
@@ -29,10 +29,73 @@ pub trait TilingOrder: 'static + Send + Sync + Clone + Copy {
 }
 
 #[derive(CubeType, Clone, Copy)]
+/// Tiles laid out in row-major order.
+///
+/// Each tile is contiguous, and tiles are placed side by side,
+/// row by row (left to right, top to bottom).
+/// /// Example tile indices:
```
nathanielsimard, line 36: double /// but nice visual !

crates/cubecl-linalg/src/matmul/tests/cmma_matmul/matmul_test_launcher.rs
```diff
@@ -71,6 +71,9 @@ pub fn test_matmul_algorithm<A, P, R>(
         &out.strides,
         out.strides.len() - 1,
     );
+    problem.lhs_line_size = 4;
+    problem.rhs_line_size = 1;
+    problem.out_line_size = 1;
```
nathanielsimard, line 76: Should we keep that?

PR #684, opened 2025-05-10
Fix buffer offset alignment and total size calculation

wingertge:
Fixes a small oversight that was introduced while I was trying to clean up the code before the PR merge, as well as a misunderstanding about which alignment needs to be respected for buffer offsets. This is required to properly handle multi-tensor handles on wgpu.

### Instructions


crates/cubecl-cuda/src/runtime.rs
```diff
@@ -230,6 +230,7 @@ impl Runtime for CudaRuntime {
 
         let mut sorted = strides.to_vec();
         sorted.sort();
+        sorted.reverse();
```
nathanielsimard, line 233: Is it required?

crates/cubecl-linalg/src/tensor/contiguous.rs
```diff
@@ -208,8 +208,42 @@ pub fn into_contiguous<R: Runtime, E: CubePrimitive>(
...
+    // Vectorization is only enabled when the last dimension is contiguous.
+    let rank = input.strides.len();
+    let vectorization_factor = tensor_line_size_parallel(
+        R::supported_line_sizes().iter().cloned(),
+        input.shape,
+        input.strides,
+        rank - 1,
+    );
+    let num_vecs = num_elems / vectorization_factor as usize;
+    let approx_sm = 64;
```
nathanielsimard, line 236: Btw the number of SM is in the client hardware properties now.

crates/cubecl-wgpu/src/compute/server.rs
```diff
@@ -221,8 +221,21 @@ impl ComputeServer for WgpuServer {
...
+            .map(|it| it.shape.iter().product::<usize>() * it.elem_size)
+            .collect::<Vec<_>>();
         let bindings = bindings.into_iter().map(|it| it.binding).collect();
-        self.read(bindings)
+        let data = self.read(bindings);
+        Box::pin(async move {
+            let mut data = data.await;
+
+            for (data, expected_size) in data.iter_mut().zip(expected_sizes) {
+                data.truncate(expected_size);
```
nathanielsimard, line 234: Why is there an expected size now?

crates/cubecl-cuda/src/runtime.rs
wingertge: yes because the strides need to be smallest last not largest last. sort + reverse is the fastest way to get a vector sorted by largest first.

crates/cubecl-wgpu/src/compute/server.rs
wingertge: Because for tensors we actually know the size of the element, so we can properly truncate the read bytes. CUDA already does this because of how the function is implemented, so this just matches functionality. Previously, burn would need to handle this because CubeCL didn't know the element size or shape and couldn't tell when a handle was padded for alignment.

nathanielsimard (approved):

PR #686, opened 2025-05-13
Unit Matmul

louisfd:
Unit Matmul

The goal is to replace Tiling2d (eventually deleting it). 
It will use the same batch and global components as cmma matmuls, but the behaviour changes at the stage and tile levels. 
- Unit matmul (stage level): Divides the stage memory into unit-wise tiles. That is, there are as many tiles as there are units. This has impact on the unloader (now writer), we can't use the tilewise unloader. So I had to refactor unloaders so the stage decides which to use as an associated type. Also they're simpler and renamed writers. 
- Register matmul (tile level): Stores tiles into arrays. Also have a mode to load line lazily, but it seems less good. Supports inner and outer product versions for any layout combination. 
- Refactored selectors. Burn will need to be updated. 

### Instructions

https://github.com/tracel-ai/burn/pull/3214


crates/cubecl-linalg/src/convolution/base.rs
```diff
@@ -37,20 +36,20 @@ pub trait Convolution<MP: MatmulPrecision>: 'static + Send + Sync {
     type Config: ConvGemmConfig;
     type AccumulatorLoader: AccumulatorLoader<MP>;
 
-    type Out: OutputLoader<MP::EO>;
+    type Out: Writer<MP::EO>;
```
nathanielsimard, line 39: `GlobalWriter`

crates/cubecl-linalg/src/matmul/tests/test_macros/cmma/suite.rs
```diff
@@ -27,7 +27,8 @@ pub fn test_algo<A: Algorithm, P: TestPrecision, R: Runtime>(
         m: problem.m as usize,
         n: problem.n as usize,
         k: problem.k as usize,
-        batches: (vec![2], vec![2]),
+        batches: (vec![1], vec![1]),
+        // batches: (vec![2], vec![2]),
```
nathanielsimard, line 31: deadcode

PR #698, opened 2025-05-23
Feat/mlir initial implementation of a MLIR CPU backend

marcantoinem:
It's a rough backend for the moment with some syntax not supported and a lot of optimization are still needed, but it's in a mergeable state and further syntax support will come in more PRs.

# TODO before this can be merged
- [x] Support all simple operation in cubecl_ir
- [x] Implement efficient simulation of CubeCount with a thread pool
- [x] Integrate the fork of Melior
- [x] Add a compilation cache 

The burn-cpu backend needs some more things to be implemented


crates/cubecl-cpu/LICENSE-APACHE
```diff
@@ -0,0 +1,201 @@
+                              Apache License
```
nathanielsimard, line 1: We can create a simlink to the root licenses

crates/cubecl-cpu/src/compiler/mod.rs
```diff
@@ -0,0 +1,33 @@
+pub mod mlir;
+
+use cubecl_core::{Compiler, ExecutionMode, ir, prelude::KernelDefinition};
+use mlir::MlirEngine;
+
+#[derive(Clone, Debug, Default)]
+pub struct MLIRCompiler {}
```
nathanielsimard, line 7: It's probably better to name it `MlirCompiler` to respect Rust naming convention.

crates/cubecl-cpu/src/runtime.rs
```diff
@@ -0,0 +1,117 @@
...
+
+type Server = CpuServer;
+type Channel = MutexComputeChannel<Server>;
+
+fn create_client(options: RuntimeOptions) -> ComputeClient<Server, Channel> {
+    let max_thread = std::thread::available_parallelism()
+        .map(|n| n.get() as u32)
+        .unwrap_or(1);
+    // TODO determine how parallelism should be distributed between cube_count and cube_dim
+    // For now cube_count is parallel and cube_dim sequential
```
nathanielsimard, line 39: It is the opposite, `cube_dim` needs be parallel and `cube_count` can be whatever.

crates/cubecl-cpu/src/runtime.rs
```diff
@@ -0,0 +1,117 @@
...
+    let max_cube_count = CubeDim::new(max_thread, 1, 1);
```
nathanielsimard, line 40: CubeCount doesn't use the struct `CubeDim`

crates/cubecl-cpu/src/runtime.rs
```diff
@@ -0,0 +1,117 @@
...
+    fn supported_line_sizes() -> &'static [u8] {
+        &[8, 1, 1, 1]
+    }
+
+    fn max_cube_count() -> (u32, u32, u32) {
+        let max_thread = std::thread::available_parallelism()
+            .map(|n| n.get() as u32)
+            .unwrap_or(1);
+        (max_thread, 1, 1)
+    }
```
nathanielsimard, line 108: Since this dimension can be sequential, there is no meaninful max.

flake.nix
```diff
@@ -0,0 +1,36 @@
+{
```
nathanielsimard, line 1: I would remove the flake files.

crates/cubecl-cpu/LICENSE-APACHE
marcantoinem: Done

nathanielsimard (approved):

PR #701, opened 2025-05-27
Profiling with tracy

ArthurBrussee:

### Instructions


crates/cubecl-hip/README.md
```diff
@@ -14,6 +14,6 @@ with your use cases.
 
 ## RDNA 2
 
-RDNA 2 should work with cubecl-hip, however due RDNA2 not officially being supported by AMD, errors might be encountered. In order for an RDNA 2 GPU to work, the environment variable `HSA_OVERRIDE_GFX_VERSION` must be set to `10.3.0`.
+RDNA 2 should work with cubecl-hip, however due RDNA2 not officially being supported by AMD, errors might be encountered. In order for an RDNA 2 GPU to work, the environment variable `HAS_OVERRIDE_GFX_VERSION` must be set to `10.3.0`.
```
nathanielsimard, line 17: Not sure we should rename the env

crates/cubecl-runtime/src/client.rs
```diff
@@ -65,13 +74,31 @@ where
...
             #[cfg(multi_threading)]
-            spin::RwLock::new(None),
-        );
+            current_profiling: spin::RwLock::new(None),
+            // Create the GPU client if needed.
+            #[cfg(feature = "profile-tracy")]
+            gpu_client: client
+                .clone()
+                .new_gpu_context(
+                    Some("wgpu"),
```
nathanielsimard, line 92: We could use the server name here?

crates/cubecl-runtime/src/client.rs
```diff
@@ -349,17 +383,71 @@ where
...
+    #[track_caller]
+    pub fn profile(
+        &self,
+        func: impl FnOnce(),
+        #[allow(unused)] func_name: &str,
+    ) -> ProfileDuration {
+        // Get the outer caller. Maybe this should be lazy, idk.
+        // For execute() this could also maybe point straight to the
+        // cube kernel instead. For general profiling that's not possible.
+        #[cfg(feature = "profile-tracy")]
```
nathanielsimard, line 395: You can profile multiple kernels at the same time, so pointing to a kernel location isn't really possible or desirable.

crates/cubecl-runtime/src/client.rs
```diff
@@ -349,17 +383,71 @@ where
...
+                    method,
+                );
+            }
 
-        let result = self.channel.end_profile(token);
+            result
+        } else {
+            func();
+            self.channel.end_profile(token)
+        };
```
nathanielsimard, line 450: Not sure what's the difference between the two blocks. When tracy is activated, it has to create a span and collect the result, but the `func` and `end_profile` are the same. Maybe put a single `cfg(profile-tracy)` above the block that add the span, and log a warning is device timing method isn't in used? Also is it necessary to have device profiling for tracy to work?

crates/cubecl-wgpu/src/compute/stream.rs
```diff
@@ -241,95 +252,58 @@ impl WgpuStream {
...
+            Timings::System(..) => {
+                // Sync before profiling as well to get a cleaner measurement, we don't want to
+                // include any queued up work so far.
+                future::block_on(self.sync());
+                self.system_profiler().start()
+            }
+            Timings::Device(query) => {
+                // Close the current compute pass so that we start a new one. This keeps
+                // the timestamps separated.
+                self.compute_pass = None;
```
nathanielsimard, line 274: Do we need to flush here?

crates/cubecl-wgpu/src/compute/timings.rs
```diff
@@ -35,82 +36,216 @@ struct QuerySetItem {
...
+    device.create_buffer(&wgpu::BufferDescriptor {
+        label: Some("CubeCL gpu -> cpu map buffer"),
+        size: (QUERY_SIZE * count) as u64,
+        usage: wgpu::BufferUsages::MAP_READ | wgpu::BufferUsages::COPY_DST,
+        mapped_at_creation: false,
+    })
+}
+
+// Measure a timestamp to align the CPU & GPU timelines.
+fn synchronize_timestamps(queue: &wgpu::Queue, device: &wgpu::Device) -> (Instant, u64) {
```
nathanielsimard, line 66: Is it only useful when using tracy? If yes I would put that behind a feature flag

crates/cubecl-wgpu/src/compute/timings.rs
```diff
@@ -35,82 +36,216 @@ struct QuerySetItem {
...
         let mut timestamps = self.timestamps.remove(&token).unwrap();
+        let Timestamp { start, end } = &mut timestamps;
+        *end = self.current;
+
+        // TODO: We could optimize this by having a single handle for both `start` and `end`
+        // when a single query_set is used, but it probably doesn't impact the real
+        // performance all that much.
+        let (Some(start), Some(end)) = (start, end) else {
+            return None;
+        };
```
nathanielsimard, line 175: Having a single handle broke the alignment, that's why I used two handles, since storage takes care of the aligment problem.

crates/cubecl-wgpu/src/compute/timings.rs
ArthurBrussee: Sounds good, also simplfies the wasm stuff as it's more just that th feature flag doesn't work on wasm

crates/cubecl-wgpu/src/compute/timings.rs
ArthurBrussee: I should've mentioned but I moved the readback / buffer management into the query profiler as it made things have a bit less overhead. There's still two handles - but only one "mapped" buffer that contains the two timestamps.

ArthurBrussee:
Thanks! I"ve rebased things to head now

ArthurBrussee:
I've updated this again to head. Handling some autotuning error slightly different than at head because just returning a MAX duration isn't trivial anymore (need a start/end). I think this might fix some weird median times being reported though. Lmk how this looks

onurcanbektas:
Is there any estimate on when this will be merged into cubecl? It would greatly help with my current work.
I tried pulling the pull request locally but somehow there are some version conflicts etc. - couldn't make it work.

PR #708, opened 2025-06-02
Refactor cubecl-linalg into cubecl-matmul and cubecl-convolution

louisfd:
Also a part went into cubecl-std

### Instructions


crates/cubecl-cuda/src/lib.rs
```diff
@@ -25,14 +25,13 @@ mod tests {
...
     cubecl_std::testgen!();
 
-    cubecl_linalg::testgen_matmul_plane_accelerated!();
-    cubecl_linalg::testgen_matmul_unit!();
-    cubecl_linalg::testgen_matmul_tma!();
-    cubecl_linalg::testgen_matmul_quantized!();
-    cubecl_linalg::testgen_matmul_simple!([f16, bf16, f32]);
-    cubecl_linalg::testgen_matmul_tiling2d!([f16, bf16, f32]);
-    cubecl_linalg::testgen_conv2d_accelerated!([f16: f16, bf16: bf16, f32: tf32]);
-    cubecl_linalg::testgen_tensor_identity!([f16, bf16, f32, u32]);
```
laggui, line 35: Accidentally removed this test in the process?

```rust
cubecl_std::testgen_tensor_identity!([f16, bf16, f32, u32]);
```

crates/cubecl-std/src/tensor/mod.rs
```diff
@@ -1,4 +1,14 @@
...
-pub use base::*;
+pub use contiguous::*;
+pub use handle::*;
+pub use identity::*;
+pub use matrix_batch_layout::*;
 pub mod r#virtual;
+
+// /// Tests for tensor kernels
+// #[cfg(feature = "export_tests")]
+// pub mod tests;
```
laggui: Dead code

crates/cubecl-convolution/src/tests/test_macros/mod.rs
```diff
@@ -21,6 +21,7 @@ macro_rules! testgen_conv2d_accelerated {
     ($eg:ty, $es:ty) => {
         type Precision = ($eg, $es);
 
+        #[cfg(feature="conv_tests")]
```
laggui, line 24: What's the reason behind feature gating this test? Maybe this is intended, just not clear at first glance why

crates/cubecl-cuda/src/lib.rs
louisfd: Good catch!

crates/cubecl-convolution/src/tests/test_macros/mod.rs
louisfd: It's because compiling macro suites has become expensive and can make the LSP throttle. It was especially true for matmul but then it was weird to not have the matmul tests by default but have the convolution ones.

PR #727, opened 2025-06-19
Cubecl farm

Use-AIrs:
Every Farm is composed out of FarmGroups. In a FarmGroup we find a vector of FarmUnit. When a unit (a single GPU) is linked through something like a NcclComm, it has the state Linked. Otherwise, it's seen as Single. With the given implementations, we should be able to get a good frame to iterate over our Farm.

For now, I will align the FarmCube trait along with NCCL. The grouping of NCCL here is moved to Rust. This should make the concept more usable for Rust devs. The goal here will be to first just start each device in its own thread, to then use a thread per group to manage all concurrent devices. With some helper functions, we should be able to send kernels to all devices and so on. Later, there will be CPU thread strategies implemented. Also, posting to the CPUs memory while having training loops could be implemented further down the road.


Use-AIrs:
I will continue this weekend. Thought about implementing it trough FarmClient and FarmChannel , to maintain consistency for DX. But these will not implement the Compute traits. Instead, this will be a new trait structure. FarmClient will implement functions like `fn create_copied(group: usize, data: &[u8])` and `fn create_individual(group: usize, data: Vec<&[u8])`  and so on... FarmChannel will here to organize CPU threads, for now i make the assumption that cores/threads is min. `gpu_count + group_count`. Any concerns or input ?

Use-AIrs:
Ok, just pushed a basic starting point... FarmClient comes next. 
```rust
#[derive(Clone, Debug)]
pub struct FarmClient<FR: FarmRuntime, FC: FarmChannel<FR>> {
    farm: Arc<Farm<FR>>,
    channel: FC,
    _phantom: PhantomData<FR>,
}
```
Will look like that and implement a interface for the dev. 
This should also make it easy to implement over features.

Use-AIrs:
This is a bit overkill in threads spawned and also leads to slower use of the Farms since we coordinate every gpu trough its own thread, which loosed time when trying sync. A CPU Strategie will be added later on to fix that issue. First i will make a Client, which allows to easily operate a cluster trough dx interface. Feel free to give some input

Use-AIrs:
Directly building FarmUnits isn't possible as i did it. Will go with a FarmConfig build trough FarmRuntime and then build the Links in a group thread trough FarmChannel, that should work.

crates/cubecl-farm/LICENSE-APACHE
```diff
@@ -0,0 +1,201 @@
+                              Apache License
```
nathanielsimard, line 1: Normally we use symlink for the licenses.

Use-AIrs: Just copied one of yours ... Sorry, wasn't sure what to do ... So i just added something.

Use-AIrs:
> It's pretty early and I'm not sure if cubecl is the right level of abstraction for that functionality.

Thats the question after all, i need it, and for my app it feels like the right place. Wanted to propose it for this abstraction layer because... 

> There's probably changes that need to be made in cubecl-cuda and other runtimes for cubecl-farm to work 

... when letting this deeper into your implementations i would argue that one could make that pretty efficient and could match the cubecl abstraction layer. 

> but we could merge those smaller changes which should benefit other use cases (like improved to_device efficiency in burn).

The reason i sended in a draft which shows what my plan is, where these small changes after all and because i just wanted to offer that implementation since i see it more in the cubecl abstraction, but fair enough ...

Use-AIrs:
> I recommand continuing the effort, maybe merge it into your own repo (it's a new crate after all)

... thats the fallback strat after all. Will build that one way or the other :)

Use-AIrs:
After digging a bit i would agree, that managing the hole grouping and so on in cubecl dosn't make sense if you guys don't need it. But nccl features and the amd equivalent should be useable, right? In case of nccl this would mean, that a `Option<cudarc::sys::ncclComm_t>` would be a helpful thing in CudaContext or CudaStorage (easier to implement). Then `Farm::create_link()`  would be easy to implement ... Would you allow or not?

Use-AIrs:
or 
```rust 
enum state {
Stream {Stream}
Comm {Com}
}
```
would also work. A bit more efficient, but more restructuring would be needed

Use-AIrs:
> or
> 
> ```rust
> enum state {
> Stream {Stream}
> Comm {Com}
> }
> ```
> 
> would also work. A bit more efficient, but more restructuring would be needed

sorry wouldn't work... comm has CUStream in itself in cudarc but not in cuda, so would be slower i guess :D

nathanielsimard:
We have a branch in Burn where we're implementing `burn-collective`, a project that uses `to_device` for intra-node communication and networking for multi-node setups. In `cubecl`, we need to reintegrate `to_device` into the Runtime trait to leverage things like NVLink.

Since the `CudaRuntime` choses its own device type, maybe having something like `CudaDevice::Host{ id: usize }` and `CudaDevice::Remote{ host: String, id: usize }` could be possible if it's exposed by `nccl`.

Use-AIrs:
Will look into this, thanks.

Use-AIrs:
Got a idea now. Will do that on a new branch, since the Farm mechanic is now moved my lib now. Will do something like you proposed...

PR #733, opened 2025-06-19
Feat/documentation

marcantoinem:
It add a section with an example in the CubeCL books.


crates/cubecl-common/src/stub.rs
```diff
@@ -31,7 +31,7 @@ impl<T> Mutex<T> {
 
     /// Locks the mutex blocking the current thread until it is able to do so.
     #[inline(always)]
-    pub fn lock(&self) -> Result<MutexGuard<T>, alloc::string::String> {
+    pub fn lock(&self) -> Result<MutexGuard<'_, T>, alloc::string::String> {
```
nathanielsimard: Why do we need a lifetime here?

cubecl-book/src/getting-started/benchmark.md
```diff
@@ -0,0 +1,246 @@
...
+    client: ComputeClient<R::Server, R::Channel>,
+    _phantom_data: PhantomData<F>,
+}
+```
+## Implementing the benchmark trait
+To benchmark a CubeCL kernel, it is recommended to implement the `Benchmark` trait that defines the necessary methods for preparing, executing, and synchronizing the benchmark, because GPU are asynchronous and most benchmarking tool will not wait for the GPU to finish executing the kernel before measuring the time it took to execute it with a sync.
+```rust,ignore
+pub trait Benchmark {
+    /// Benchmark arguments.
+    type Args: Clone;
```
nathanielsimard: On main it is now `Input` and `Output`. I would update that code for use the latest version.

cubecl-book/src/getting-started/installation.md
```diff
@@ -1,19 +1,22 @@
...
 
-For `cuda`, simply install the CUDA drivers on your device. You can follow the installation
-instructions provided on the [NVIDIA website](https://developer.nvidia.com/cuda-downloads).
+| Platform | Runtime  | Supported OS              | Requirements                                             | Installation/Notes                                                                                                                                   | Feature Flag              |
+|----------|----------|---------------------------|----------------------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------|---------------------------|
+| WebGPU   | wgpu     | Linux, Windows, macOS     | Vulkan drivers (typically pre-installed on modern OSes)  | On WSL, manually install Vulkan drivers if missing. Refer to Microsoft’s WSL documentation. macOS uses Metal via WGPU; no additional drivers needed. | wgpu                      |
```
nathanielsimard: Vulkan only for Windows & Linux, Nothing for Mac and wasm

nathanielsimard (commented):
It would be great to add the vectorization benchmarks after its introduction in the book.

I would also add a third dimension to the problem to show how to leverage `CubeCount` when `CubeDim` is saturated by the number of threads/units in the GPU.

cubecl-book/src/getting-started/simple_reduction.md
```diff
@@ -0,0 +1,50 @@
+# Simple Reduction
+
+To get started with CubeCL, we will implement a simple reduction operation on a multidimensional array (tensor). This example will help you understand the basic concepts of CubeCL and how to use it to perform parallel computations on tensors.
+
+## An example of a CPU reduction in Rust without CubeCL
+This example demonstrates how to perform a simple reduction operation on a multidimensional array (tensor) using pure Rust. The code is designed to be easy to understand and serves as a starting point for more complex operations that can be parallelized with CubeCL. It is not optimized for performance, but it illustrates the basic concepts of working with tensors and performing reductions.
+
+### CpuTensor Struct
+Tensor are the basic data structure used in CubeCL to represent multidimensional arrays. Here is a simple implementation of a tensor in pure Rust. It is not optimized for performance, but it is easy to understand.
```
louisfd: Tensor are -> Tensors are

cubecl-book/src/getting-started/simple_reduction.md
```diff
@@ -0,0 +1,50 @@
...
+{{#rustdoc_include code_example/bin/v1-cpu.rs:17:31}}
+```
+
+## A first example of a GPU reduction with CubeCL
+This example demonstrates how to perform a simple reduction operation on a multidimensional array (tensor) using CubeCL. It is a simple implementation that will be used as a starting point to show how to use CubeCL in the next chapters.
+
+## GpuTensor struct
+The `GpuTensor` struct is a representation of a tensor that resides on the GPU. It contains the data handle, shape, strides, and marker types for the runtime and floating-point type. The `GpuTensor` struct provides methods to create tensors, read data from the GPU, and convert them into tensor arguments for kernel execution. Please note that it is generic over the runtime and floating-point type, allowing it to work with different CubeCL runtimes and floating-point types (e.g., `f16`, `f32`). Also, the strides can be computed using the `compact_strides` function from the `cubecl::std::tensor` module, which will compute the strides for a given shape with a compact representation.
+
+<div class="warning">
```
louisfd, line 38: This text doesn't appear as any different for me. Perhaps use the `>` markdown notation?

cubecl-book/src/getting-started/simple_reduction.md
```diff
@@ -0,0 +1,50 @@
...
+If you need a tensor library instead of defining your own kernel and tensor, you should use Burn directly instead.
```
louisfd: Burn should be clickable and lead to its github

cubecl-book/src/getting-started/code_example/gpu_tensor.rs
```diff
@@ -0,0 +1,76 @@
...
+            TensorArg::from_raw_parts::<F>(
+                &self.data,
+                &self.strides,
+                &self.shape,
+                vectorization_factor,
+            )
+        }
+    }
+
+    /// Return the data from the client
```
louisfd: Client/server concepts are introduced without explanation

cubecl-book/src/getting-started/code_example/gpu_tensor.rs
```diff
@@ -0,0 +1,76 @@
...
+    shape: Vec<usize>,
+    strides: Vec<usize>,
+    _r: PhantomData<R>,
+    _f: PhantomData<F>,
+}
+
+impl<R: Runtime, F: Float + CubeElement> Clone for GpuTensor<R, F> {
+    fn clone(&self) -> Self {
+        Self {
+            data: self.data.clone(),
```
louisfd: Maybe a comment stating it essentially only clones a pointer.

cubecl-book/src/getting-started/simple_reduction.md
```diff
@@ -0,0 +1,50 @@
...
+</div>
+
+```rust,ignore
+{{#include code_example/gpu_tensor.rs}}
+```
+
+## Reduce function
+Compared to the previous example, this function is similar but uses CubeCL's `cube` macro to define the kernel. The kernel performs the same reduction operation, summing the values of each row and storing the result in a new tensor. The variable `F` is a generic type that implements the `Float` trait, allowing the function to work with different floating-point types (e.g., `f32`, `f64`). The tensor is provided by cubecl::prelude, which includes the necessary traits and types for using CubeCL.
```
louisfd, line 47: A transition is missing, it's unclear what is "this" function. We should add a sentence saying we will port the previous function to CubeCL but keeping its sequential nature for now.

cubecl-book/src/getting-started/benchmark.md
```diff
@@ -0,0 +1,81 @@
+# Benchmarking reduction
+Now that we have a basic understanding of how to perform a reduction operation, let's benchmark it to see how it performs in terms of speed and efficiency.
+
+## Benchmarking struct
+For benchmarking, we will create a struct that holds the necessary information for the benchmark, such as the input shape, device, and client. This struct will be used to run the benchmark tests and configure the benchmarking environment. Please note that the `Runtime` and `Float` traits are used to make the benchmark generic over different CubeCL runtimes and floating-point types. A `PhantomData` is used to indicate that the struct holds a type parameter `F` without actually storing a value of that type, which is useful for generic programming in Rust for more information see the [Rust documentation](https://doc.rust-lang.org/std/marker/struct.PhantomData.html) and in our case allows us to change easily the type of float used in the benchmark.
```
louisfd: for more information see the Rust documentation should be inside parenthesis or something

louisfd: change easily -> easily change

cubecl-book/src/getting-started/benchmark.md
```diff
@@ -0,0 +1,81 @@
...
+    /// defined in the crate Cargo.toml
+    fn name(&self) -> String;
+
+    /// Wait for computation to complete.
+    fn sync(&self);
+}
+
+```
+
+In the `prepare` method, we will create the input data and return a handle as a Args that will be used in the `execute` method. The `execute` method will launch the kernel and the sync method will wait for the GPU to finish executing the kernel before measuring the time it took to execute it. Don't forget to add the function that we want to benchmark.
```
louisfd: as a Args?

cubecl-book/src/getting-started/benchmark.md
```diff
@@ -0,0 +1,81 @@
...
+  Timing      system
+  Samples     10
+  Mean        241.018ms
+  Variance    1.068µs
+  Median      240.943ms
+  Min         239.734ms
+  Max         243.782ms
+―――――――――――――――――――――――――
+```
+Somehow our time is not that good, but it is expected because we are using a very simple kernel that does not take advantage of the GPU parallelism. In the next chapter, we will see how to optimize our kernel to take advantage of the GPU parallelism and improve the performance of our reduction operation.
```
louisfd: It feels odd to say the time isn’t great without any point of reference. We seem to expect the reader to realize that, but there’s no way for them to know. We should rephrase with something like "As we will see..."

cubecl-book/src/getting-started/parallel_reduction.md
```diff
@@ -55,3 +60,36 @@ often not present in other languages, except WebGPU with `local_invocation_index
 | ABSOLUTE_POS_Z | N/A         | global_id.z            |
 
 </details>
+
+## Parallel Reduction Example
```
louisfd, line 64: Note that we needs -> Note that we need

cubecl-book/src/getting-started/vectorized_reduction.md
```diff
@@ -0,0 +1,39 @@
+# Vectorized Reduction
+In this section, we will explore how to implement a vectorized reduction operation using CubeCL. Vectorization is a powerful technique that allows us to process multiple data elements simultaneously, significantly improving performance for certain types of computations.
```
louisfd: Especially I/O operations

cubecl-book/src/getting-started/vectorized_reduction.md
```diff
@@ -0,0 +1,39 @@
...
+
+## What is vectorization?
```
louisfd, line 4: In all of CubeCL the term line size is preferred over vectorization factor. We should introduce it and rename VECTORIZATION_FACTOR to LINE_SIZE.

louisfd (changes requested):

crates/cubecl-common/src/stub.rs
marcantoinem: This give a warning in nightly, but it is not a problem on stable, so I remove it for the time being.

cubecl-book/src/getting-started/benchmark.md
marcantoinem: done

nathanielsimard (approved):

PR #740, opened 2025-06-22
Speedup spirv compilation and make matmul autotune key softer.

nathanielsimard:


crates/cubecl-matmul/src/tune_key.rs
```diff
@@ -110,10 +85,16 @@ impl MatmulAutotuneKey {
         let matrix_layout_lhs = matrix_batch_layout(lhs_strides);
         let matrix_layout_rhs = matrix_batch_layout(rhs_strides);
 
+        let bigguest_axis = usize::max(m, n);
```
louisfd: biggest

PR #755, opened 2025-06-29
CudaDevice new and small ctx addition

Use-AIrs:
Here we got `CudaDevice` as enum and added an `Option<ncclComm_t>` to the Context. Would add functions in the ctx implementation so it's easy to initialize a comm if you guys allow these changes in `cubecl-cuda`.

### DIFF
--- a/crates/cubecl-cuda/Cargo.toml
+++ b/crates/cubecl-cuda/Cargo.toml
@@ -13,12 +13,12 @@ version.workspace = true
 [features]
 compilation-cache = ["cubecl-common/cache"]
 default = [
-    "std",
-    "cubecl-runtime/default",
-    "cubecl-common/default",
-    "cubecl-core/default",
-    "cudarc/cuda-12050",
-    "cudarc/dynamic-loading",
+  "std",
+  "cubecl-runtime/default",
+  "cubecl-common/default",
+  "cubecl-core/default",
+  "cudarc/cuda-12050",
+  "cudarc/dynamic-loading",
 ]
 std = ["cubecl-runtime/std", "cubecl-common/std", "cubecl-core/std"]
 ptx-wmma = []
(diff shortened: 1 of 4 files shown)


Use-AIrs:
Would you allow that if i build the needed functionalities, would build it different if thats a no for you...

crates/cubecl-cuda/Cargo.toml
```diff
@@ -40,55 +40,56 @@ matmul_tests_f32 = ["cubecl-matmul/matmul_tests_f32"]
...
 cudarc = { version = "0.16.3", features = [
-    "std",
-    "driver",
-    "cuda-version-from-build-system",
-    "dynamic-loading",
+  "std",
+  "driver",
+  "cuda-version-from-build-system",
+  "dynamic-loading",
+  "nccl",
```
nathanielsimard, line 92: I would put that behind a feature flag,

```toml
[features]
nccl = ["cudarc/nccl"]
```

crates/cubecl-cuda/src/compute/server.rs
```diff
@@ -575,6 +576,40 @@ impl CudaContext {
...
+    }    
+
+    pub fn new_with_comm(
+        memory_management: MemoryManagement<CudaStorage>,
+        compilation_options: CompilationOptions,
+        stream: cudarc::driver::sys::CUstream,
+        comm: cudarc::nccl::sys::ncclComm_t,
+        context: *mut CUctx_st,
+        arch: CudaArchitecture,
+    ) -> Self {
```
nathanielsimard, line 593: We don't need two constructors, instead we can add the field `comm` when the feature flag `nccl` is activated.

crates/cubecl-cuda/src/device.rs
```diff
@@ -4,13 +4,61 @@
...
+    },
+    Linked {
+        id: usize,
+        group: usize,
+    },
+    LinkedRemote {
+        id: usize,
+        group: usize,
+        host: String,
+    },
```
nathanielsimard, line 24: I would keep the same struct as before and keep the `index` field (to be compatible with previous API).

I would add a new field `kind` that contains that enum (behind the feature flag `nccl`). Since all variants of that enum have `id` it's the same as the current `index`, we can remove that in each variant of the updated enum `CudaDeviceKind`.

Use-AIrs:
Ok, thanks allot! Will update this weekend.

PR #756, opened 2025-06-30
Fix/plane size

nathanielsimard:

### Instructions

### DIFF
--- a/crates/cubecl-core/src/runtime_tests/cmma.rs
+++ b/crates/cubecl-core/src/runtime_tests/cmma.rs
@@ -767,21 +767,19 @@ pub fn test_cmma_strided<R: Runtime>(
 macro_rules! testgen_cmma {
     () => {
         use super::*;
-        use cubecl_core::CubeDim;
+        use cubecl_core::prelude::*;
 
         #[test]
         fn test_cmma_simple_1() {
             let client = TestRuntime::client(&Default::default());
-            // In HIP the thread block size must be 32
-            let cube_dimensions = CubeDim::new(32, 2, 1);
+            let cube_dimensions = cube_dim::<TestRuntime>(&client);
             cubecl_core::runtime_tests::cmma::test_simple_1::<TestRuntime>(client, cube_dimensions);
         }
 
         #[test]
         fn test_cmma_simple_1_lined() {
             let client = TestRuntime::client(&Default::default());
-            // In HIP the thread block size must be 32
-            let cube_dimensions = CubeDim::new(32, 2, 1);
+            let cube_dimensions = cube_dim::<TestRuntime>(&client);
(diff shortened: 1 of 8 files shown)


ArthurBrussee:
This is not legit - it's a variable precisely because there are no guarantees GPU side that this is a constant. It can vary _per workgroup_ depending on your architecture.

On Desktop nvidia and AMD it likely doesn't, but on Intel it very much does, let alone mobile GPU's and co. It'll also break on WASM as the readback needs to be async.

nathanielsimard:
> This is not legit - it's a variable precisely because there are no guarantees GPU side that this is a constant. It can vary _per workgroup_ depending on your architecture.
> 
> On Desktop nvidia and AMD it likely doesn't, but on Intel it very much does, let alone mobile GPU's and co. It'll also break on WASM as the readback needs to be async.

Well we need to know the plane_size to create `CubeDim(plane_size, .., ..)` because otherwise we have an error with AMD devices on vulkan. Is there another way to be sure what's the plane_size at runtime? We can always ignore that on wasm (blocking not enabled)

ArthurBrussee:
That's what I'm saying - "plane_size at runtime" doesn't make sense, it doesn't have a fixed value as on some architectures as it can vary per shader execution / or even per SM. If it can't vary WGPU should already be able to return a value so that min == max size.

The usual workaround is to do things based on worst-case assumptions and handle the rest in the shader runtime. You can maybe hardcode it for HIP / Cuda as that makes some more guarantees, but definitely can't on the WGPU level which makes no such guarantees. 

Vulkan makes no guarantees either, though modern Vulkan (1.3+) has a way of setting the subgroup size, so maybe worth looking into that.

nathanielsimard:
> That's what I'm saying - "plane_size at runtime" doesn't make sense, it doesn't have a fixed value as on some architectures as it can vary per shader execution / or even per SM. If it can't vary WGPU should already be able to return a value so that min == max size.

I don't fully agree, as the SPIR-V validation layer on my AMD laptop returns an error if the first cube dimension (CUBE_DIM_X) isn't a factor of the subgroup size. This requires knowing the subgroup size before compiling a compute kernel. While the Vulkan API can support multiple subgroup sizes at runtime, we still need to verify this, as otherwise, we cannot correctly code or launch kernels that use subgroup operations like `cmma`. We could reflect this fact by having a function, `defined_subgroup_size(client, cube_dim_x)`, since we can assume that the `subgroup_size` will be consistent for all kernels with a specified `cube_dim`.

For now, I think we can test the following:

* Get the subgroup size with `CubeDim(subgroup_min, 1, 1)`
* Get the subgroup size with `CubeDim(subgroup_max, 1, 1)`

If both values are the same, we set it as the default `subgroup_size`. We could limit this algorithm to the Vulkan backend and disable the subgroup feature otherwise, since none of our kernels that use subgroup instructions can run with an unknown subgroup size.

PR #764, opened 2025-07-04
Fix/spirv various

marcantoinem:
This PR contains fix to avoid segfault in burn notably:
- Merge same expression could merge expression that doesn't share the same out type.
- get_or_insert_global would not generate the necessary instruction when it is already cached. Removing the repetition of constant should be done in cubecl-opt not in spir-v.

It also introduce a printer for CubeCL IR by implementing Display for Scope.

### Instructions

- [X] Create a new branch or fork of the [burn repo](https://github.com/Tracel-AI/burn)
- [X] Update the main [Cargo.toml](https://github.com/tracel-ai/burn/blob/40aa993fb5969351a006b560ec89da5119dc9721/Cargo.toml#L159=L161) with this PR hash.
- [X] Fix any broken tests or compilation errors in burn.
- [X] Submit a PR in burn with your fixes and link it here.

### DIFF
--- a/crates/cubecl-spirv/src/lookups.rs
+++ b/crates/cubecl-spirv/src/lookups.rs
@@ -164,13 +164,9 @@ impl<T: SpirvTarget> SpirvCompiler<T> {
         item: Item,
         insert: impl FnOnce(&mut Self) -> Word,
     ) -> Word {
-        if let Some(id) = self.state.constants.get(&(value, item.clone())) {
-            *id
-        } else {
-            let id = insert(self);
-            self.state.constants.insert((value, item), id);
-            id
-        }
+        let id = insert(self);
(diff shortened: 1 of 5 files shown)


crates/cubecl-spirv/src/lookups.rs
```diff
@@ -164,31 +164,23 @@ impl<T: SpirvTarget> SpirvCompiler<T> {
...
-            let id = insert(self);
-            self.state.constants.insert((value, item), id);
-            id
-        }
+        let id = insert(self);
+        self.state.constants.insert((value, item), id);
+        id
     }
 
     pub fn get_or_insert_global(
```
nathanielsimard, line 172: Is the same of the function still true? It doesn't seem like a get anymore.

PR #766, opened 2025-07-04
Feat/nccl

Use-AIrs:
So started with a NcclDevice behind a nccl feature flag. With some CudaDevice implementations which come with the feature to make it easier to set up groups.

 CudaDevice is needed to be passed trough client so it comes trough .info() now with the feature on.

Next comes NcclOp with NcclOpBuilder. NcclOpBuilder will first translate the input Handel(s) to get ptr und size. Then it can be executed, which will build a `cudarc::nccl::sys::ncclComm_t`, then starts the nccl group executes Futures on pointers step by step to then stop the nccl group. At least thats the plan for now.


Use-AIrs:
No Future here, that would be a overcomplicated abstraction ^^

crates/cubecl-cuda/src/compute/nccl.rs
```diff
@@ -0,0 +1,194 @@
...
+                    nccl: Some(nccl),
+                };
+                devices.push(cuda);
+            }
+        }
+
+        devices
+    }
+
+    pub fn linked_groups(split: Vec<f64>) -> Vec<Self> {
```
nathanielsimard: What is a split? Maybe add some docs in each public functions to help users

Use-AIrs: In this Context you get a f64 for each device group you want do build. Each group gets devices assigned depending on the proportion of each f64 compared to the sum of Vec<f64>...

Just asking early on that frequently, because i want to get a felling if getting deeper in the code gets backslash, sorry for not pointing that out.
Since you did not comment the use of using `CudaRuntime::Info` i assume that using existing code is allowed as long as its behind the feature flag ...

Use-AIrs:
Will add tests soon...

Use-AIrs:
When testing this, the gpus mem runs full. don't know why jet, if someone got an idea ... Will look into that

PR #784, opened 2025-07-19
Fix typo in example code of cubecl-book

AtomsForHire:
Hello,

I believe this line should be using the `output` variable instead of `input`. 

This is my first pull request and I have little to no experience with git in a collaborative environment, so if I should have contacted a developer instead, I apologise. I'm also a bit unsure about this step (since it is a typo fix):

- [] Update the main [Cargo.toml](https://github.com/tracel-ai/burn/blob/40aa993fb5969351a006b560ec89da5119dc9721/Cargo.toml#L159=L161) with this PR hash.

A bit of guidance would be greatly appreciated :slightly_smiling_face:


nathanielsimard (approved):

nathanielsimard:
For a typo, the steps are not required

PR #786, opened 2025-07-20
Added gpu_count to Runtime

Use-AIrs:
Just added and implemented `fn gpu_count() -> usize;` to `Runtime` and implemented it for all Runtimes. 
Wasn't able to test the hip implementation, should be tested before merging.


nathanielsimard:
I would rename the function `gpu_count` to `device_count` since some runtimes will not be using the GPU (CPU, TPU, etc.)

Use-AIrs:
Sry, still got the wrong gpg, will fix that.

nathanielsimard (approved):

PR #792, opened 2025-07-27
feat: Add comprehensive NCCL operations support for CubeCL CUDA runtime

Use-AIrs:
Ok, so I made the first NCCL implementation which works now. The design isn't for bigger clusters since it's a single-threaded async approach. For small device counts, this works but not really for HPC - I'll be able to make this more interesting for HPC in a few weeks... But this is a usable approach without grouping, just a simple straightforward approach for now.

Tests work, have a short doc entry with that, and no more memory leaks :)

### DIFF
--- a/crates/cubecl-cuda/src/compute/nccl.rs
+++ b/crates/cubecl-cuda/src/compute/nccl.rs
@@ -1,5 +1,5 @@
 #![allow(unused)]
-
+//! **Disclaimer**: This is not safe! Most of it is, but you should definitely know or find out how many GPUs you got and don't forget to start counting at 0. With the wrong rank you give a wrong pointer. This is subject to change in the future. 
 //! NCCL Operations for CubeCL CUDA Runtime
 //!
 //! This module provides high-level NCCL collective operations like AllReduce,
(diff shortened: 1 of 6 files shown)


Use-AIrs:
So took a step back to the grouping abstraction as discussed. Hope thats more the direction :)

crates/cubecl-cuda/src/compute/nccl.rs
```diff
@@ -0,0 +1,584 @@
...
+                    .unwrap();
+                }
+            }
+        }
+    }
+
+    pub async fn barrier(&self) {
+        for (_, device) in &self.map {
+            device.client.sync().await;
+        }
```
nathanielsimard, line 313: We should await in parallel instead of sequential using `join_all` or something equivalent.

crates/cubecl-cuda/src/compute/nccl.rs
```diff
@@ -0,0 +1,584 @@
...
+
+    #[test]
+    fn test_all_gather_single_device() {
+        async {
+            let nccl_op = NcclOp::<f32>::init::<CudaRuntime>().init_all().await;
+            let send_data = vec![1.0, 2.0];
+            let expected_recv_size = send_data.len() * nccl_op.device_count;
+            let cu_device = CudaDevice { index: 0 };
+            let client = <CudaRuntime as Runtime>::client(&cu_device);
+            let send_handle = create_test_data(&client, send_data.clone());
```
nathanielsimard, line 449: It's weird to test the send/recv on the same device no?

crates/cubecl-cuda/src/compute/nccl.rs
Use-AIrs: Fair

crates/cubecl-cuda/src/compute/nccl.rs
Use-AIrs: Just got one gpu ^^ Getting my Cluster later this week. But had to test somehow.

PR #793, opened 2025-07-28
Feat/mlir add support shared memory

marcantoinem:
Add support for shared_memory and init_line.

There is still some problem remaining with sync_cube to make Burn work correctly.

### DIFF
--- a/crates/cubecl-cpu/src/compiler/builtin.rs
+++ b/crates/cubecl-cpu/src/compiler/builtin.rs
@@ -1,8 +1,10 @@
 use cubecl_core::{CubeDim, ir::Builtin};
 
+const NB_PASSED_BUILTIN: usize = 9;
+
 #[derive(Default, Debug, Clone)]
 pub struct BuiltinArray {
-    pub dims: [u32; 9],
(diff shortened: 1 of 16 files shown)


crates/cubecl-cpu/src/compiler/builtin.rs
```diff
@@ -45,7 +45,7 @@ impl BuiltinArray {
         self.dims[7] = unit_pos[1];
         self.dims[8] = unit_pos[2];
     }
-    pub(crate) const fn len(&self) -> usize {
-        self.dims.len()
+    pub(crate) const fn len() -> usize {
+        9
```
nathanielsimard: Is it always 9?

nathanielsimard (approved):

crates/cubecl-cpu/src/compiler/builtin.rs
marcantoinem: It's a magic number, I've fixed it. It is always the same number of builtins that is passed each times

nathanielsimard (approved):

PR #800, opened 2025-08-04
[Refactor] Memory management API

wingertge:
Reworks the server API for managing memory to make it simpler, and add a way to execute a memcpy after a tensor is created.
User-facing client APIs are kept almost the same, with minimal breakage. They should probably be reworked at some point, but I'd like to minimize the impact of the PR.

# Breaking changes
The only breaking changes are with `empty_tensors`, `create_tensors` and `read_tensors` (only the ones with multiple handles), as well as making `read` take a `Handle` instead of a `Binding` to make it consistent with the rest of the memory management API.

# Testing
Many tests fail, but the same test fail in the same ways on `main`. Same with `burn`.

### DIFF
--- a/crates/cubecl-cpu/src/compute/server.rs
+++ b/crates/cubecl-cpu/src/compute/server.rs
@@ -64,7 +64,7 @@ impl CpuServer {
         ) -> Result<Vec<Vec<u8>>, IoError> {
             let mut result = Vec::with_capacity(descriptors.len());
             for desc in descriptors {
-                let binding = desc.handle.binding();
+                let binding = desc.binding;
                 let resource = ctx
                     .memory_management
                     .get_resource(binding.memory, binding.offset_start, binding.offset_end)
@@ -134,7 +134,7 @@ impl ComputeServer for CpuServer {
                 return Err(IoError::UnsupportedStrides);
             }
 
-            self.copy_to_binding(desc.handle.binding(), data);
+            self.copy_to_binding(desc.binding, data);
         }
         Ok(())
     }
@@ -157,8 +157,7 @@ impl ComputeServer for CpuServer {
     ) {
         let cube_count = match count {
             CubeCount::Static(x, y, z) => [x, y, z],
-            CubeCount::Dynamic(handle) => {
-                let binding = handle.binding();
+            CubeCount::Dynamic(binding) => {
                 let handle = self
                     .ctx
                     .memory_management
--- a/crates/cubecl-wgpu/src/compute/stream.rs
+++ b/crates/cubecl-wgpu/src/compute/stream.rs
@@ -259,9 +259,9 @@ impl WgpuStream {
         })
     }
 
-    pub fn read_binding(&mut self, handle: Handle) -> DynFut<Result<Vec<u8>, IoError>> {
-        let shape = [handle.size() as usize];
-        let data = self.read_buffers(vec![CopyDescriptor::new(handle.binding(), &shape, &[1], 1)]);
+    pub fn read_binding(&mut self, binding: Binding) -> DynFut<Result<Vec<u8>, IoError>> {
+        let shape = [binding.size() as usize];
+        let data = self.read_buffers(vec![CopyDescriptor::new(binding, &shape, &[1], 1)]);
         Box::pin(async move {
             let data = data.await?.remove(0);
             Ok(data)
@@ -338,7 +338,7 @@ impl WgpuStream {
                 //
                 // For now, instead do a dummy readback. This *seems* to wait for the entire
                 // queue to be done.
-                let fut = self.read_binding(buf.clone());
+                let fut = self.read_binding(buf.clone().binding());
                 core::mem::swap(&mut buffer, &mut self.sync_buffer);
                 Box::pin(async move {
                     fut.await.expect("Failed to read sync buffer");


crates/cubecl-cpu/src/compute/server.rs
```diff
@@ -80,74 +89,62 @@ impl ComputeServer for CpuServer {
...
 
-    fn read(&mut self, bindings: Vec<Binding>) -> DynFut<Vec<Vec<u8>>> {
-        Box::pin(self.read_async(bindings))
-    }
+    fn create(
+        &mut self,
+        descriptors: Vec<AllocationDescriptor<'_>>,
+    ) -> Result<Vec<Allocation>, IoError> {
+        let align = 8;
+        let strides = descriptors
```
nathanielsimard, line 97: Are all new allocations have strides? Is it forced to be a tensor?

crates/cubecl-cuda/src/compute/server.rs
```diff
@@ -83,83 +89,76 @@ unsafe impl Send for CudaServer {}
...
+            let mut result = Vec::with_capacity(descriptors.len());
+
+            for descriptor in descriptors {
+                let CopyDescriptor {
+                    handle,
+                    shape,
+                    strides,
+                    elem_size,
+                } = descriptor;
+                let binding = handle.binding();
```
nathanielsimard: There was a reason to use bindings instead of handles! A binding doesn't carry the same reference count, so after an operation is built, even if it's not received by the server, the resource will be marked as available. We could do the handle-to-binding conversion on the client to have a consistent API, but I would try to avoid it on the server.

crates/cubecl-cuda/src/compute/server.rs
```diff
@@ -907,3 +904,28 @@ fn oob_to_cuda(fill: OobFill) -> CUtensorMapFloatOOBfill {
         OobFill::NaN => CU_TENSOR_MAP_FLOAT_OOB_FILL_NAN_REQUEST_ZERO_FMA,
     }
 }
+
+pub fn valid_strides(shape: &[usize], strides: &[usize]) -> bool {
```
nathanielsimard, line 906: If you find functions that could be reused between runtimes, we could put those in `cubecl-runtime`.

crates/cubecl-cuda/src/compute/storage.rs
```diff
@@ -126,12 +127,23 @@ impl ComputeStorage for CudaStorage {
...
-            unsafe { cudarc::driver::result::malloc_async(self.stream, size as usize).unwrap() };
+        let ptr = unsafe { cudarc::driver::result::malloc_async(self.stream, size as usize) };
+        let ptr = match ptr {
+            Ok(ptr) => ptr,
+            // I don't think this actually triggers immediately, might be returning the error on the next call
+            // Need to figure out how to handle this
+            Err(DriverError(cudarc::driver::sys::CUresult::CUDA_ERROR_OUT_OF_MEMORY)) => {
+                Err(IoError::BufferTooBig(size as usize))?
+            }
+            Err(other) => panic!("{other}"),
```
nathanielsimard, line 140: Since we're not doing a lot of allocations, mostly to fill the memory pages, we could sync the stream when the alloc size is getting close to the limit. The storage could keep two values (MAX TOTAL MEMORY and MAX BUFFER ALLOC) when one of those is close we could sync.

crates/cubecl-runtime/src/server.rs
```diff
@@ -138,6 +140,63 @@ pub struct Handle {
...
+    /// Optimized for access speed. In practice this means row-aligned with padding for runtimes
+    /// that support it.
+    Optimized,
+}
+
+/// Descriptor for a new tensor allocation
+#[derive(new, Debug, Clone, Copy)]
+pub struct AllocationDescriptor<'a> {
+    /// Layout for the tensor
+    pub type_: AllocationType,
```
nathanielsimard: Sometimes just to avoid the `type` naming clash, I use `Kind` and `kind` instead (in enums and fields)

crates/cubecl-runtime/src/server.rs
```diff
@@ -264,14 +323,14 @@ pub struct Binding {
...
 
 /// A binding with shape and stride info for non-contiguous reading
-#[derive(new, Debug)]
-pub struct BindingWithMeta {
-    /// Binding for the memory resource
-    pub binding: Binding,
+#[derive(new, Debug, Clone)]
+pub struct CopyDescriptor<'a> {
+    /// Handle for the memory resource
+    pub handle: Handle,
```
nathanielsimard: Maybe here it should be a binding?

crates/cubecl-cpu/src/compute/server.rs
wingertge: Technically yes, but a 1D allocation (by just giving size) would still just be unit stride with no padding. The buffers always needed to be aligned to the memory alignment in the first place, so the padding between buffers is the same.

crates/cubecl-cuda/src/compute/server.rs
wingertge: Gotcha. The main reason I changed it is because wgpu actually used `Handle` internally for it's read function and it seemed simpler since it has a length on it. Maybe `Binding` could just get that same field for consistency?

crates/cubecl-cuda/src/compute/server.rs
wingertge: It's just for CUDA/HIP and then wgpu separately rn since CUDA and HIP have mirrored APIs with 2D copy but wgpu doesn't have it. Didn't seem like it made sense to pull it out when it's essentially only used by one API (and its identical clone in HIP).

crates/cubecl-runtime/src/server.rs
wingertge: That's a better name anyways

crates/cubecl-runtime/src/server.rs
wingertge: Yeah as mentioned above, changing it back and giving `Binding` a size field so you don't always have to fetch the resource to get the size.

crates/cubecl-cuda/src/compute/server.rs
wingertge: I do think the binding conversion should happen in the `Client` though, I noticed how much cleaner the API is with `Handle` because we were always passing `handle.binding()` to that function in the implementations

crates/cubecl-cpu/src/compute/server.rs
nathanielsimard: Makes sense

crates/cubecl-cuda/src/compute/server.rs
nathanielsimard: Yeah and wgpu can't easily reuse buffers in the same command buffer, so it's not a major problem there.

crates/cubecl-cuda/src/compute/storage.rs
wingertge: I'd probably do that in a separate PR though. The infrastructure for error handling is at least there, and there are situations particularly in wgpu where `BufferTooBig` is actually thrown properly. Extending the error handling is easy once the API is in place.

PR #803, opened 2025-08-06
[Feat] Automatic unrolling for line sizes above max

wingertge:
It's a bit fragile so might need fixes as quantization kernels start using it. Currently doesn't change max line size based on type, but could easily be adapted if it's ever needed.

The only non-IR code that required changes is the checked indexing, because it needs to translate user-facing length to the internal unrolled length at lower vectorization.

# Testing
No regressions in burn/cubecl, adds two new tests for basic validation of the unroll.


crates/cubecl-core/src/post_processing/unroll.rs
```diff
@@ -0,0 +1,478 @@
...
+                .iter()
+                .any(|arg| arg.vectorization_factor() > self.max_line_size)
+        {
+            let line_size = args
+                .iter()
+                .map(|it| it.vectorization_factor())
+                .max()
+                .unwrap();
+            let line_size =
+                line_size.max(inst.out.map(|out| out.vectorization_factor()).unwrap_or(1));
```
nathanielsimard: I would extract that into a function I think

crates/cubecl-core/src/post_processing/unroll.rs
```diff
@@ -0,0 +1,478 @@
...
+                Operation::Operator(Operator::Index(op)) if op.list.is_array() => {
+                    let mut indices = vec![];
+
+                    for _ in 0..unroll_factor as usize {
+                        indices
+                            .push(*alloc.create_local(Item::new(u32::as_elem_native_unchecked())));
+                    }
+
+                    let mut list = op.list;
+                    list.item.vectorization = NonZero::new(self.max_line_size);
```
nathanielsimard: `max_line_size` is always right if all line sizes are factor of smaller line sizes. (8, 4, 2, 1) etc.

crates/cubecl-core/src/post_processing/unroll.rs
```diff
@@ -0,0 +1,478 @@
...
+
+                    let out = mappings.get(alloc, inst.out(), unroll_factor, self.max_line_size);
+                    let instructions = (0..unroll_factor as usize)
+                        .flat_map(|i| {
+                            let add_idx =
+                                add_index_inst(alloc, op.index, unroll_factor, i, indices[i]);
+                            let index = Instruction::new(
+                                Operator::Index(IndexOperator {
```
nathanielsimard: The index operation could be replaced by a `CopyMemoryBulk`?

crates/cubecl-core/src/post_processing/unroll.rs
```diff
@@ -0,0 +1,478 @@
...
+                }
+                TransformAction::Replace(replacement) => {
+                    new_instructions.extend(replacement);
+                }
+            }
+        }
+
+        new_instructions
+    }
+}
```
nathanielsimard, line 591: General comment for this transformation: it would be easier if most match branches were extracted into functions. Those functions could be commented with the strategy used to unroll them.

crates/cubecl-opt/src/lib.rs
```diff
@@ -175,15 +173,15 @@ impl Optimizer {
     pub fn new(
         expand: Scope,
         cube_dim: CubeDim,
-        mode: ExecutionMode,
-        transformers: Vec<Rc<dyn IrTransformer>>,
+        transformers: Vec<Mutex<Box<dyn IrTransformer>>>,
```
nathanielsimard: We could add a comment on why the mutex is necessary

crates/cubecl-spirv/src/lookups.rs
```diff
@@ -103,7 +103,10 @@ impl<T: SpirvTarget> SpirvCompiler<T> {
         self.state.buffers = kernel
             .buffers
             .into_iter()
-            .map(|binding| {
+            .map(|mut binding| {
+                if binding.item.vectorization() > MAX_VECTORIZATION {
+                    binding.item.vectorization = NonZero::new(MAX_VECTORIZATION);
+                }
```
nathanielsimard, line 111: Would it break correcness? we should add a comment on why this is correct.

crates/cubecl-wgpu/src/compiler/wgsl/compiler.rs
```diff
@@ -102,7 +109,12 @@ impl WgslCompiler {
             buffers: value
                 .buffers
                 .into_iter()
-                .map(|it| self.compile_binding(it))
+                .map(|mut it| {
+                    if it.item.vectorization() > MAX_LINE_SIZE {
+                        it.item.vectorization = NonZero::new(MAX_LINE_SIZE);
+                    }
```
nathanielsimard, line 117: Same

crates/cubecl-core/src/post_processing/unroll.rs
wingertge: We require line sizes to be power of two at this point, right? So yeah, `max_line_size` should always be a divisor of a line size that's `> max_line_size`.

crates/cubecl-core/src/post_processing/unroll.rs
wingertge: I think so, let me check and try to remember why I did it like this instead 😅

crates/cubecl-opt/src/lib.rs
wingertge: Ah that's a leftover, but I'd like to remove it entirely and replace it with `Processor` eventually. I'll look through the changes and remove it. I initially implemented the unroll as an `IrTransformer` and made it mutable so it can carry state.

crates/cubecl-spirv/src/lookups.rs
wingertge: It doesn't break correctness if anything above `MAX_VECTORIZATION` is also unrolled. But a comment is a good idea, to note that it must be paired with the processor.

crates/cubecl-core/src/post_processing/unroll.rs
wingertge: Done 👍

nathanielsimard (approved):

PR #810, opened 2025-08-10
Add MMA op with manually managed registers

wingertge:
Adds an API to execute MMA with manually managed registers, giving full control over memory layout and zero fill, along with increasing type support.

# Testing
Adds new test for the manual MMA instruction


crates/cubecl-core/src/frontend/cmma.rs
```diff
@@ -210,6 +258,145 @@ impl<C: CubePrimitive> Matrix<C> {
...
+                MatrixIdent::Accumulator => Elem::size_bits(&self.cd_elem) as u32,
+            };
+            let register_size = scope.runtime_properties.mma.register_size_bits;
+            // div_ceil for potential compatiblity with f64
+            register_size.div_ceil(bits)
+        })
+    }
+
+    /// Returns the coordinates of the `nth` element handled by the plane index `lane_id`
+    /// Each lane contains [`elems_per_lane`] elements in [`line_size`] chunks.
```
nathanielsimard, line 355: It would be cool in the MmaDefinition struct to have definitions of : `lane` (Is it the plane or the unit). From that I've seems previously, the line_size is determined by the number of bits in each lane, so we could document that as well.

crates/cubecl-core/src/frontend/cmma.rs
```diff
@@ -441,6 +628,80 @@ pub mod cast {
...
+
+    /// Expand method of [execute_manual()].
+    #[allow(unused_variables)]
+    pub fn expand<AB: CubePrimitive, CD: CubePrimitive>(
+        scope: &mut Scope,
+        matrix: MmaDefinitionExpand<AB, CD>,
+        a_registers: SequenceExpand<Line<AB>>,
+        b_registers: SequenceExpand<Line<AB>>,
+        c_registers: SequenceExpand<Line<CD>>,
+        d_registers: SequenceExpand<Line<CD>>,
```
nathanielsimard: Is there a reason to use Sequence instead of Array? Both works, just wondering how using sequence is easier (in terms of API).

crates/cubecl-core/src/runtime_tests/cmma.rs
```diff
@@ -762,6 +764,169 @@ pub fn test_cmma_strided<R: Runtime>(
...
+pub fn kernel_manual<C: CubePrimitive>(
+    a: &Tensor<C>,
+    b: &Tensor<C>,
+    c: &Tensor<f32>,
+    out: &mut Tensor<f32>,
+    #[comptime] size_m: u32,
+    #[comptime] size_n: u32,
+    #[comptime] size_k: u32,
+) {
+    let def = cmma::MmaDefinition::<C, f32>::new(size_m, size_n, size_k);
```
nathanielsimard: The struct definition is so great btw!

crates/cubecl-core/src/runtime_tests/cmma.rs
```diff
@@ -762,6 +764,169 @@ pub fn test_cmma_strided<R: Runtime>(
...
+    let lane_id = UNIT_POS_PLANE;
+
+    let a_elems = def.elems_per_lane(MatrixIdent::A);
+    let a_line_size = def.line_size(MatrixIdent::A);
+    let a_lines = comptime!(a_elems / a_line_size);
+    let mut a_registers = Sequence::<Line<C>>::new();
```
nathanielsimard: Update the variables to something like the following:

```rust
let elem_count_a;
let line_size_a;
let line_count_a;
```
We could declare all of those variables for `a`, `b`, `c` and `d` in the beginning. The block of code will be easily scanned.

crates/cubecl-core/src/frontend/cmma.rs
```diff
@@ -441,6 +628,80 @@ pub mod cast {
...
+#[allow(unused_variables)]
+pub fn execute_manual<AB: CubePrimitive, CD: CubePrimitive>(
+    matrix: &MmaDefinition<AB, CD>,
+    a_registers: &Sequence<Line<AB>>,
+    b_registers: &Sequence<Line<AB>>,
+    c_registers: &Sequence<Line<CD>>,
+    d_registers: &mut Sequence<Line<CD>>,
+) {
+    unexpanded!()
+}
```
nathanielsimard: I think the API would be better if we would call `execute` on the definition. Execute manual is not the prettiest name.

```rust
def.execute(a, b, c, d)
```

Is a better API.

nathanielsimard: Even the previous cmma API could be migrated toward a definition based API.

crates/cubecl-cpp/src/cuda/processors.rs
```diff
@@ -0,0 +1,129 @@
...
+        MatrixIdent::Accumulator => {
+            let group_id = lane_id / 4;
+            let offset = (i << 2) & 8;
+            group_id + offset
+        }
+    }
+}
+
+#[cube]
+fn col_index(
```
nathanielsimard: I would add a link to the CUDA documentation for those calculations.

crates/cubecl-cpp/src/shared/dialect.rs
```diff
@@ -694,3 +695,8 @@ pub trait DialectWmmaCompiler<D: Dialect>:
     ) -> std::fmt::Result;
     fn supported_wmma_combinations(arch: &D::Architecture) -> SupportedWmmaCombinations;
 }
+
+// Processors
+pub trait DialectProcessors<D: Dialect> {
```
nathanielsimard, line 710: I think more documentation is needed here !

crates/cubecl-ir/src/runtime_properties.rs
```diff
@@ -0,0 +1,34 @@
+use crate::{MatrixLayout, TypeHash};
+
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[derive(Debug, Clone, PartialEq, Eq, TypeHash, Default)]
+pub struct RuntimeProperties {
```
nathanielsimard: What's the difference between RuntimeProperties and DeviceProperties/HardwareProperties in cubecl-core? There probably a reason why this is the IR, but it's not obvious to me here, so more docs would help

crates/cubecl-core/src/frontend/cmma.rs
wingertge: Yeah we actually don't have a consistent term right now for a unit within a plane, as opposed to a unit within a cube. Not sure about the terminology, I just used `lane` because it's fairly universal for SIMD groups

crates/cubecl-core/src/frontend/cmma.rs
wingertge: Arrays require the data to be replicated in each position of the array, while sequence allows reusing the same register multiple times (i.e. a zero register for OOB values).

crates/cubecl-core/src/runtime_tests/cmma.rs
wingertge: The compiler currently gets mad if variables are declared without an assignment in expand code. Something about how we're generating the expands. But putting the letter at the end does make it easier to scan

wingertge: I moved the fragment definitions to the top and added comments to each load loop.

crates/cubecl-core/src/frontend/cmma.rs
wingertge: Oh yeah that's much better. I just essentially copied and modified the existing `execute` function, so never thought about that option.

crates/cubecl-cpp/src/cuda/processors.rs
wingertge: The main issue is there's no single definition, I took all the modern shapes for each element size and derived a generic formula myself. I can put a link to the start of the `Matrix Shapes` section I guess.

crates/cubecl-ir/src/runtime_properties.rs
wingertge: The problem is I'm not sure what to properly call it and it needs a bit more thought. It was just a hacky solution to get runtime variables into the expand scope. The difference with `DeviceProperties` and `HardwareProperties` is those give properties relevant to running the kernel, while `RuntimeProperties` are relevant for compiling it. Ideally, this should be on the `Compiler` trait and be called `CompilerProperties`, but we don't have an easy way to get access to the compiler instance when creating the scope. And you need the instance because we have dynamic compilers.

wingertge: I'll rename it `TargetProperties` and add a comment that it requires more work. I think the PR is big enough, properly integrating it with the compiler would require more changes to how we compile kernels.

crates/cubecl-core/src/frontend/cmma.rs
wingertge: I added docs to explain it to the two functions that use the term. I think we might want to eventually settle on an "official" term for the unit relative to the plane, and also use that term for the position (i.e. `LANE_POS`) instead of `UNIT_POS_PLANE` which is a bit ambiguous.

nathanielsimard (approved):

PR #813, opened 2025-08-12
[Chore] Use `float4`/`float8` for CPU conversion of FP8 and FP4 types

wingertge:
This allows implementing `Numeric` and `CubeElement` for these types, so the CMMA test for FP8 can now run.
I had to slightly modify that test to account for the rounding during conversion, but it seems to be correct.


nathanielsimard (commented):
LGTM. Maybe we could add a feature flags for those new precisions. We don't have to download float4 and float8 when using the vulkan backend for instance.

wingertge:
Not sure if that's worth it, they're both pretty tiny crates. And might be useful anyways for encoding or decoding some models, since MXFP4 is used by default for the new GPT models. I just checked and float4 is 1400 lines including the massive doc comment blocks, so maybe 400-500 lines of actual code. Not worth feature flagging something that small I think.

PR #814, opened 2025-08-12
Fix memory corruption on wgpu

ArthurBrussee:
After #800 , my project is broken on WGPU. This is because the wgpu backend has some tricky requirements wrt memory management.

`create()` (now `write()`) uses write_buffer on wgpu, which is scheduled on the _queue_ not on the encoder, so you have to be super careful the ordering is valid. This was previously done by keeping an exclusion list and not picking any from that list when using create(). This cannot really work for the new `write()` approach.

An alternative is to just `flush()` whenever you're writing something. This is bad though, as kernel dispatches often write some small bits of uniform memory. These are passed in as seperate bindings since a recent refactor however, so we can now efficiently create those seperately, without having to flush() or stop the compute pass.

That also allows to properly mark these bindings as read-only. Ideally they'd be marked as <uniform> but that requires 16 byte strides, so will need some bigger compiler refactor.


crates/cubecl-wgpu/src/compute/server.rs
```diff
@@ -166,7 +166,7 @@ impl ComputeServer for WgpuServer {
             if contiguous_strides(desc.shape) != desc.strides {
                 return Err(IoError::UnsupportedStrides);
             }
-            self.stream.copy_to_binding(desc.binding, data);
+            self.stream.write(desc.binding, data);
```
nathanielsimard, line 169: Does that mean that all scalar inputs are going to trigger a flush?

ArthurBrussee: No, scalar inputs are passed in as a seperate thing to execute(), which is why we're able to have write() do a flush (as its infrequent) and then have a fast path for uniform variables that doesn't flush

nathanielsimard (approved):

PR #815, opened 2025-08-13
[Feat] Block scaled mma

wingertge:
Adds support for CUDA's block scaled MMA instruction for fp8/fp6/fp4 matrices.

Should merge https://github.com/tracel-ai/cubecl/pull/813 first.

# Testing

Adds new test for the block scaled MMA, but only tests fp8 for now because fp6 doesn't support conversion and fp4 needs more thought for how to handle packed values.


crates/cubecl-core/src/runtime_tests/cmma.rs
```diff
@@ -939,6 +942,321 @@ pub fn test_cmma_manual<R: Runtime, AB: CubeElement + Numeric, CD: CubeElement +
...
+                // Dot product over k-dimension
+                let lhs_val = lhs_data[i * k + l]; // LHS[i, l]
+                let lhs_scale = lhs_scales_data[i * scales_factor + l_scales].to_f32();
+                let rhs_val = rhs_data[j * k + l];
+                let rhs_scale = rhs_scales_data[j * scales_factor + l_scales].to_f32();
+                sum += lhs_val * lhs_scale * rhs_val * rhs_scale;
+            }
+            expected.push(sum);
+        }
+    }
```
nathanielsimard, line 1269: Could we extract that in a function? It's probably the same used in many tests?

wingertge: Not sure they're similar enough to be worth pulling into a function. The test setups in that module are all fairly specific and the code required to make it generic would probably end up being more complicated. Just in the new tests, manual and scaled use different matrix layouts (row/row for manual, row/col for the scaled one because it makes fp4 easier to handle if they're already in the right format for CUDA and don't need repacking). The one above needs to handle strided inputs, etc.

wingertge: Also one uses integers because it minimizes precision mismatch, the other one uses f32 because the small floats can't properly represent larger integers (or small integers in the case of e8m0)

PR #816, opened 2025-08-13
Mlir/fixing reduce

marcantoinem:
cubecl-reduce now works with cubecl-cpu, also a config error is added in cubecl-matmul to not try to make an invalid config when plane dim is 1


crates/cubecl-core/src/runtime_tests/line.rs
```diff
@@ -1,18 +1,50 @@
 use crate::{self as cubecl, as_bytes};
 use cubecl::prelude::*;
 
+#[cube(launch_unchecked)]
+pub fn kernel_line_copy<F: Float>(out: &mut Array<F>) {
+    let mut acc = Array::<F>::vectorized(16u32, 1u32);
+    // let mut acc = Line::empty(64u32);
```
nathanielsimard: dead code

nathanielsimard (approved):

PR #817, opened 2025-08-14
Feat/fusion quant

nathanielsimard:
Useful to implement dequantize on read in Burn.


crates/cubecl-quant/src/scheme.rs
```diff
@@ -106,6 +106,16 @@ impl QuantValue {
...
 
+impl QuantStore {
+    /// Returns the size of the quantization input type in bits.
+    pub fn size_bits(&self, value: QuantValue) -> usize {
+        match self {
+            QuantStore::Native => value.size_bits(),
+            QuantStore::U32 => 32,
+        }
+    }
+}
```
laggui: `scheme.size_bits_stored()` should call this now 🙂

PR #822, opened 2025-08-16
[Chore] Refactor conv

wingertge:
Refactors `cubecl-convolution` to match `cubecl-matmul`.
Also adds accumulator to matmul inputs for `A * B + C` form of matmul, and bias for convolution.
Fixes a bug causing deadlock in TMA matmul/convolution, but performance is regressed at the moment due to split barriers being required.


crates/cubecl-matmul/src/components/global/args.rs
```diff
@@ -39,7 +39,7 @@ pub trait ConcreteOutputFactory: LaunchArg {
 /// Arguments for the matrix multiplication algorithm.
 pub trait MatmulArgs: Send + Sync + 'static + Clone {
     /// Type used for the input.
-    type Input<Lhs: Numeric, Rhs: Numeric>: LaunchArg + CubeType;
+    type Input<Lhs: Numeric, Rhs: Numeric, EO: Numeric>: LaunchArg + CubeType;
```
louisfd, line 42: To match the current nomenclature (which could be improved btw), I think it should be EA rather than EO if the goal is to treat the accumulator as an input.

crates/cubecl-convolution/src/components/global/multi_stage/tma/convolution.rs
```diff
@@ -0,0 +1,249 @@
...
+                        Self::LhsLoader::fill_stage(&mut lhs_loader, lhs_barrier, stage, config);
+                        Self::RhsLoader::fill_stage(
+                            &mut rhs_loader,
+                            rhs_barrier,
+                            stage,
+                            stage_config,
+                        );
+
+                        arrive_tma::<LhsS<MP>>(lhs_barrier, stage_elems_lhs);
+                        arrive_tma::<RhsS<MP>>(rhs_barrier, stage_elems_rhs);
```
louisfd, line 180: Now I see what the problem was 😆 About that, the only thing forcing us to have two barriers is that LhsS and RhsS _may_ be different. If they are the same we could in principle keep the original way where they are together, but we'd have to hack our way through the type system. And if they are actually different, I'm not even sure the barrier fundamentally need to be type-dependant.

wingertge: I brought this up to @nathanielsimard on discord too, I'm not sure why the barrier is typed rather than the load op. The linear async loads don't care about the type, and TMA arrive works on bytes. Could be abstracted, but it isn't right now and I think it would be better to get rid of the type at some point so you can wait for combined loads of different types.

This will be even more relevant for block scaling, where you want to load the quantized values and the corresponding scales in parallel and then wait for both since they're tightly coupled.

crates/cubecl-matmul/src/components/global/args.rs
wingertge: But the type of the tensor would probably need to be the output type, and it would need to be converted to the accumulator type after load. As I understand, EA is only for the accumulator type at as local level and not the global memory type.

wingertge: I think in the long run it might be better to switch to the generally used convention of A, B, C, D instead of `Lhs`, `Rhs`, and `Acc`. We currently mix them too, where matrix idents are `A`, `B`, and `Acc`.

crates/cubecl-convolution/src/components/global/multi_stage/tma/convolution.rs
louisfd: Ok I will merge this PR but it would be a priority to refactor the barrier, i made an issue about it: https://github.com/tracel-ai/cubecl/issues/825

crates/cubecl-matmul/src/components/global/args.rs
louisfd: Makes sense, you can leave it like this.

PR #830, opened 2025-08-21
Reduce sum for msl

nathanielsimard:


louisfd (approved):
Not sure all the metal functions are the correct ones, will have to make tests. But let's merge for now

PR #835, opened 2025-08-21
feat: Matmul tensor layouts

wingertge:
Adds a new facility to transform tensors (and eventually shared memory) with custom layouts. Currently used for mapping the 4D convolution tensors to 2D matmul tiles, eliminating the need for a reshape and a custom loader and tensor reader.
It can also be used for swizzling or more complex tensor layouts later (i.e. interleaving, sparse layouts), by simply composing different layouts together.

# Testing
Tests are passing in cubecl and burn

# Performance
Performance in the benchmarks is as good as, if not slightly better than, main. No statistically significant regression is visible.

# Design thoughts
I let myself get inspired by how cutlass handles things and made a less convoluted and more "rusty" version of it. There's a few new pieces:

## Layout
The layout describes a mapping from a set of coordinates into absolute linear space. Layouts can be composed to further increase or decrease dimensionality, or to apply things like swizzing.

## Virtual Layouts
Similar to virtual tensors, allows us to carry a dynamic layout without having to spread a generic everywhere.

## Tensor View
This is the most complex part of the PR, it represents a conceptual tensor mapped to a linear storage, with the storage and layout being entirely dynamic. It's called a `TensorView` because it allows code to "view" a linear buffer as a multidimensional tensor, while abstracting the actual mapping into linear space. A `TensorView` allows reading, writing, and slicing (sort of) using multi-dimensional indices, with the `Layout` being responsible for translating them into absolute indices.

I wanted to minimize changes so haven't fully applied this concept, but in theory this could replace shared memory tiling logic as well, by using a 2D layout that correctly indexes the tiles. We have a `TilingLayout` already, but it doesn't completely handle these responsibilities, and could be unified with the `TensorView` concept to allow for easier composition, including swizzling.


crates/cubecl-convolution/src/components/global/layout/im2col.rs
```diff
@@ -0,0 +1,158 @@
...
+    },
+};
+
+use crate::components::{
+    ConvGemmConfig,
+    global::{layout::virtual_layout, load::im2col_tma::div_mod_seq},
+};
+
+#[derive(CubeType, Clone)]
+pub struct Im2colGlobalLayout<C: ConvGemmConfig> {
```
nathanielsimard: Could you add a bit of docs over what each field is related to the algo!

crates/cubecl-convolution/src/components/global/layout/write.rs
```diff
@@ -0,0 +1,133 @@
...
+
+        (linear_pos, in_bounds)
+    }
+
+    fn shape(this: &Self) -> Self::Coordinates {
+        (1, this.shape_m, this.shape_n)
+    }
+}
+
+mod r#virtual {
```
nathanielsimard: Could you use the macro for this?

crates/cubecl-convolution/src/tests/test_macros/suite.rs
```diff
@@ -148,20 +148,20 @@ macro_rules! conv2d_standard_tests {
...
+
+        // mod p1x2x2 {
+        //     use super::*;
+        //     $crate::conv2d_standard_tests!($tile, PartitionSize { m: 1, n: 2, k: 2 });
+        // }
+
+        // mod p1x4x2 {
+        //     use super::*;
+        //     $crate::conv2d_standard_tests!($tile, PartitionSize { m: 1, n: 4, k: 2 });
+        // }
```
nathanielsimard, line 164: Dead code?

crates/cubecl-matmul/src/components/global/load/loader/sync_full_loader.rs
```diff
@@ -56,7 +58,7 @@ pub struct SyncFullLoader<IP: InputPrecision, G: GlobalConfig, L: SyncFullLoadin
 impl<IP: InputPrecision, G: GlobalConfig, L: SyncFullLoadingStrategy> SyncFullLoader<IP, G, L> {
     /// Create a new SyncFullLoader
     pub fn new(
-        tensor: VirtualTensor<IP::Global>,
+        tensor: TensorView<IP::Global, Coords3d>,
```
nathanielsimard, line 61: Coord3d is pretty much the rank right?

crates/cubecl-matmul/src/components/global/memory/layout.rs
```diff
@@ -0,0 +1,103 @@
...
+
+        (idx, in_bounds)
+    }
+
+    fn shape(this: &Self) -> Self::Coordinates {
+        (1, this.rows, this.columns)
+    }
+}
+
+mod r#virtual {
```
nathanielsimard, line 70: Could this be a macro?

crates/cubecl-matmul/src/components/global/multi_stage/ordered/matmul.rs
```diff
@@ -185,8 +187,9 @@ where
         batch_offset: u32,
         #[comptime] config: Self::Config,
     ) -> Self::RhsLoader {
+        let layout = SimpleGlobalLayout::new(&rhs, config.global_memory_config(MatmulIdent::Rhs));
         SyncPartialLoader::<MP::Rhs, Self::Config, RL>::new(
-            rhs,
+            rhs.view(layout.into_virtual()),
```
nathanielsimard: It would be cool to only have `rhs.view(layout)` maybe by receiving the layout as a generic param and calling `into_virtual` directly into the `view` function. Just syntax sugar.

crates/cubecl-std/src/tensor/layout/base.rs
```diff
@@ -0,0 +1,47 @@
+use cubecl::prelude::*;
+use cubecl_core::{self as cubecl};
+
+use super::Coordinates;
+
+#[cube]
+pub trait Layout: CubeType + Clone + Send + Sync + 'static {
```
nathanielsimard, line 14: We could add documentations here. Especially on how to implement new layouts.

crates/cubecl-matmul/src/components/global/multi_stage/ordered/matmul.rs
wingertge: the reason I didn't do that is because the macro causes dodgy issues around generics, so you always need to manually specify it. So it was actually cleaner this way. This is also why I went with the method on `VirtualTensor`, to abstract away the manually specified type on `TensorView::new`.

crates/cubecl-matmul/src/components/global/memory/layout.rs
wingertge: There's only one in the crate, and I'm not sure I want to make the macro public since it's so limited (doesn't allow for generics, for example). `cubecl-convolution` has a macro for this, because it uses 3 layouts. When we add more layouts to matmul, we can add the same macro used by that.

crates/cubecl-matmul/src/components/global/load/loader/sync_full_loader.rs
wingertge: the conceptual rank, yes. matmul treats the global tensor as a 3D tensor, `(batch, x, y)` (where batch_offset is treated as unit stride in the loader).

wingertge: But `Coord3d` is specifically an alias for the coordinate system for the view, in this case `(u32, u32, u32)`. There's `CoordsDyn` for example that uses `Sequence` for the coordinates to allow for n-dimensional layouts.

crates/cubecl-convolution/src/tests/test_macros/suite.rs
wingertge: I wanted to leave it but disable it for now. It causes a bunch of failures because the bias loader is actually incorrectly implemented for partitions. It was making it hard to determine whether a test failure was caused by the partitioning issues or this PR. We should fix partitioning at some point and re-enable these tests.

crates/cubecl-convolution/src/components/global/layout/write.rs
wingertge: This made me realize I forgot to commit the last refactor I did 😅 It uses one now

crates/cubecl-std/src/tensor/layout/base.rs
wingertge: Absolutely, I gotta fix some bugs that are happening only in burn for some reason and then get to documenting. I wanted the design to settle first before writing whole paragraphs for things that end up being deleted later.

nathanielsimard (approved):

PR #837, opened 2025-08-22
Partition Scheduler

louisfd:
A partition scheduler whose goal is to send each plane to different tiles in shared memory. 

```rust
    fn partition_schedule_scheme(&self) -> PartitionSchedulerScheme {
//        PartitionSchedulerScheme::Naive
        PartitionSchedulerScheme::Offset
    }
 ```


nathanielsimard:
We probably want to put Naive by default for now.

nathanielsimard (approved):

PR #838, opened 2025-08-23
Fix wgpu extensions possibly being emitted multiple times with the same name

amfaber:
## Make sure wgpu extensions are emitted with different names

I experienced a panic with burn due to the function "powf_primitive" being defined multiple times, similarly to https://github.com/tracel-ai/burn/issues/3461.

Digging around in the source code, I think I've found and fixed the issue. During wgpu shader compilation, CubeCL iterates through all the instructions the shader should contain. If any of them warrant introducing, an extension, it is first checked if this extension has already been added to the set of extensions, and if it has not, it is added.

The issues with the old implementation were threefold

 - The extensions are characterized by both the vectorization factor and the element type. However they all get the same name. That means if the same element type with different vectorization levels are used, the two different generated functions will have the same name. To fix this, I added the vectorization level to the function name
 - The vectorized versions are composed of the simpler "primitive" version which always takes a scalar. However in the set of extensions, this primitive extension is also characterized by the vectorization level even though only one copy should ever be added, regardless of vectorization level. To fix this, I changed the associated data of the enum from Item to Elem, as the primitive does not care about vectorization levels at all. This will multiple primitives to compare equal and thus only be added to the set of extensions once.
 - The SafeTanh variant did not distinguish between the "primitive" function and the vectorized variants. The primitive version was also named "scalar" previously - I renamed this to primitive for consistency. This is a problem since two versions of the primitive function would be added if two different levels of vectorization was passed. I haven't been able to test this locally as I don't have a macos computer at hand currently

With these changes, my training pipeline no longer panics.

## Further work
As I've been typing this out, it dawned on me that functions can still be emitted with the same name if different types are used with the same vectorization level. Perhaps this should be addressed as well, or maybe it is impossible to generate the IR that would result in this?
If the IR can be generated, it would be easy to fix by also suffixing the function names with the type

## Burn PR for tests
https://github.com/tracel-ai/burn/pull/3606


nathanielsimard:
> As I've been typing this out, it dawned on me that functions can still be emitted with the same name if different types are used with the same vectorization level. Perhaps this should be addressed as well, or maybe it is impossible to generate the IR that would result in this? If the IR can be generated, it would be easy to fix by also suffixing the function names with the type

Yes it is possible with fusion, where multiple tensors of potentially different data types are merged into a single kernel.

crates/cubecl-wgpu/src/compiler/wgsl/extension.rs
```diff
@@ -1,34 +1,39 @@
-use super::base::Item;
+use super::base::{Elem, Item};
 use std::fmt::Display;
 
 /// Not all functions are native to WGSL, so this struct allows to support more functions.
 #[derive(Debug, PartialEq, Eq, Clone)]
 pub enum Extension {
     PowfScalar(Item),
-    PowfPrimitive(Item),
+    PowfPrimitive(Elem),
```
nathanielsimard, line 8: I agree this is the right way of handling this!

crates/cubecl-wgpu/src/compiler/wgsl/extension.rs
```diff
@@ -1,34 +1,39 @@
...
 }
 
 fn format_powf_scalar(f: &mut core::fmt::Formatter<'_>, item: &Item) -> core::fmt::Result {
+    let vec_factor = item.vectorization_factor();
     match item {
         Item::Vec4(elem) => write!(
             f,
             "
-fn powf_scalar(lhs: {item}, rhs: {elem}) -> {item} {{
+fn powf_scalar_{vec_factor}(lhs: {item}, rhs: {elem}) -> {item} {{
```
nathanielsimard: We should also put the element type in the function definition.

crates/cubecl-wgpu/src/compiler/wgsl/extension.rs
```diff
@@ -1,34 +1,39 @@
...
     return vec4(
         powf_primitive(lhs[0], rhs),
```
nathanielsimard: The implementation looks wrong, we should recursively resolve the inner function name:

```rust
let powf_fn_name = self.powf_fn_name(item);
```

I don't think we need both powf_primitive and powf_scalar extension, only one that all call the one with item.vectorization = 1.

amfaber: The presence of both scalar and primitive confused me in the beginning, but then I figured out that primitive is the one where all inputs are scalars and the return is scalar. The signature of "scalar" is actually lhs being a vector and rhs being a scalar and the return (obviously) being a vector. I decided to just preserve that behavior.

Are you saying that there is no need for the fn(vec, scalar) -> vector variant?

nathanielsimard: Ah I was confused as well I think. I guess it's fine, but we should add the vector length and the element types in function signatures.

amfaber:
Since the logic was getting a bit hairy, I decided to abstract a bit and have the extensions module handle when to use powf / powf_scalar at both the extension generation site and the code generation site.
I also made the function generation programmatic. I'm not entirely sure if I like that or not, its not very readable anymore. Atleast it composes fairly well and I'd argue its easier to see what the difference between powf and powf_scalar is now

amfaber:
What's blocking this? Is there anything I can do to unblock it?

laggui:
> What's blocking this? Is there anything I can do to unblock it?

I think the rust conf was the blocker 😅 just requested another review but it'll probably only happen when he comes back from the conference.

amfaber:
Ah, of course - I hadn't realized rust conf was ongoing :)

amfaber:
Resolved conflicts by incorporating changes from https://github.com/tracel-ai/cubecl/pull/857 and https://github.com/tracel-ai/cubecl/pull/867

PR #839, opened 2025-08-23
feat: Make `StridedLayout` an actual `Layout`

wingertge:
Makes `StridedLayout` an actual layout implementing the new trait, and uses it to fix the random kernel. It was broken for non-contiguous out tensors.
Also adds some quality of life to the tensor views to make them more convenient.

### DIFF
--- a/crates/cubecl-std/src/tensor/virtual.rs
+++ b/crates/cubecl-std/src/tensor/virtual.rs
@@ -253,10 +253,7 @@ impl<E: Numeric, IO: Clone + 'static> VirtualTensor<E, IO> {
 impl<E: Numeric> VirtualTensor<E, ReadWrite> {
     /// Create a mutable conceptual view over this tensor, allowing for multi-dimensional indexing
     /// with custom layouts
-    pub fn view_mut<C: Coordinates>(
-        &self,
-        layout: VirtualLayout<C>,
-    ) -> View<E, C, ReadWrite> {
+    pub fn view_mut<C: Coordinates>(&self, layout: VirtualLayout<C>) -> View<E, C, ReadWrite> {
         View::new_mut::<VirtualTensor<E, ReadWrite>>(*self, layout)
     }
 }


crates/cubecl-matmul/src/components/global/args.rs
```diff
@@ -306,6 +323,14 @@ impl<Lhs: Numeric, Rhs: Numeric, EO: Numeric, MA: MatmulArgs> VirtualTensorOpera
     ) -> ExpandElementTyped<TensorMap<EO>> {
         unimplemented!("TensorOutputExpand can't be turned into a tensor map");
     }
+
+    fn __expand_line_size_method(&self, scope: &mut Scope) -> u32 {
+        TensorOutputExpand::__expand_line_size_method(self.clone(), scope)
+    }
+
+    fn line_size(&self) -> u32 {
+        todo!()
```
nathanielsimard: Should we implement that?

crates/cubecl-matmul/src/components/global/args.rs
```diff
@@ -370,6 +395,14 @@ impl<Lhs: Numeric, Rhs: Numeric, EO: Numeric, MA: MatmulArgs> VirtualTensorOpera
     ) -> ExpandElementTyped<TensorMap<Lhs>> {
         TensorLhsExpand::__expand_as_tensor_map_method(self.clone(), scope)
     }
+
+    fn __expand_line_size_method(&self, scope: &mut Scope) -> u32 {
+        TensorLhsExpand::__expand_line_size_method(self.clone(), scope)
+    }
+
+    fn line_size(&self) -> u32 {
+        todo!()
```
nathanielsimard: Same

crates/cubecl-matmul/src/components/global/args.rs
```diff
@@ -434,6 +467,14 @@ impl<Lhs: Numeric, Rhs: Numeric, EO: Numeric, MA: MatmulArgs> VirtualTensorOpera
     ) -> ExpandElementTyped<TensorMap<Rhs>> {
         TensorRhsExpand::__expand_as_tensor_map_method(self.clone(), scope)
     }
+
+    fn __expand_line_size_method(&self, scope: &mut Scope) -> u32 {
+        TensorRhsExpand::__expand_line_size_method(self.clone(), scope)
+    }
+
+    fn line_size(&self) -> u32 {
+        todo!()
```
nathanielsimard: Same

crates/cubecl-matmul/src/components/global/args.rs
```diff
@@ -498,6 +539,14 @@ impl<Lhs: Numeric, Rhs: Numeric, EO: Numeric, MA: MatmulArgs> VirtualTensorOpera
     ) -> ExpandElementTyped<TensorMap<EO>> {
         TensorAccExpand::__expand_as_tensor_map_method(self.clone(), scope)
     }
+
+    fn __expand_line_size_method(&self, scope: &mut Scope) -> u32 {
+        TensorAccExpand::__expand_line_size_method(self.clone(), scope)
+    }
+
+    fn line_size(&self) -> u32 {
+        todo!()
```
nathanielsimard: Same

crates/cubecl-std/src/tensor/layout/linear.rs
```diff
@@ -0,0 +1,125 @@
...
+            LinearLayout::Strided(strided_layout) => strided_layout.shape(),
+            LinearLayout::Permuted(permuted_layout) => permuted_layout.shape(),
+        }
+    }
+}
+
+virtual_layout!(LinearLayout, LinearLayoutExpand);
+
+pub type LinearTensorView<E, IO = Read> = TensorViewTyped<E, LinearLayout, IO>;
+pub type LinearTensorViewLaunch<'a, R> = TensorViewTypedLaunch<'a, LinearLayout, R>;
```
nathanielsimard: We should add docs if those types are supposed to be used often. Also I'm not a fan of the naming, should we drop the `View` postfix? Only `LinearTensor` is probably enough info. What do you think?

crates/cubecl-std/src/tensor/view.rs
```diff
@@ -0,0 +1,541 @@
...
+use cubecl::prelude::*;
+use cubecl_core::{self as cubecl, io::read_masked, unexpanded};
+
+use crate::tensor::{
+    layout::{Coordinates, VirtualLayout, VirtualLayoutExpand},
+    r#virtual::{Read, ReadWrite},
+};
+
+#[derive(Clone)]
+pub struct TensorView<E: CubePrimitive, C: Coordinates, IO: Clone = Read> {
```
nathanielsimard: We should add a bit more docs here.

nathanielsimard (commented):
Only a few comments! This is going to help a lot making sure every kernel handle the strides correctly!

crates/cubecl-matmul/src/components/global/args.rs
wingertge: Maybe it can just use a dummy scope. The main issue is that it's a `cube` trait so all methods take the scope by default.

wingertge: To be clear, this is specifically for an edge case where you do `comptime!(x.line_size())` on an already expanded type. There's only one occurance in the cubecl codebase rn.

crates/cubecl-std/src/tensor/view.rs
wingertge: Looks like I accidentally put the docs on the expand instead of the base type. Oops

crates/cubecl-std/src/tensor/layout/linear.rs
wingertge: If anything we should drop the `Tensor` part. To me it seems like `Tensor` refers to a concrete structure in memory, with an associated shape/strides. A `TensorView` is just a specific way of viewing the underlying structure, through the "lens" of the layout.

A linear layout allows you to view any kind of underlying structure as a contiguous 1D array, but you can also view the same (or a different) structure as 2D, or 3D, or swap dimensions, etc. So if you're going to shorten it, it should just be a `View`. It applies to any list after all, including shared memory or local arrays.

Of course the launchable version explicitly represents a tensor and is constructed from a `TensorHandleRef`, so I think the launchable version should keep the `Tensor` part of the name.

nathanielsimard: `LinearView` is great, it would be cool to have static method to create the launch type, like `LinearView::launch_from_tensor` or something.

wingertge: that was the idea behind the `linear_tensor` function, I want to support other launchable layouts though that might need extra info so I don't want to make that a method on the typed view itself. So I think a freestanding constructor function is the easiest way to handle it.

wingertge: I did the rename 👍

PR #844, opened 2025-08-27
chore: Item rework

wingertge:
Refactors `Item` to make it easier to handle conceptual types (i.e. `fp4`) vs concrete storage (i.e. packed `fp4x2`). Separates the `Elem` enum (now `ElemType`) from atomics and `E2M1x2`, and introduces a new "middle layer" that handles atomics, packing, and semantic types without a concrete layout like `Barrier`.

# Testing
I tested all compilers I have access too (all except metal and ROCm) on both cubecl and burn.


crates/cubecl-core/src/frontend/element/bool.rs
```diff
@@ -28,8 +28,8 @@ impl CubeType for bool {
 }
 
 impl CubePrimitive for bool {
-    fn as_elem_native() -> Option<Elem> {
-        Some(Elem::Bool)
+    fn as_type_native() -> Option<StorageType> {
```
nathanielsimard, line 31: I'm unsure about `as_type` returning `StorageType` instead of `Type`. The method could be renamed to `as_storage_type_native`. It's verbose, but since it's not spread across the entire codebase, I suppose that's fine.

crates/cubecl-ir/src/item.rs
```diff
@@ -52,74 +50,82 @@ pub enum UIntKind {
...
+}
+
+/// Physical type containing one or more elements
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[derive(Debug, Clone, Copy, TypeHash, PartialEq, Eq, Hash, PartialOrd, Ord)]
+pub enum StorageType {
+    /// `ElemType` is the same as the physical type
+    Scalar(ElemType),
+    /// Packed values of type `ElemType`
+    Packed(ElemType, u32),
```
nathanielsimard, line 78: The second argument is the count I guess?

crates/cubecl-ir/src/item.rs
```diff
@@ -301,37 +278,194 @@ impl Elem {
...
+        Type::new(StorageType::Semantic(val))
+    }
+}
+
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[derive(Debug, Clone, Copy, TypeHash, PartialEq, Eq, Hash, PartialOrd, Ord)]
+pub struct Type {
+    pub storage: StorageType,
+    pub line_size: LineSize,
+}
```
nathanielsimard: Do you think we should change the `Type` struct to an enum? It currently has only a single variant, but this would open up the API for non-breaking changes by allowing more variants later, potentially even user-defined types.

Also we could remove `Semantic` from `StorateType` since it doesn't make sense for it to have a line size. So I'm thinking ...

```rust
pub enum Type {
    Primitive(PrimitiveType, LineSize), // Current storage type
    Composite(CompositeType), // Composite types could be Array, Tensor, etc.
    Accelerator(AcceleratorType), // barrier, pipeline, tensor map. Things that are related to accelerator.
    ...
}
```

crates/cubecl-ir/src/item.rs
```diff
@@ -301,37 +278,194 @@ impl Elem {
...
+
+pub type LineSize = Option<NonZero<u8>>;
```
nathanielsimard: It should be a single `u8` and 0 is the same as not defined. `Some(0)` would be wrong right now, that's why we have `NonZero` but if `0 == None` we don't have that problem. The Option and NonZero is so unconvenient.

crates/cubecl-ir/src/item.rs
wingertge: yes. I figured it's obvious enough even without a named field.

crates/cubecl-core/src/frontend/element/bool.rs
wingertge: I'm not sure, `StorageType` is still a type, and it's specifically the concrete type, that's why it's the one implemented on the actual Rust types.  I feel like the naming is actually more clear this way.

crates/cubecl-ir/src/item.rs
wingertge: the reason I didn't do that at least for now is because it makes things very awkward for code that changes the line size, especially code that sets it to an optional size. But I might just prototype that version and see how it is to actually use in the expands.

wingertge: as for the naming, I didn't use primitive for a reason. It's actually pretty ambiguous, because you could be talking about hardware primitives (so it would be the `StorageType`), or the primitive type that makes up a concrete one (so the `ElemType`). I believe LLVM uses it in the latter way, for example. So I decided to go with `ElemType` and `StorageType`, since a `StorageType` is a concrete type that describes how the elements are actually stored.

wingertge: It did turn out to be nicer, so I applied that suggestion (minus the naming).

crates/cubecl-ir/src/item.rs
wingertge: Done

crates/cubecl-ir/src/type.rs
```diff
@@ -301,37 +275,223 @@ impl Elem {
...
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[derive(Debug, Clone, Copy, TypeHash, PartialEq, Eq, Hash, PartialOrd, Ord)]
+pub enum Type {
+    /// Scalar type containing a single storage element
+    Scalar(StorageType),
+    /// Line wrapping `n` storage elements
+    Line(StorageType, u32),
+    /// No defined physcial representation, purely semantic. i.e. barrier, pipeline
+    Semantic(SemanticType),
+}
```
nathanielsimard, line 381: That's great! Very clear, and yeah line size is better as a u32!

nathanielsimard (approved):

PR #855, opened 2025-09-02
feat: Stacked layouts

wingertge:
Changes layout to always transform coordinates between different systems, instead of forcing the output coordinates to be linear. This way, views can now be stacked. A view with `Coords4d` can be turned into a view of `Coords2d` by stacking a `Coords4d -> Coords2d` layout on top of it.
To do this, the concrete storage types are treated as a 1D view with a simple passthrough layout.

Also makes the primitive type of the view the actual primitive, rather than forcing lines. This choice was more of a remnant from when it was restricted to `VirtualTensor` (which always uses lines).

# Testing
Convolution now uses the stacked layouts and all tests pass.


crates/cubecl-std/src/tensor/view/as_view.rs
```diff
@@ -0,0 +1,97 @@
...
+    }
+}
+
+pub trait AsViewExpand<E: CubePrimitive> {
+    fn __expand_view_method<C: Coordinates + 'static>(
+        self,
+        scope: &mut Scope,
+        layout: VirtualLayoutExpand<C, Coords1d>,
+    ) -> ViewExpand<E, C, ReadOnly>;
+}
```
nathanielsimard, line 27: Does the expand trait can be generated with the `#[cube]` macro?

wingertge: Not right now but that's something I'd like to implement at some point. Currently we don't support split traits, and after trying to add a very simple `split` flag to the macro I realized there's more rearchitecting needed so I put it off for now. It currently always generates one trait for both the base and expand, which doesn't work for methods.

PR #857, opened 2025-09-03
feat: Add powi for CUDA/HIP

wingertge:
Adds new test for powi as well.
Not supported on Vulkan/WGSL/Metal, so it casts to f32 instead


crates/cubecl-cpp/src/shared/binary.rs
```diff
@@ -221,6 +222,55 @@ impl<D: Dialect> Binary<D> for Powf {
...
+    ) -> std::fmt::Result {
+        let elem = item.elem;
+        let lhs = lhs.to_string();
+        let rhs = rhs.to_string();
+        match elem {
+            Elem::F16 | Elem::F16x2 | Elem::BF16 | Elem::BF16x2 => {
+                let lhs = format!("float({lhs})");
+
+                write!(f, "{elem}(")?;
+                D::compile_instruction_powf(f, &lhs, &rhs, Elem::F32)?;
```
nathanielsimard, line 243: Is it supposed to be powf here instead of powi?

wingertge: yes, because powi is just an overload for powf in CUDA/HIP. It's a bit confusing because of the `f`, but that actually just stands for 32-bit here (as opposed to `pow` for doubles). The signature for this overload in CUDA is `powf(float, int)`.

PR #859, opened 2025-09-04
Introduced a new Virtual Storage type in cubecl-cuda

RaulMoldes:
## PR Requirements

1. Confirmed that ```cargo xtask validate`` command has been executed
2. Confirmed that ```cargo run checks`` command has been executed on burn after replacing the commit ID with that of this PR.
3. Validated PR with burn.

## Summary
Implements **CUDA Virtual Memory Management (VMM)** for CubeCL, introducing a new **VirtualStorage** struct that eliminates memory fragmentation through address space separation. This should provide a more efficient alternative to the current allocation strategy, specially during batched inference, or any workload in which the batch size can change across iterations.

## Core Components
* VirtualStorage: Main allocator implementing ComputeStorage trait
* HandlePool: internal FIFO pool for physical memory handle reuse. De-couples the memory mapping logic from the physical memory allocation, allowing deallocated handles to be remapped to new blocks.
* VirtualBlock: Represents contiguous virtual memory ranges.

## Characteristics:
* **Backward Compatible**: Opt-in feature, existing code continues working
* **ComputeStorage Trait impl**: Seamless integration with current memory management
* The new **CudaStorageType Enum**: Allows for runtime dispatch between traditional and virtual storage
* **Unified Interface**: Same API as existing storage implementations.

## Breaking changes
None

## Notes
The virtual memory implementation follows NVIDIA's best practices from their [Virtual Memory Management blog post](https://developer.nvidia.com/blog/introducing-low-level-gpu-virtual-memory-management/). The handle size is deliberately aligned to GPU memory minimum granularity (2MB). I also want to indicate that I did not add support for sharing mapped memory across devices (e.g. via RDMA). It would be easy to add, though, but I am not sure that the current CubeCL architecture supports it.

## Questions
1. Should I have added a feature flag to gate virtual memory support behind cuda-vmm feature?
2. Would you prefer immediate deallocation instead of the lazy pattern for better VMM performance?

### DIFF
--- a/crates/cubecl-cuda/src/runtime.rs
+++ b/crates/cubecl-cuda/src/runtime.rs
@@ -232,7 +232,7 @@ fn create_client<M: DialectWmmaCompiler<CudaDialect<M>>>(
     let virtual_storage_gpu = GpuVirtualStorage::new(device_ptr, granularity);
     let memory_management_gpu = MemoryManagement::from_configuration(
         storage,
-        None, // PLACEHOLDER. NEED TO SET THIS
+        Some(virtual_storage_gpu),
         &mem_properties,
         options.memory_config.clone(),
     );
--- a/crates/cubecl-cuda/src/compute/storage/gpu.rs
+++ b/crates/cubecl-cuda/src/compute/storage/gpu.rs
@@ -234,6 +234,7 @@ impl GpuMemoryRange {
     ///
     /// This should be called after successful `cuMemUnmap` operations to allow
     /// for future remapping or release of the virtual address space.
+    #[allow(dead_code)]
     fn set_unmapped(&mut self) {
         self.is_mapped = false;
     }
@@ -560,6 +561,7 @@ impl VirtualStorage for GpuVirtualStorage {
     /// invalid handles indicate programming bugs that should be caught during
     /// development rather than handled at runtime.
     fn get(&mut self, handle: &StorageHandle) -> Self::Resource {
+      
         let mapping = self.active_mappings.get(&handle.id)
             .expect("Storage handle not found in active mappings. Handle may be unmapped, deallocated, or invalid");
 
--- a/crates/cubecl-runtime/src/memory_management/memory_manage.rs
+++ b/crates/cubecl-runtime/src/memory_management/memory_manage.rs
@@ -252,10 +252,9 @@ impl<Storage: ComputeStorage, Virtual: VirtualStorage> MemoryManagement<Storage,
             })
             .collect();
 
-        let virtual_pool = if properties.virtual_memory && virtual_storage.is_some() {
-            let virtual_ref = virtual_storage
-                .as_ref()
-                .expect("This is unreachable, as virtual storage must be defined at this point.");
+        let virtual_pool = if properties.virtual_memory
+            && let Some(ref virtual_ref) = virtual_storage
+        {
             Some(VirtualStaticPool::new(
                 virtual_ref,
                 properties.min_granularity as u64,
@@ -307,10 +306,10 @@ impl<Storage: ComputeStorage, Virtual: VirtualStorage> MemoryManagement<Storage,
 
     /// Returns the storage from the specified binding
     pub fn get(&mut self, binding: SliceBinding) -> Option<StorageHandle> {
-        if self.uses_virtual() {
-            if let Some(val) = self.static_pool.get(&binding) {
-                return Some(val.clone());
-            }
+        if self.uses_virtual()
+            && let Some(val) = self.virtual_pool.as_mut()?.get(&binding)
+        {
+            return Some(val.clone());
         }
 
         if let Some(val) = self.static_pool.get(&binding) {
@@ -334,15 +333,20 @@ impl<Storage: ComputeStorage, Virtual: VirtualStorage> MemoryManagement<Storage,
         let handle = self.get(binding);
 
         if self.uses_virtual() {
+
+
             handle.map(|handle| {
                 let handle = match offset_start {
-                    Some(offset) => handle.offset_start(offset),
+
+                    Some(offset) =>{
+                     handle.offset_start(offset)},
                     None => handle,
                 };
                 let handle = match offset_end {
                     Some(offset) => handle.offset_end(offset),
                     None => handle,
                 };
+
                 self.virtual_storage().get(&handle) // Devuelve Option<Virtual::Resource> (impl Send)
             })
         } else {
@@ -363,6 +367,7 @@ impl<Storage: ComputeStorage, Virtual: VirtualStorage> MemoryManagement<Storage,
     /// Finds a spot in memory for a resource with the given size in bytes, and returns a handle to it
     pub fn reserve(&mut self, size: u64) -> Result<SliceHandle, IoError> {
         if self.uses_virtual() {
+            println!("LLAMO A ALLOC");
             let virtual_storage = self.virtual_storage.as_mut().unwrap();
             return self
                 .virtual_pool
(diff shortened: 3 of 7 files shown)


build.sh
```diff
@@ -0,0 +1 @@
+404: Not Found
```
nathanielsimard: unused file

build_llvm.sh
```diff
@@ -0,0 +1,80 @@
+#!/bin/bash
```
nathanielsimard: I don't think this is related to the PR.

crates/cubecl-cuda/src/compute/storage.rs
```diff
@@ -154,3 +184,651 @@ impl ComputeStorage for CudaStorage {
...
+        } else {
+            Ok(())
+        }
+    }};
+}
+
+// I decided to implement this simple FIFO pool instead of creating a storage that returns handles and create a new pool in the memory manage module because I think it fits more with your current structure.
+// The goal of the pool is to automatically manage physical memory reusability. This way when blocks are deallocated by the pool, they are just unmapped, but memory is not completely released until pool goes out of scope or you call the [`flush`] method inside the pool.
+// Think of it as a data structure owned by the [`VirtualStorage`] that just holds the physical memory, which when working with VMM, must be separated from virtual memory (represented as [`VirtualBlocks`] in my implementation)
+struct HandlePool {
```
nathanielsimard: I'm not sure I would use a memory pool inside the `Storage`. This is actually handled by memory management algos where we have a lot of different pool configurations.

The only reason to have a pool in the storage would be to support non-contiguous memory. A single handle that points to virtual memory where the mapping is on multiple hardware blocks. Otherwise, pools can be multi-platform and work with any provided storage type.

RaulMoldes:
Hi Nathaniel. First of all, thanks for the review and the comments.

I saw the pools when implementing it. The reason I decided to introduce the
pool inside the storage is mainly because the main advantage of using
Virtual Memory is that it enables you to decouple the physical handles from
the actual mapped memory. This way you could reuse physical memory handles
via mapping/unmapping them to contiguous memory ranges. As you said, this
guarantees that memory ranges are always contiguous (virtually).

I see your point, but then I have a question. Your existing memory pools do
not have a mapping method. Your storage trait neither does. Therefore, to
properly implement virtual memory, you would need to either add a new pool
or modify the existing ones. I do not think you want to use a feature that
is specific to some versions of cuda or maybe ROCM ( i am not so much into
other platforms, sorry), in your multi-platform memory pools. That is way I
decided to kept it internal to the storage. Anyway, I appreciate your
advice, and will  give it a thought.

(Sorry for the extra files, i forgot to remove them)

THanks!

nathanielsimard:
@RaulMoldes We have slices instead of map and unmap. If your virtual memory pool allows to merge blocks that aren't contiguous in physical memory, I would say that it makes sense to have it here. Otherwise, slicing works just fine.

nathanielsimard:
@RaulMoldes By looking at the code more carefully I think a mapped virtual memory can be composed on non-contiguous physical handles, which is the main reason to use this over the current storage allocator. So overall I would make it clearer in the current documentation that the pool is doing that! 

I'm still unsure if we should move the pool to `cubecl-runtime`, maybe by adding  a new trait `VirtualStorage`, which `cubecl-cuda` would implement.

I'm unsure if we should combine the current slice memory pool with the virtual mapped memory pool, we would have to benchmark models and look at the padding added for total allocations.

RaulMoldes:
I see. It actually should allow you, indeed, if I am not missing anything.
Deallocated physical handles do not necessary need to come from a block of
any address or any size. That is the point of having the handle pool there.
When you make a new allocation you go to the freelist and there you will
have handles which previously where on whichever address, but will be
remapped into a new block of the size you specify. With slices I think that
is not completely possible as far as I know

crates/cubecl-cuda/src/compute/storage.rs
```diff
@@ -154,3 +184,651 @@ impl ComputeStorage for CudaStorage {
         self.perform_deallocations();
     }
 }
+
+
+// This is a utility I added to avoid repeating CUDA error checking code
```
wingertge: This is built in for cudarc, just call `.result()` on the return value

RaulMoldes:
O I had not read your last message! Sorry for the documentation, I tried my
best but I know I am not super good at that haha. I can make it clearer if
that is what you like.
I thought about the virtual storage trait when implementing it.  Perhaps
you could do something similar to the storage trait, but adding map/unmap
methods. That should give you the flexibility to handle it at memory
management level. Just happy that you understood the point I tried to make.
Might make a new attempt for a PR once I have the time!

crates/cubecl-cuda/src/runtime.rs
```diff
@@ -93,15 +109,73 @@ fn create_client<M: DialectWmmaCompiler<CudaDialect<M>>>(
...
         let mut bytes = MaybeUninit::uninit();
         cuDeviceTotalMem_v2(bytes.as_mut_ptr(), device_ptr);
         bytes.assume_init() as u64
     };
-    let storage = CudaStorage::new(mem_alignment, stream);
-    let mem_properties = MemoryDeviceProperties {
-        max_page_size: max_memory / 4,
-        alignment: mem_alignment as u64,
+
+    let (mem_properties, storage) = if !vmm_supported && !options.expandable_storage_enabled {
```
wingertge: Pretty sure this should be `||`, not `&&`

.vscode/settings.json
```diff
@@ -1,3 +1,17 @@
 {
-    "files.eol": "\n"
-}
\ No newline at end of file
+    "files.eol": "\n",
```
wingertge, line 2: This shouldn't be commited to the repo

Cargo.toml
```diff
@@ -91,5 +91,6 @@ futures-lite = { version = "2.3.0", default-features = false }
 sysinfo = "0.36.1"
 
 [profile.dev]
-debug = 0     # Speed up compilation time and not necessary.
+incremental = true
```
wingertge, line 117: Should reset to normal values

crates/cubecl-cuda/src/compute/storage.rs
```diff
@@ -154,3 +184,651 @@ impl ComputeStorage for CudaStorage {
...
+
+        let mut handle = 0;
+        let handle_type = {
+            #[cfg(unix)]
+            {
+                CUmemAllocationHandleType_enum::CU_MEM_HANDLE_TYPE_POSIX_FILE_DESCRIPTOR
+            }
+            #[cfg(target_os = "windows")]
+            {
+                CU_MEM_HANDLE_TYPE_WIN32
```
wingertge: Should be `CUmemAllocationHandleType_enum::CU_MEM_HANDLE_TYPE_WIN32`

crates/cubecl-cuda/src/lib.rs
```diff
@@ -39,4 +40,10 @@ mod tests {
     cubecl_reduce::testgen_reduce!([f16, bf16, f32, f64]);
     cubecl_random::testgen_random!();
     cubecl_reduce::testgen_shared_sum!([f16, bf16, f32, f64]);
+
+    // NOTE THIS TESTS ARE MEMORY INTENSIVE. I ADDED THEM TO THE CRATE BUT NOT SURE TO ADD THEM HERE SINCE ON MY LOCAL GPU OTHER TESTS ARE FAILING WITH OOM (maybe due to that).
```
wingertge: These tests don't seem to be commited?

crates/cubecl-cuda/src/runtime.rs
```diff
@@ -93,15 +109,73 @@ fn create_client<M: DialectWmmaCompiler<CudaDialect<M>>>(
...
+        let handle_type = {
+            #[cfg(unix)]
+            {
+                CU_MEM_HANDLE_TYPE_POSIX_FILE_DESCRIPTOR
+            }
+            #[cfg(target_os = "windows")]
+            {
+                CU_MEM_HANDLE_TYPE_WIN32
+            }
+        };
```
wingertge: This causes an unresolved import and an unused import on Windows, should be qualified with `CUmemAllocationHandleType_enum`

PR #861, opened 2025-09-04
feat: add trigonometric functions

relativityhd:
# Add trigonometric functions (atan2 etc.)

Adds the following functions for all Floats:

- [x] Hyperbolic Sine (`sinh`)
- [x] Hyperbolic Cosine (`cosh`)
- [x] Inverse Sine (`asin`)
- [x] Inverse Hyperbolic Sine (`asinh`)
- [x] Inverse Cosine (`acos`)
- [x] Inverse Hyperbolic Cosine (`acosh`)
- [x] Inverse Tangent (`atan`)
- [x] Inverse Hyperbolic Tangent (`atanh`)
- [x] 2-argument arctangent (`atan2`)
- [x] Degrees (`degrees`)
- [x] Radians (`radians`)

## Open Questions / Missing parts

- [x] LLVM MIR implementation
- [x] Non f64 / f32 correctness
- [x] Metal atomic / safe operations

### LLVM MIR implementation

I have setup a placeholder currently with code which I assumed to work commented out.
It seems that support for these functions needs to be added in the tracel-llvm repository
, but I have no clue where.

### Non f64/f32

At some point there is limited support for the "special" floats and for e.g. powf which I used as an example a conversion was needed to normal floats.
I am unsure at which points this is necessary and where I can find out whether these conversions are necessary.

### Metal safe operations

Since I used the existing sin, cos, tanh and powf functions as examples on how to add the other functions, i stumbled across the implementation for metal for tanh:

```rust
    fn compile_instruction_tanh_scalar<T: Component<Self>>(
        f: &mut std::fmt::Formatter<'_>,
        input: T,
    ) -> std::fmt::Result {
        write!(f, "safe_tanh_scalar({input})")
    }
```

I couldn't find anything about this "safe" version in the metal documentation, but I am clearly not an expert.
For which functions is a "safe" implementation needed and which are fine without?

### Instructions

- [] Create a new branch or fork of the [burn repo](https://github.com/Tracel-AI/burn)
- [] Update the main [Cargo.toml](https://github.com/tracel-ai/burn/blob/40aa993fb5969351a006b560ec89da5119dc9721/Cargo.toml#L159=L161) with this PR hash.
- [] Fix any broken tests or compilation errors in burn.
- [] Submit a PR in burn with your fixes and link it here.


relativityhd:
`cargo xtask validate` just takes too long on my machine, aborted after 10 minutes of execution. I have run `cargo test -p cubecl-wgpu` and `cargo test -p cubecl-core` instead and they seem to work.

wingertge:
Not sure if it might be a good idea to pull all the trigonometry functions (including the existing ones) into a new enum for the IR, because that's a lot of added stuff, and it's now enough to warrant its own category I think. We should at some point also think about applying a similar separation to the compilers themselves, but that's a larger rework that wouldn't go into this PR.

I'm currently looking into the MLIR functions for that stuff - the bindings are auto-generated I believe, so if the functions don't exist, they might be in a different namespace or need a different way to handle them.

As for the float types, casting to float for unsupported float types (i.e. F16, BF16) is reasonable. There do appear to be double versions of the functions, so that will work natively.

crates/cubecl-cpu/src/compiler/visitor/operation/arithmetic.rs
```diff
@@ -28,6 +28,117 @@ impl<'a> Visitor<'a> {
                 let result = self.append_operation_with_result(operation);
                 self.insert_variable(out, result);
             }
+            Arithmetic::Sinh(_sinh) => {
```
wingertge: The relevant functions are actually in the [`dialect::ods::math`](https://mlir-rs.github.io/melior/melior/dialect/ods/math/index.html) module (not `dialect::ods::llvm`). So you'll need to import that and adjust as appropriate.

crates/cubecl-cpu/src/compiler/visitor/operation/arithmetic.rs
```diff
@@ -28,6 +28,117 @@ impl<'a> Visitor<'a> {
...
+                todo!("intr_atan does not exist")
+                /*let value = self.get_variable(acos.input);
+                let result = self.append_operation_with_result(llvm_ods::intr_atan(
+                    self.context,
+                    value,
+                    self.location,
+                ));
+                self.insert_variable(out, result);*/
+            }
+            Arithmetic::Degrees(_degrees) => {
```
wingertge: Degrees and radians are just constant multiplications, so they don't have intrinsics. You'll need to decompose them into `x * (180.0 / PI)` and `x * (PI / 180.0)` respectively.

crates/cubecl-cpu/src/compiler/visitor/operation/arithmetic.rs
relativityhd: I implemented everything. Now I get this error message when running tests:
error: cannot be converted to LLVM IR: missing `LLVMTranslationDialectInterface` registration for dialect for op: math.atan2
However, according to code hints everything seems fine...

relativityhd:
@wingertge 

> Not sure if it might be a good idea to pull all the trigonometry functions (including the existing ones) into a new enum for the IR, because that's a lot of added stuff, and it's now enough to warrant its own category I think. 

Must I implement this in this PR or will this be part of another refactor?

> As for the float types, casting to float for unsupported float types (i.e. F16, BF16) is reasonable. There do appear to be double versions of the functions, so that will work natively.

Can you please double-check my implementation? I don't really understand at which point the right functions are written to the shader for each dialect...

crates/cubecl-cpu/src/compiler/visitor/operation/arithmetic.rs
wingertge: I'll have to tag in @marcantoinem here because I don't understand much about the MLIR backend. I'm guessing the dialect needs to be registered somewhere but not sure about the details.

wingertge:
> @wingertge
> 
> > Not sure if it might be a good idea to pull all the trigonometry functions (including the existing ones) into a new enum for the IR, because that's a lot of added stuff, and it's now enough to warrant its own category I think.
> 
> Must I implement this in this PR or will this be part of another refactor?

I think that would be better to do after all the outstanding work has been merged, so I'll do it separately.
One more thing I'd add retroactively is that since there's no intrinsics for converting degrees/radians, I think it would be better to just implement those as `cube` functions in `cubecl-std`, rather than the compiler. The implementation would be trivial:
```rust
#[cube]
pub fn to_degrees<F: Float>(val: F) -> F {
    val * F::new(180.0 / f32::PI)
}
```
It would keep it in one place rather than implemented separately for each compiler.

wingertge:
> Can you please double-check my implementation? I don't really understand at which point the right functions are written to the shader for each dialect...

CUDA is already correct, not sure what the WGSL behaviour even is because it's an experimental extension. It might just support f16 overloads on those functions by default, but I can't yet test it because of some issues with features on Vulkan.
For SPIR-V I think it would be best implemented by a preprocessor but they're poorly documented, so I think the easiest way to do it is to open an issue to track that it still needs to be done and I'll deal with it myself.

relativityhd:
> there's no intrinsics for converting degrees/radians

There are intrinsics for WGSL and SPIR-V (at least rspirv-ext), also Rusts f32 and f64 do have to_degrees and to_radians operations. Also, as a user, I would probably search in the Float Namespace for these functions instead of the cubecl-std. That's why I think having them there would make still sense.

I removed the dialect specific compile_instruction_xxx_scalar calls and instead hardcode the operations directly for the cubecl-cpp, since all CUDA, HIP and Metal should convert implicit to the right type and share the same syntax.
I also added the to_degrees and to_radians functions to the cubecl-std, so that you can decide which one fits best / should survive a refactor.

crates/cubecl-cpu/src/compiler/visitor/operation/arithmetic.rs
marcantoinem: Good catch, I didn't see that there is a dialect other than arith for more complex mathematical operation, I was using llvm intrisic instead. The error you got is because the pass to transform the math dialect to llvm intrisic is not registered. To fix it you just needs to add `pass_manager.add_pass(pass::conversion::create_math_to_llvm());` in `src/compiler/module.rs`

relativityhd: Locally, I still get the same error, even after run cargo clean, deleting the target directory and passing --features mlir-dump to cargo test -p cubecl-cpu. I also checked whether the line was called and it seems like it did - but nothing happend (maybe because I missed another cache?)
I pushed a commit with the added line, maybe it runs on your machine?

marcantoinem: I tried your code and I get the same error the pass seems to do nothing. I didn't use it anywhere else in the MLIR compiler so it must be a problem on the C API of MLIR that is not running correctly this pass. For the moment you can use LLVM intrisic I think, it will needs more investigation to find why registering the math_to_llvm doesn't work. The only problem with using llvm intrisic instead of math is for portability if we want to port the MLIR compiler to GPU,. but it is not a concern right now.

relativityhd: The arc-* operations are not available in the llvm intrinsic, hence I disabled them for now until the math module becomes available.

relativityhd:
Before things are getting stale, I mark this as ready to review.

Currently, 21 tests are failing, 20 of them related to the missing CPU implementation of the inverse trigonometric functions (arc-*) and one is that `to_degrees` fails for `f16`:

```
Values differ more than epsilon: actual=89.9375, expected=90, difference=0.0625, epsilon=0.020004272
index: 1
actual: [0.0, 89.9375, 179.875, -89.9375, -179.875]
expected: [0.0, 90.0, 180.0, -90.0, -180.0]
```

For both problems I can't do more, since the first one needs to be fixed by @marcantoinem I guess? and the second one needs to be fixed by a design decision whether `to_degrees` (and `to_radians`) should be present for Float types or not and if yes, whether it should be okay to lower the test-epsilons for this.

If I should do more, let me know.

crates/cubecl-std/src/lib.rs
```diff
@@ -23,3 +24,13 @@ pub mod tests;
...
+
+#[cube]
+pub fn to_degrees<F: Float>(val: F) -> F {
+    val * F::new(180.0 / f32::consts::PI)
+}
+
+#[cube]
+pub fn to_radians<F: Float>(val: F) -> F {
+    val * F::new(f32::consts::PI / 180.0)
+}
```
nathanielsimard: I would put that under a trigo module. It doesn't make sense to be at the root of std. We should also add documentation, since those are public facing APIs

wingertge:
> Currently, 21 tests are failing, 20 of them related to the missing CPU implementation of the inverse trigonometric functions (arc-*) and one is that `to_degrees` fails for `f16`:

that epsilon looks too tight for f16, might have another bug in the precision-aware comparison like when I forgot an `abs` and made the tolerance way too tight for negative values. This one is positive, but seems excessively precise regardless so something needs to be adjusted there. Maybe we can modify all tests to specify epsilon, since it can depend on the exact operation.

relativityhd:
I changed the runtime tests for `unary` so that epsilon is now an optional paramater for the macro, defaulting to 0.02, and the epsilon for the to_degrees test set to 0.3 -> This test is now passing.

I moved the `to_degrees` and `to_radians` functions to their own trigonometric module and added documentation and tests.
I also asked an LLM to add more trigonometric functions so that this module is not so empty... I validated everything, I think most functions seem useful to some extent and all tests pass except for one functions with the cpu code because tan2 is not correctly implemented there and therefore also blocked by @marcantoinem MLIR magic :D
I see that this was somewhat unnecessary, so if you want, I can delete this.

relativityhd:
Since I used a lot of PI and PI*2 etc., especially in the Tests, I wondered whether it would be useful to expose constants under the Float trait. I know that rust f32 and f64 already have a lot of constants under ::consts and that there are also a lot of constants already implemented for the cubecl f16 type.

relativityhd:
Another thing: I wondered whether this function signature [here](https://github.com/tracel-ai/cubecl/blob/6c213fbfff429a89bce80cdc838f1f00b4086737/crates/cubecl-core/src/frontend/element/float.rs#L133-L135) would prevent the creation of super-precise f64 floats, since it is only possible to pass in f32.

nathanielsimard:
The code looks fine to me, but the tests don't pass on the CI. @relativityhd We may also remove some functions in the trigo modules, unsure they are necessary.

relativityhd:
@nathanielsimard yes the CI is failing because as I mentioned this PR is blocked by the MLIR melior project, because the math module there wont register correctly, as @marcantoinem mentioned.
The tests are failing because of the empty implementations of the functions depending on the melior llvm math module.

> We may also remove some functions in the trigo modules, unsure they are necessary.

Sure, just tell which ones I should remove.

nathanielsimard:
> Sure, just tell which ones I should remove.

The ones that are not stricly necessary, using if statements is also very bad on most hardware, so we can remove the functions that have if statements.

relativityhd:
Sorry for the long wait - my vacation is over and time limited again...

I've removed unnecessary trig functions from the std, only keeping hypot, to_radians and to_degrees since they are also present in rusts f32.

This PR is still blocked by the MLIR melior project. I still get the following error when testing (I've removed the dummy code and enabled ods math support again, as @marcantoinem described):

```
error: cannot be converted to LLVM IR: missing `LLVMTranslationDialectInterface` registration for dialect for op: math.acos
```

@nathanielsimard What repository is responsible for this? Maybe I can check and fix it there. I am quite confused about the different LLVM repositories and packages used.

nathanielsimard:
> @nathanielsimard What repository is responsible for this? Maybe I can check and fix it there. I am quite confused about the different LLVM repositories and packages used.

https://github.com/tracel-ai/tracel-llvm :)

relativityhd:
So I checked whether I would find any reason why the math dialect is not properly registered, but I just can't find it. Seems to be present everywhere where other dialects are also present, e.g. Arith.
But I am also completely foreign to LLVM and compilers in general, so I think someone of your group should take a look here.
@marcantoinem can you maybe take a look and find out what is missing?

crates/cubecl-cpu/src/compiler/visitor/operation/arithmetic.rs
vaijira: With this change https://github.com/tracel-ai/tracel-llvm/pull/6 and the one suggested by @marcantoinem  pass_manager.add_pass(pass::conversion::create_math_to_llvm()); in src/compiler/module.rs it seems to work for me.

vaijira: It works for me only with pass_manager.add_pass(pass::conversion::create_math_to_llvm()); in src/compiler/module.rs i. It should be my linux distribution version.

relativityhd: Thanks for testing this!

I also added pass_manager.add_pass(pass::conversion::create_math_to_llvm()); in src/compiler/module.rs but it does not work - and the tests in the CI also don't work.

Do you think you could pinpoint what exactly mismatches in the linux distribution version?

vaijira: I'm not testing this patch exactly i was testing a implementation for rsqrt and It seems math.rsqrt works for me but not math.sin or math.asinh or math.sinh, weird.I'm on ubuntu 25.10 but i don't think that is the issue now. Do math.rsqrt works for you?

relativityhd: hmm yes, this is odd, rsqrt works but sinh etc. not. I am on Fedora 42, but I would agree that the OS should not be relevant here...

relativityhd: Where is this actually implemented? I know of the tracel-llvm project, but this just pulls from LLVM, so maybe there is something missing?

vaijira: the crate tracel-mlir-sys generate the bindings from llvm. The wrapper for the bindings are in crates/tracel-mlir-rs  it implements the ods dialect operations in "cargo expand  -p tracel-mlir-rs --features ods-dialects dialect::ods"

vaijira: I think i got it for other mathematical conversions you need to add too: pass_manager.add_pass(pass::conversion::create_math_to_libm());

relativityhd: Yes and as far as I can see there is nothing which would prevent acosh etc. from loading, since the dialects are directly imported from mlir:

https://github.com/tracel-ai/tracel-llvm/blob/9d51e26fab8542949c94f7c14c2519113648bf27/crates/tracel-mlir-sys/wrapper.h#L19

and

https://github.com/tracel-ai/tracel-llvm/blob/9d51e26fab8542949c94f7c14c2519113648bf27/crates/tracel-mlir-rs/src/pass/conversion.rs#L35

and

https://github.com/tracel-ai/tracel-llvm/blob/9d51e26fab8542949c94f7c14c2519113648bf27/crates/tracel-mlir-rs/src/dialect/ods.rs#L144

Maybe some files are missing in https://github.com/tracel-ai/tracel-llvm/blob/9d51e26fab8542949c94f7c14c2519113648bf27/crates/tracel-mlir-rs/src/dialect/ods.rs#L146 ?

relativityhd: I have no clue how I could generate these files or where they are generated ...

vaijira:
Those are the files for the operands, but you need to include too the new
conversion i mentioned earlier so the operands could be converted to libm
calls otherwise it seems the dialect is unable to be converted. With
CUBECL_MLIR_DEBUG env variable targetting a directory you could actually
see the mlir files generated at each step conversion, for example rsqrt
ends as llvm 1/sqrt.

crates/cubecl-cpu/src/compiler/visitor/operation/arithmetic.rs
relativityhd: Sorry, I haven't seen your earlier comment about `math_to_libm`.

So I tested it and it now I get this error instead:
`cannot be converted to LLVM IR: missing `LLVMTranslationDialectInterface` registration for dialect for op: func.func`

I really have no clue where func.func is coming from. I only use the math_ods::... stuff

relativityhd:
I have added you all to my repo in case I am blocking your development, just in case

relativityhd:
I have merged from main, but still no success. Still got the `error: cannot be converted to LLVM IR: missing `LLVMTranslationDialectInterface` registration for dialect for op: func.func` when running `cargo xtask validate`...

I've tried running with the debug flag you mentioned `CUBECL_MLIR_DEBUG="./debug-mlir/" cargo xtask validate` but I can't see a new directory anywhere...

vaijira:
You have to activate feature mlir-dump, for example:

CUBECL_DEBUG_MLIR=/tmp/mlir2 cargo test --features mlir-dump -p cubecl-cpu  tests::f32_ty::unary:: -- --nocapture

The order of the registration of the conversion of the dialects is important, the above commit should fix it.

relativityhd:
OMG yes it finally works, thank you!

So I guess this PR is now ready for merge! @wingertge @nathanielsimard

crates/cubecl-cpp/src/shared/dialect.rs
```diff
@@ -633,6 +633,10 @@ pub trait DialectInstructions<D: Dialect> {
         }
     }
 
+    fn compile_instruction_atan2(f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
```
wingertge: This seems to be identical to metal, so there's no reason to have it as a dialect function

crates/cubecl-spirv/src/extensions.rs
```diff
@@ -72,10 +84,58 @@ pub mod glcompute {
             b.gl_cos_id(ty, Some(out), input).unwrap();
         }
 
+        fn tan(b: &mut SpirvCompiler<T>, ty: Word, input: Word, out: Word) {
+            b.cl_tan_id(ty, Some(out), input).unwrap();
```
wingertge: These are all from the wrong dialect,  `cl_` are the OpenCL functions. It should be `gl_tan_id` etc.

relativityhd: Good catch!

crates/cubecl-cpp/src/shared/dialect.rs
relativityhd: Removed it 👍🏻

relativityhd:
Sooo, tests failing on my machine, but the functions which are failing are outside this PR (plane-shuffle function). From my end everything _should_ be fine. 👍🏻

crates/cubecl-cpp/src/shared/unary.rs
```diff
@@ -150,10 +150,20 @@ macro_rules! function {
 
 function!(Log, "log");
 function!(FastLog, "__logf", false);
-function!(Cos, "cos");
-function!(FastCos, "__cosf", false);
 function!(Sin, "sin");
+function!(Cos, "cos");
+function!(Tan, "tan");
```
wingertge: This fails on CUDA because `htan` is undefined, I think this should be `false` for the f16 support

crates/cubecl-cpp/src/shared/unary.rs
```diff
@@ -202,6 +212,38 @@ impl<D: Dialect> Unary<D> for Tanh {
...
 
+pub struct Degrees;
+
+impl<D: Dialect> Unary<D> for Degrees {
+    fn format_scalar<Input: Component<D>>(
+        f: &mut std::fmt::Formatter<'_>,
+        input: Input,
+        _out_elem: Elem<D>,
+    ) -> std::fmt::Result {
+        write!(f, "{input}*57.29577951308232f")
```
wingertge: This is broken, you can't multiply a float by half. It should convert to the correct type. Same with radians.

antimora:
@nathanielsimard when is the plan to add these to Burn?

PR #864, opened 2025-09-08
Cuda device change optimization

nathanielsimard:
Added the possibility for runtimes to optimize data transfer between devices.

### DIFF
--- a/crates/cubecl-runtime/src/channel/base.rs
+++ b/crates/cubecl-runtime/src/channel/base.rs
@@ -28,10 +28,10 @@ pub trait ComputeChannel<Server: ComputeServer>: Clone + core::fmt::Debug + Send
     /// Write bytes to each binding
     fn write(&self, descriptors: Vec<(CopyDescriptor<'_>, &[u8])>) -> Result<(), IoError>;
 
-    /// Send data to another server. Returns when recv_from_peer has been called and the transfer has been registered.
+    /// Send data to another server.
     fn data_transfer_send(&self, id: DataTransferId, src: CopyDescriptor<'_>);
 
-    /// Receive data from another server. Returns when send_to_peer has been called and the transfer has been registered.
+    /// Receive data from another server. Returns when the transfer has been registered.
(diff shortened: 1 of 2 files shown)


crates/cubecl-runtime/src/channel/base.rs
```diff
@@ -27,6 +28,12 @@ pub trait ComputeChannel<Server: ComputeServer>: Clone + core::fmt::Debug + Send
     /// Write bytes to each binding
     fn write(&self, descriptors: Vec<(CopyDescriptor<'_>, &[u8])>) -> Result<(), IoError>;
 
+    /// Send data to another server. Returns when recv_from_peer has been called and the transfer has been registered.
```
louisfd: Method names in docs are outdated

crates/cubecl-wgpu/src/device.rs
```diff
@@ -52,3 +54,86 @@ impl Default for WgpuDevice {
...
+                        wgpu::DeviceType::Other => 4,
+                        wgpu::DeviceType::IntegratedGpu => 1,
+                        wgpu::DeviceType::DiscreteGpu => 0,
+                        wgpu::DeviceType::VirtualGpu => 2,
+                        wgpu::DeviceType::Cpu => 3,
+                    };
+
+                    // When best available we return everything.
+                    let is_default = type_id == 4;
+                    adapter_type_id == type_id || is_default
```
louisfd: This is quite confusing. Could be refactored as:

```rust
// When best available we return everything.
if type_id == 4 {
    return true;
}

// Otherwise, only accept matching device types
match adapter.get_info().device_type {
    wgpu::DeviceType::DiscreteGpu => type_id == 0,
    wgpu::DeviceType::IntegratedGpu => type_id == 1,
    wgpu::DeviceType::VirtualGpu => type_id == 2,
    wgpu::DeviceType::Cpu => type_id == 3,
    wgpu::DeviceType::Other => type_id == 4,
}


```

louisfd (approved):
LGTM, two small comments only

PR #867, opened 2025-09-09
fix: Fix powf_scalar for wgsl

wingertge:
Ensures rhs is cast to the scalar version of lhs, rather than the vector version


wingertge:
Obsoleted by https://github.com/tracel-ai/cubecl/pull/838

PR #868, opened 2025-09-09
feat: Split traits

wingertge:
Extend macro to support split traits (with an expand for methods), and use that new functionality to clean up some implementations.
New macro options `expand_base_traits` to set base traits for the expand trait (for split traits), and `receiver_type` to allow taking expanded methods by reference. It's a bit awkward, but only needed for dynamic dispatch so it'll rarely be needed outside the standard library.
Line size is now split into a new trait because it doesn't follow the typical conventions and would make it impossible to generate the `List`, `VirtualTensorOperations` and `VirtualViewOperations` via the macro.


crates/cubecl-core/src/frontend/container/slice/operator.rs
```diff
@@ -223,96 +219,40 @@ impl<E: CubePrimitive> SliceMutOperatorExpand<E> for SliceExpand<E, ReadWrite> {
     }
 }
 
+#[cube(receiver_type = "ref")]
```
nathanielsimard: what is `receiver_type`

wingertge: it's the type for the expanded `self` parameter, receiver is the name used by syn and in official docs, but maybe `self_type` would be a better name. I think I'll rename it.

wingertge:
> Before merging I would create a page in the cubecl book documenting the `#[cube]` macro for `trait`. Or maybe we can even add the documentation in the code and explain the parameters.

A page doesn't sound bad, I'll try to explain it as best as I can. There's a lot of magic going on to make things work so it's a bit confusing, but I'll try my best.

crates/cubecl-core/src/frontend/container/slice/operator.rs
wingertge: It's worth noting this shouldn't need to be set in most normal cases, it's only relevant where you need dynamic dispatch (which is only in complex framework stuff like views and virtual tensors for fusion).

wingertge:
I've added the docs and also while writing them realized that Rust allows duplicate supertraits, so the macro really should add the `CubeType<ExpandType: ExpandTrait>` bound. I left it out to allow specifying additional bounds, but you can just add a second `CubeType` supertrait to do that. It further cleaned things up.

PR #877, opened 2025-09-10
core: Safer tensor handles + checked vectorization

ariawisp:
This PR makes constructing tensor handles and choosing a vectorization factor safer and more ergonomic—especially for host wrappers and FFI—by surfacing misuse as clear, early errors instead of relying on scattered `unsafe` and assumptions.

Behavior
- Safe constructors validate shape/stride basics; a checked conversion verifies the factor is supported by the runtime.
- The checked path intentionally does not enforce inner‑most contiguity or divisibility: kernels may vectorize on other axes or handle pitched/tail cases internally.
- Existing `unsafe` entry points remain available; no changes to kernel or runtime ABIs.

Impact
- Reduces boilerplate and the likelihood of undefined behavior in host integrations.
- Provides consistent, actionable errors when a vectorization factor is not supported.

Validation
- CubeCL workspace validates: audit, format, clippy, unit tests, and docs all pass.

Notes
- Includes a tiny macOS‑only clippy cleanup (WGSL SafeTanh) to satisfy workspace `-D warnings`.

PR has been validated with Burn — no compilation or test errors.

### DIFF
--- a/crates/cubecl-core/src/frontend/container/tensor/launch.rs
+++ b/crates/cubecl-core/src/frontend/container/tensor/launch.rs
@@ -13,6 +13,36 @@ use crate::{
 
 use super::Tensor;
 
+/// Errors that can occur when constructing a tensor handle safely.
+#[non_exhaustive]
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum TensorHandleError {
+    /// Rank of shape and strides differ.
+    RankMismatch {
+        shape_rank: usize,
+        stride_rank: usize,
+    },
+    /// Element size must be > 0.
+    ElemSizeZero,
+    /// A stride is zero for a dimension with extent > 1.
+    ZeroStride { axis: usize },
+}
+
+/// Errors that can occur when converting a handle to a runtime tensor argument.
+#[non_exhaustive]
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum TensorArgError {
+    /// Requested vectorization factor is not supported by the runtime.
+    UnsupportedVectorization {
+        requested: u8,
+        supported: &'static [u8],
+    },
+    /// Inner-most dimension is not contiguous (stride != 1) while vectorization > 1.
+    NonContiguousInner,
+    /// Inner-most dimension is not divisible by the vectorization factor.
+    MisalignedVectorization { last_dim: usize, factor: u8 },
+}
+
 /// Argument to be used for [tensors](Tensor) passed as arguments to kernels.
 #[derive(Debug)]
 pub enum TensorArg<'a, R: Runtime> {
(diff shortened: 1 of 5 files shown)


crates/cubecl-core/src/frontend/container/tensor/launch.rs
```diff
@@ -178,17 +205,33 @@ impl<R: Runtime> ArgSettings<R> for TensorArg<'_, R> {
...
+    /// legally vectorize along axes other than the innermost.
+    pub fn try_as_tensor_arg(
+        &'a self,
+        vectorization: u8,
+    ) -> Result<TensorArg<'a, R>, TensorArgError> {
+        if !R::supported_line_sizes().contains(&vectorization) {
+            return Err(TensorArgError::UnsupportedVectorization { requested: vectorization, supported: R::supported_line_sizes() });
+        }
+        Ok(self.as_tensor_arg(vectorization))
+    }
```
nathanielsimard: The problem I see here is that the validation is done in 100% of cases in Burn and CubeCL when we choose the line size, not when applying it! It's kind of wasteful to do it multiple times, especially since we iterate over a list each time. I don't mind too much having a try function, but it would not be good practice to use them; validation should be done before creating the tensor argument.

crates/cubecl-core/src/frontend/container/tensor/launch.rs
```diff
@@ -209,4 +252,67 @@ impl<'a, R: Runtime> TensorHandleRef<'a, R> {
...
+        }
+        if elem_size == 0 {
+            return Err(TensorHandleError::ElemSizeZero);
+        }
+        // Disallow zero strides when corresponding dimension extent > 1 (broadcasted dims with extent 1 are allowed).
+        for (i, (&s, &d)) in strides.iter().zip(shape.iter()).enumerate() {
+            if s == 0 && d > 1 {
+                return Err(TensorHandleError::ZeroStride { axis: i });
+            }
+        }
```
nathanielsimard, line 317: The same thing here: most of those things are validated in other places, and it's kind of wasteful to do those validations multiple times.

wingertge:
I actually think it could be a good idea, but specifically gated behind `cfg!(debug_assertions)`. These checks are not that expensive (you're only iterating over a handful of elements here), and would be useful to just assert for sanity checks while running unoptimized builds.

ariawisp:
Per feedback, I reverted internal hot paths to the fast path (`as_tensor_arg`) and moved validation to selection time. I added `debug_assert!` guards in the constructors so debug builds catch mistakes with zero release overhead. The `try_*` APIs are kept only for external entry points—e.g., language bindings, tools/CLI, or test harnesses—where we construct handles from raw parts and prefer Result-based errors over relying on unsafe preconditions. Docs now call this out explicitly. Workspace build and clippy are green. If you’d like, I can also gate the `try_*` APIs behind a feature flag, but this seems like the right safety/perf balance.

ariawisp:
Branch has been rebased on main (https://github.com/tracel-ai/cubecl/commit/cf40b4e31156064aee3271093d4cdd3a57bf3993) and confirmed CI workflow passes on my fork.

antimora:
Stale

PR #881, opened 2025-09-11
Feat/bytes

nathanielsimard:
Big refactor of the `Bytes` structure in Burn, now in `cubecl-common`.

Burn PR: https://github.com/tracel-ai/burn/pull/3701


crates/cubecl-common/src/bytes/buffer.rs
```diff
@@ -0,0 +1,125 @@
+use crate::bytes::default_controller::{self, DefaultAllocationController};
+use crate::bytes::{Allocation, AllocationController};
+use core::alloc::Layout;
```
laggui, line 5: Ahhh we need to also use `alloc::vec::Vec` and `alloc::boxed::Box` for no-std

PR #884, opened 2025-09-12
fix: Fix casting for packed values

wingertge:
Fixes casting of packed values (currently only `e2m1`). Line size is adjusted for both input and output type packing.


crates/cubecl-cpp/src/shared/element.rs
```diff
@@ -146,6 +146,13 @@ impl<D: Dialect> Elem<D> {
         }
     }
 
+    pub const fn packing_factor(&self) -> usize {
```
nathanielsimard, line 150: I would put some documentations here. We should start add more docs for our compilers and runtimes.

I'm refactoring our runtimes and trying to add more docs and cleanup some implementations.

wingertge: Done 👍

nathanielsimard (approved):

PR #885, opened 2025-09-12
Pinned Memory

nathanielsimard:
Add the possibility to use pinned memory when reading data back from the GPU.

### DIFF
--- a/crates/cubecl-cuda/src/runtime.rs
+++ b/crates/cubecl-cuda/src/runtime.rs
@@ -12,10 +12,10 @@ use crate::{
 };
 use cubecl_common::profile::TimingMethod;
 use cubecl_core::{
-    AtomicFeature, CubeCount, CubeDim, Feature, MemoryConfiguration, Runtime, TmaFeature,
+    CubeCount, CubeDim, MemoryConfiguration, Runtime,
     ir::{
-        ElemType, FloatKind, IntKind, MatrixLayout, MmaProperties, StorageType, TargetProperties,
-        UIntKind,
+        ElemType, FloatKind, MatrixLayout, MmaProperties, SemanticType, StorageType,
+        TargetProperties,
     },
 };
 use cubecl_cpp::{
@@ -28,7 +28,7 @@ use cubecl_cpp::{
     },
 };
 use cubecl_runtime::{
-    ComputeRuntime, DeviceProperties,
+    ComputeRuntime, DeviceProperties, Plane, Tma, TypeUsage,
     channel::MutexComputeChannel,
     client::ComputeClient,
     memory_management::{HardwareProperties, MemoryDeviceProperties, MemoryManagement},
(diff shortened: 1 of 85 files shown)


crates/cubecl-wgpu/src/compute/storage.rs
```diff
@@ -22,9 +22,15 @@ impl core::fmt::Debug for WgpuStorage {
 #[derive(new, Debug)]
 pub struct WgpuResource {
     /// The wgpu buffer.
-    buffer: wgpu::Buffer,
-    offset: u64,
-    size: u64,
+    pub buffer: wgpu::Buffer,
+    /// The size the handle is pointing to in memory.
```
wingertge: That comment seems incorrect unless I don't fully understand it

crates/cubecl-wgpu/src/compute/stream.rs
```diff
@@ -188,30 +200,33 @@ impl WgpuStream {
...
-            });
+            let aligned_len = resource.size.div_ceil(align) * align;
+            let (staging, binding) = self.mem_manage.reserve_staging(aligned_len).unwrap();
+
+            // let staging_buffer = self.device.create_buffer(&wgpu::BufferDescriptor {
+            //     label: None,
+            //     size: aligned_len,
+            //     usage: wgpu::BufferUsages::MAP_READ | wgpu::BufferUsages::COPY_DST,
+            //     mapped_at_creation: false,
+            // });
```
wingertge: Dead code

crates/cubecl-wgpu/src/compute/controller.rs
```diff
@@ -0,0 +1,51 @@
+use cubecl_common::bytes::{Allocation, AllocationController};
+use cubecl_runtime::memory_management::SliceBinding;
+use std::ptr::NonNull;
+
+/// Controller for managing wgpu staging buffers managed by a memory pool.
+pub struct WgpuAllocController {
```
wingertge, line 6: "Controller" is one of those meaningless words that are almost never a good choice to use. I'd suggest `WgpuAllocator` (and `Allocator` for the trait, I didn't review the PR where it was introduced). Rust has module scopes, so no need to worry about colliding with the one from `alloc`.

nathanielsimard: The thing is, it's not an allocator, it doesn't allocate, only deallocate or detach the allocation.

wingertge: It inherits a lot of the methods from `alloc::Allocator` so I'd argue it's still an allocator, just one with a more limited interface. removing the `zeroed` variations, `alloc::Allocator` adds `allocate`, `deallocate`, `grow` and `shrink`. This implements half of those, and could implement 3/4 if `shrink` was ever needed.

wingertge: And `grow` is certainly an allocation.

crates/cubecl-cuda/src/runtime.rs
```diff
@@ -155,8 +162,22 @@ fn create_client<M: DialectWmmaCompiler<CudaDialect<M>>>(
...
+    // expect the CPU to have at least the same amount of RAM as GPU memory.
+    let memory_management_cpu = MemoryManagement::from_configuration(
+        PinnedMemoryStorage::new(),
+        &MemoryDeviceProperties {
+            max_page_size: mem_properties.max_page_size,
+            alignment: PINNED_MEMORY_ALIGNMENT as u64,
+            data_transfer_async: false,
+        },
+        options.memory_config,
+    );
```
wingertge, line 180: That's not necessarily a good assumption, as evidenced by my 5090 with only 32GB of total system RAM 😅 
That being said, it should only grow as big as needed for the transfer, correct? If so it might be fine.

wingertge (approved):
The pinned memory stuff looks good

crates/cubecl-cuda/src/runtime.rs
nathanielsimard: Yup that's right!

crates/cubecl-wgpu/src/compute/controller.rs
nathanielsimard: I'm going to keep "controller" for now. I prefer an ugly name to a pretty name that might be slightly wrong. There's a higher chance we’ll update the ugly name to something more appropriate sooner.

PR #888, opened 2025-09-15
refactor: Rename matmul components for clarity

wingertge:
Renames matmul components for clarity. Enforces a `Reader -> Loader -> Reader` pattern with prefixes to clarify where in the chain each component is. The full chain is now:
`TensorReader -> StageLoader -> StageReader -> TileLoader`
Renames all `fill` functions that aren't actually filling with `load`. Renames the stage and global level `Accumulator` to `Accumulators` to better represent what it actually is (a sequence of accumulators rather than a single tile accumulator).

`GlobalWriter` currently clashes with this naming scheme, what could be a better name? It should be the inverse of a `StageLoader` that writes to a `TensorWriter`. `StageUnloader`? Or is there a better name? There will also be a tile equivalent in a future PR.


louisfd (approved):
LGTM

PR #894, opened 2025-09-18
fix: Align shift right behaviour to align with most other compilers

wingertge:
Changes shift behaviour to use arithmetic shift for signed integers, to align with the default behaviour in most languages.
It's actually implementation defined in C++, but it seems like all GPU targets use the de-facto standard, with logical left shifts and arithmetic right shifts for signed integers. Left shift on signed integers is UB and will always use logical shifts.


nathanielsimard (commented):
We should probably remove the clippy `is_multiple_of` since that can cause problem with cubecl kernels.

wingertge:
> We should probably remove the clippy `is_multiple_of` since that can cause problem with cubecl kernels.

I replaced all occurrances outside cube code and added the ignore for cube functions

nathanielsimard (approved):

PR #896, opened 2025-09-18
Feat/multi streams 2

nathanielsimard:
Big refactoring of CUDA and HIP runtimes to support multiple concurrent streams with automatic synchronization. They are deactivated right now because there are still some bugs, and I might change the multi-stream strategy, but it's still a major cleanup, so it's probably better to merge sooner rather than later.


crates/cubecl-runtime/src/stream/base.rs
```diff
@@ -0,0 +1,462 @@
...
+        &mut self,
+        stream_id: StreamId,
+        analysis: SharedBindingAnalysis,
+    ) -> SharedBindingAnalysis {
+        if analysis.slices.is_empty() {
+            return analysis;
+        }
+
+        log::info!("Analysis for stream {stream_id} => {analysis:?}");
+        // println!("Analysis for stream {stream_id} => {analysis:?}");
```
louisfd: dead code

PR #898, opened 2025-09-19
feat: Saturating add/sub

wingertge:
Includes fixes from https://github.com/tracel-ai/cubecl/pull/893 because they're required for tests to run

Adds saturating addition and subtraction for all backends, using intrinsics for metal (which compile to hardware instructions according to the docs) and CUDA for i32, while falling back to a polyfill for all other targets/types.
The unsigned fallback is based on the CUDA PTX output for optimized GPU saturation. The signed version is based on the branchless implementation from this article:
<https://locklessinc.com/articles/sat_arithmetic/>
Modified to make it generic over bit width. Unliked the unsigned version, CUDA does not detect this polyfill as saturating addition, so I'm not aware if there's a better version for GPU.

# Testing
Adds new test for both signed and unsigned saturating arithmetic, with values adjusted for the type to ensure saturation cases are all tested.


wingertge:
> Code looks good, we should really disable multiple_of

I think it's a lot more readable, so I think I'd limit the disabling to only where it's needed for cube functions

wingertge:
We might even want to just add a quick implementation of only the expand version for people using it in the future, same with `div_ceil`. Just adding the method only, on any `ExpandElementTyped<T: Int>`

wingertge:
I added a compat implementation for `div_ceil` and `is_multiple_of`, I think it's much cleaner this way.

nathanielsimard (approved):

PR #913, opened 2025-09-23
refactor: Remove stage reader

wingertge:
* Removes the `StageReader` because it was practically redundant with `StageMemory`.
* Renames all `StageReader` associated types and traits to `Stage`, so the tile matmul now only gets the type of the stage (i.e. strided smem, filled), instead of the "reader" which was always an implementation detail.
* Refactors `StageMemoryConfig` to be more like `GlobalMemoryConfig`, to get rid of all the generics everywhere
* Renames the `TileReader` types to the now unused `StageReader`, because that's more consistent with `GlobalReader`. `GlobalReader` reads a tile from global memory into stages, `StageReader` reads a tile from the stage into fragments.

# Views
I did not end up using views for representing the stage SMEM, because it would require significant rewrites to the tile matmuls. Saving that for some future time.

# Tests

Two more tests for attention are now failing. But since matmul tests are fine and there are already 10/18 tests failing on main so the two new failures may have just been relying on undefined behaviour. I can't find any mistake in this PR for attention specifically.

### Instructions


crates/cubecl-attention/src/components/stage/dummy/attention.rs
```diff
@@ -24,13 +22,13 @@ pub struct DummyStageAttention<AP: AttentionPrecision, R, TA: TileAttention<AP>>
...
 #[cube]
-impl<AP: AttentionPrecision, R: StageReader<AP::ES, TileKind = Strided>, TA: TileAttention<AP>>
+impl<AP: AttentionPrecision, R: Stage<AP::ES, TileKind = Strided>, TA: TileAttention<AP>>
     StageAttention<AP> for DummyStageAttention<AP, R, TA>
 {
     type Config = DummyStageConfig<TA::Config>;
 
-    type KeyReader = R;
-    type ValueReader = R;
+    type KeyStage = R;
```
louisfd: I guess R is no longer the right letter

wingertge: Done 👍

PR #914, opened 2025-09-23
Feat/wgpu/multi stream

nathanielsimard:
This PR introduces a new multi-streaming approach for a single GPU queue, scheduling tasks from different streams in a specific order, independent(ish) of multi-threaded enqueue logic.


crates/cubecl-wgpu/src/compute/stream.rs
```diff
@@ -325,21 +267,12 @@ impl WgpuStream {
         self.mem_manage.reserve(size, stream_id)
     }
 
-    fn create_uniform(&mut self, data: &[u8]) -> WgpuResource {
+    pub fn create_uniform(&mut self, data: &[u8]) -> WgpuResource {
```
ArthurBrussee: Why pub? there's still some tricky stuff to keep in mind wrt. ordering of the write and executions, so best not to use this WgpuResource anywhere if not needd.

crates/cubecl-runtime/src/stream/base.rs
```diff
@@ -1,458 +1,97 @@
...
 
-    fn get_mut(&mut self, stream_id: &StreamId) -> &mut StreamWrapper<B> {
+    /// Retrieves a mutable reference to a stream for a given stream ID.
+    pub fn get_mut(&mut self, stream_id: &StreamId) -> &mut F::Stream {
+        // Calculate the index for the stream ID.
         let index = self.stream_index(stream_id);
 
-        // The pool is init, can't go over the index because of stream_index.
+        // Use unsafe method to retrieve the stream, assuming the index is valid.
+        // Safety: The `stream_index` function ensures the index is within bounds.
```
ArthurBrussee: But it doesn't ensure there is only mut reference - calling this twice is UB, so this function itself is unsafe.

crates/cubecl-runtime/src/stream/base.rs
```diff
@@ -1,458 +1,97 @@
...
         unsafe { self.get_mut_index(index) }
     }
 
-    unsafe fn get_mut_index(&mut self, index: usize) -> &mut StreamWrapper<B> {
+    /// Retrieves a mutable reference to a stream at the specified index, initializing it if needed.
+    ///
+    /// # Safety
+    ///
+    /// Caller must ensure the index is valid (less than `max_streams + num_special`).
```
ArthurBrussee: And ensure only one mut reference exists. Two &mut is immediate UB. I'm not even sure having two &mut to _different indices_ is allowed as they come from the same provenance. I'm really really dubious about introducing unsafe for a small performance gain.

crates/cubecl-runtime/src/stream/base.rs
```diff
@@ -1,458 +1,97 @@
...
-    /// for synchronization, flushes them, and waits on the events in the target stream.
-    fn align_streams<'a>(
-        &mut self,
-        stream_id: StreamId,
-        bindings: impl Iterator<Item = &'a Binding>,
-    ) -> SharedBindingAnalysis {
-        let analysis = self.update_shared_bindings(stream_id, bindings);
-
-        self.apply_analysis(stream_id, analysis)
+    /// Caller must ensure the index corresponds to a valid special stream.
```
ArthurBrussee: Again also calling this twice is UB

crates/cubecl-runtime/src/stream/event.rs
```diff
@@ -0,0 +1,420 @@
...
+struct GcThread<B: EventStreamBackend> {
+    sender: SyncSender<GcTask<B>>,
+}
+
+impl<B: EventStreamBackend> GcThread<B> {
+    fn new() -> GcThread<B> {
+        let (sender, recv) = std::sync::mpsc::sync_channel::<GcTask<B>>(8);
+
+        std::thread::spawn(move || {
+            while let Ok(event) = recv.recv() {
```
ArthurBrussee, line 106: This won't work on WASM. Could work as an async task, but, in general making the memory behaviour non-deterministic and depend on scheduling seems really not great, it was a nice feature of Burn to know memory could be re-used immediatly after execution.

crates/cubecl-runtime/src/stream/base.rs
nathanielsimard: Can't call the function twice because the code still has to follow Rust lifetime rules. Added docs for that

crates/cubecl-runtime/src/stream/event.rs
nathanielsimard: This part is only for native backends (hip, cuda) I added a feature flag to make it clearer.

PR #924, opened 2025-09-26
refactor: Stage and global writers

wingertge:
Adds a `StageWriter` to mirror the `StageReader` (currently hardcoded to one tile kind, but it gives the framework to support things like swizzle in the future).
Also decouples the `GlobalWriter` from the Tile matul and moves it up to the global level, with the tile matmul instead working on `Stage`s (just like it does with inputs). The writers are currently still tightly coupled to the partitioning, but they no longer have to be. This involves moving most of the `StagePartitioner` logic up to the writer, only keeping `coordinates` because it's needed for scehduling.
Attention now has its own writer, since it uses a different partitioning strategy. The write itself is still delegated to the `PlaneWriter`, but the stage is created differently.

### Instructions


crates/cubecl-matmul/src/components/global/write/plane.rs
```diff
@@ -1,71 +1,141 @@
...
+        Self::new::<S>(tensor, config, stage_config)
+    }
+
+    fn stage(this: &Self) -> Self::Stage {
+        this.stage
+    }
+}
+
+#[cube]
+pub fn plane_write<ES: Numeric, EG: Numeric>(
```
louisfd, line 106: It's not clear to me why this was removed from the writer into a standalone function. The different plane writers look very similar to me, just  working on different stage memory configs.

louisfd (approved):
LGTM. Only one question

crates/cubecl-matmul/src/components/global/write/plane.rs
wingertge: It needs to be standalone because it's reused by the attention writer (with a different stage config).

PR #925, opened 2025-09-26
Fix WGPU async readback, make pinned memory safer

ArthurBrussee:
Fix memory corruption in cubecl-wgpu readbacks, make interface safer, add tests, and fix a memory leak.

The main fix for wgpu is to make a BufferSlice that holds onto a Pin<Box<>> reference. As this adress is a stable adress on the heap, the BufferSlice can hold on to it. This is still very sketchy for sure....

While searching for this I also reworked the interface here quite a bit:

- Consumers don't deal with `Allocation` anymore as its interface was brittle, and it can be hard to track whether no ptr is exposed in other ways.
- The inner "buffer" object is merged with the AllocationController, which now "owns" the memory its responsible for instead of having it passed in as a seperate entity.
- Deallocation now just happens on drop of the controller instead of a manual call.
- try_into_vec used to forget the whole Bytes which included the Box to the alloc controller, meaning the box leaked. `miri` caught this.
- can_be_detached instead becomes try_detach so it can mark memory to not be deallocated. 
- You can now run miri like `cargo miri test -p cubecl-common --all-features`. It'd be nice to run this on more complicated code using the interfaces but `miri` can't deal with most CubeCL currently.
- DefaultAllocaitonController is now "CoreAllocationController" as it depends on using the core memory allocator.
- Explain why exposing &mut [MaybeUninit<u8>] is unsafe and annotate functions.
- Annotate some other unsafe blocks.
- Added some more tests.

Additionally, while doing this, I found that nothing stops you from making multiple bytes objects for the same resource seemingly. `wgpu` has runtime checks for this, but cubecl-cpu does not. I've at least marked the resource write() function as taking in &mut but this is not sufficient. hip, and CUDA also do not check this. Best I can tell means you can hit UB (eg. two &mut pointers or writing from one thread while reading from another etc.). This might not be a problem in practice as reading only hands out one bytes object, but, this isn't really tracked anywhere.

### DIFF
--- a/crates/cubecl-common/src/bytes/default_controller.rs
+++ b/crates/cubecl-common/src/bytes/default_controller.rs
@@ -64,7 +64,7 @@ impl<'a> Allocation<'a> {
     }
 }
 
-/// Allocation controller using the core rust [alloc crate](alloc).
+/// Allocation controller using the 'native' rust [alloc crate](alloc).
 ///
 /// SAFETY:
 ///  - If `layout.size() > 0`, `ptr` points to a valid allocation from the global allocator
@@ -72,11 +72,11 @@ impl<'a> Allocation<'a> {
 ///  - If `layout.size() == 0`, `ptr` is aligned to `layout.align()` and `len` is 0.
 ///    `ptr` is further suitable to be used as the argument for `Vec::from_raw_parts` see [buffer alloc]
 ///    for more details.
-pub(crate) struct CoreAllocationController<'a> {
+pub(crate) struct NativeAllocationController<'a> {
     allocation: Allocation<'a>,
 }
 
-impl<'a> CoreAllocationController<'a> {
+impl<'a> NativeAllocationController<'a> {
     pub(crate) fn alloc_with_data(data: &[u8], align: usize) -> Result<Self, LayoutError> {
         debug_assert!(
             align <= MAX_ALIGN,
@@ -148,7 +148,7 @@ impl<'a> CoreAllocationController<'a> {
     }
 }
 
-impl AllocationController for CoreAllocationController<'_> {
+impl AllocationController for NativeAllocationController<'_> {
     fn grow(&mut self, size: usize, align: usize) -> Result<(), AllocationError> {
         debug_assert!(
             align <= MAX_ALIGN,
@@ -210,7 +210,7 @@ impl AllocationController for CoreAllocationController<'_> {
     }
 }
 
-impl Drop for CoreAllocationController<'_> {
+impl Drop for NativeAllocationController<'_> {
     fn drop(&mut self) {
         let layout = unsafe {
             Layout::from_size_align_unchecked(self.allocation.size, self.allocation.align)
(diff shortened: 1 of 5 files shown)


crates/cubecl-common/src/bytes/default_controller.rs
```diff
@@ -1,59 +1,226 @@
...
+/// Allocation controller using the core rust [alloc crate](alloc).
 ///
 /// SAFETY:
 ///  - If `layout.size() > 0`, `ptr` points to a valid allocation from the global allocator
 ///    of the specified layout. The first `len` bytes are initialized.
 ///  - If `layout.size() == 0`, `ptr` is aligned to `layout.align()` and `len` is 0.
 ///    `ptr` is further suitable to be used as the argument for `Vec::from_raw_parts` see [buffer alloc]
 ///    for more details.
-pub struct DefaultAllocationController;
+pub(crate) struct CoreAllocationController<'a> {
```
nathanielsimard: Core seems like this is the base of all allocation controllers, or to be used as a core component. Maybe `NativeAllocationController` would be better?

crates/cubecl-cpu/src/compute/alloc_controller.rs
```diff
@@ -1,23 +1,54 @@
...
+use cubecl_runtime::{
+    memory_management::MemoryManagement,
+    storage::{BytesResource, BytesStorage},
+};
 
 pub struct CpuAllocController {
-    binding: Option<Binding>,
+    resource: BytesResource,
+    // Needed to keep the binding alive.
+    _binding: Option<Binding>,
```
nathanielsimard: Do we need to keep the option, it should always be Some?

crates/cubecl-runtime/src/storage/base.rs
```diff
@@ -111,4 +111,12 @@ impl<Resource: Send> BindingResource<Resource> {
     pub fn resource(&self) -> &Resource {
         &self.resource
     }
+
+    /// access the underlying resource. Note: The resource might be bigger
+    /// than just the original allocation for the binding. Only the part
+    /// for the original binding is guaranteed to remain, other parts
+    /// of the resource *will* be re-used.
```
nathanielsimard: Not sure I understand the comment

ArthurBrussee: I think it's saying that while you get this from a binding referring to a specific part of a resource (eg. a buffer), this refers to the whole resource, so take care. 

I've improved the comment a bit and removed the duplicate mut version. Only the dummy backend used it appearantly and it's a bit sketchy, so that backend now just uses the mem manager directly so we don't leak this into the API.

crates/cubecl-cpu/src/compute/alloc_controller.rs
ArthurBrussee: Nice done

crates/cubecl-common/src/bytes/default_controller.rs
ArthurBrussee: Sounds good, done

nathanielsimard (approved):

PR #927, opened 2025-09-27
refactor: Replace `to_source_shape` with inclusive clamped ranges.

wingertge:
Solves the issue of multi-dimensional exclusive ranges being vastly out of bounds in linear space, by performing `View` slicing with inclusive ranges and only converting to an exclusive range at the end. This allows removing `to_source_shape` on the `Layout` trait and simplifying implementations.


louisfd (approved):
LGTM

PR #931, opened 2025-09-28
feat: Reusable shared memory

wingertge:
Applies CUDA's way of handling shared memory (a single chunk sliced into individual objects) to all C++ targets, as well as SPIR-V.
Adds the ability to `free` shared memory, allowing the corresponding slice to be reused.

### Instructions


crates/cubecl-cpp/Cargo.toml
```diff
@@ -25,6 +25,7 @@ std = ["cubecl-runtime/std", "cubecl-common/std", "cubecl-core/std"]
 [dependencies]
 cubecl-common = { path = "../cubecl-common", version = "0.7.0", default-features = false }
 cubecl-core = { path = "../cubecl-core", version = "0.7.0", default-features = false }
+cubecl-opt = { path = "../cubecl-opt", version = "0.7.0", default-features = false }
```
nathanielsimard, line 29: I would add a comment telling what part of cubecl-opt is used in the cpp compiler.

crates/cubecl-matmul/src/components/global/single_stage/simple/matmul.rs
```diff
@@ -94,6 +94,10 @@ where
             rhs_reader.advance_view();
         }
 
+        sync_cube();
+        lhs_reader.free_stage();
+        rhs_reader.free_stage();
+
```
nathanielsimard, line 109: I would add a comment saying how the shared memory from the stage can be reused (the output I think).

crates/cubecl-opt/src/analyses/liveness.rs
```diff
@@ -104,3 +104,239 @@ fn calculate_block_sets(opt: &mut Optimizer, block: NodeIndex) -> BlockSets {
 
     BlockSets { generated, kill }
 }
+
+pub mod shared {
```
nathanielsimard, line 109: I would add more doc in this module overall

crates/cubecl-matmul/src/components/global/single_stage/simple/matmul.rs
wingertge: It was just a test bed and I just realized it actually no longer works as intended now that the output stage is initialized ahead of time by the writer 😅 I'll have to do a refactor to separate the writer from the execution I think to allow stages to be reused now. This PR is mainly about making sure nothing breaks with the core change to how shared memory is allocated, and then a separate PR can refactor writers to use more shared memory and reuse the input stages. I'll actually remove it for now because it no longer works anyways.

wingertge: Actually it might work because the analysis looks at usage, creating the shared memory shouldn't actually count as a usage. Let me test this further.

crates/cubecl-opt/src/analyses/liveness.rs
wingertge: I added a bunch, is it good now?

crates/cubecl-matmul/src/components/global/single_stage/simple/matmul.rs
wingertge: It actually does still work! Though I agree it's confusing and we might want to defer creating the stage anyways just to make it clearer. But not as much of a priority. I'll add a comment for now.

wingertge: Done 👍

nathanielsimard (approved):

PR #933, opened 2025-09-29
Fix/cuda/data service

nathanielsimard:
Deactivate Peer-to-Peer transfer for now on CUDA since it doesn't seem to work reliably for some reason and is not always supported. So, I added a new Serialized version for device-to-device communication, which goes through pinned memory to transfer data across devices. All the synchronization is done using lazy CUDA events, so even if it goes through the CPU, it doesn't block the GPU queue.


wingertge (approved):
Can't test it with just one GPU, but looks right at least (apart from the one failed lint obviously.

PR #934, opened 2025-09-29
Add shared state map

nathanielsimard:
WIP

### DIFF
--- a/crates/cubecl-common/src/map.rs
+++ b/crates/cubecl-common/src/map.rs
@@ -24,6 +24,15 @@ impl<V> SharedState<V> {
     }
 }
 
+impl<K, V> Default for SharedStateMap<K, V>
+where
+    K: core::hash::Hash + core::cmp::PartialEq + core::cmp::Eq,
+{
+    fn default() -> Self {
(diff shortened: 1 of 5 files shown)


crates/cubecl-common/src/map.rs
```diff
@@ -0,0 +1,102 @@
+use crate::stub::{Mutex, RwLock, RwLockReadGuard, RwLockWriteGuard};
+use alloc::sync::Arc;
+use hashbrown::HashMap;
+
+/// A thread-safe map that allows concurrent access to values using read-write locks.
+pub struct SharedStateMap<K, V> {
+    state: Mutex<Option<HashMap<K, Arc<RwLock<V>>>>>,
```
laggui, line 7: Couldn't we just drop the `Option` here and always initialize a default map?

laggui (approved):
LGTM! Just a minor comment

crates/cubecl-common/src/map.rs
nathanielsimard: Unfortunatly no, because we can't have a const constructor

PR #935, opened 2025-09-30
feat: Manual mma tile

wingertge:
Adds support and tests for the manual MMA tile matmul. Does not yet add it to the strategy, that really needs a redesign to allow different tile matmuls for each algorithm.

### DIFF
--- a/crates/cubecl-matmul/src/tests/layered/macros/plane_mma/tiling_scheme/tile.rs
+++ b/crates/cubecl-matmul/src/tests/layered/macros/plane_mma/tiling_scheme/tile.rs
@@ -4,20 +4,20 @@ macro_rules! testgen_matmul_mma_tile {
         use $crate::components::TileSize;
 
         // AMD f16/bf16
-        // #[cfg(not(target_os = "macos"))]
-        // mod t16x16x16 {
-        //     use super::*;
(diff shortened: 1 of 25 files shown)


crates/cubecl-matmul/src/tests/layered/macros/plane_mma/tiling_scheme/tile.rs
```diff
@@ -0,0 +1,46 @@
...
+        //     $crate::testgen_matmul_mma_partition!(
+        //         $algorithm,
+        //         $precision,
+        //         $tiling_scheme_builder.with_tile_size(TileSize {
+        //             m: 16,
+        //             n: 16,
+        //             k: 16
+        //         })
+        //     );
+        // }
```
nathanielsimard: Dead code?

nathanielsimard (commented):
@louisfd could review this

crates/cubecl-matmul/src/tests/layered/macros/plane_mma/tiling_scheme/tile.rs
wingertge: Good catch, forgot to uncomment that. I had it commented so I can set matmul tests to panic on skip and ensure they actually run. It's needed for testing AMD.

PR #937, opened 2025-09-30
Add `IsNan` and `IsInf` ops

laggui:
Added `IsNan` and `IsInf` ops, including WGSL via bit checks and MLIR (poylfill, similar bitcheck).

Fixes the current Burn implementation which was not accurate across backends depending on the platform.

### Instructions

### DIFF
--- a/crates/cubecl-ir/src/processing.rs
+++ b/crates/cubecl-ir/src/processing.rs
@@ -197,11 +197,8 @@ impl ScopeProcessing {
                         sanitize_constant_scalar_ref_var(&mut op.lhs, &op.rhs);
                         sanitize_constant_scalar_ref_var(&mut op.rhs, &op.lhs);
                     }
-                    Comparison::IsNan(op) => {
-                        sanitize_constant_scalar_ref_var(&mut op.input, &inst.out.unwrap());
-                    }
-                    Comparison::IsInf(op) => {
-                        sanitize_constant_scalar_ref_var(&mut op.input, &inst.out.unwrap());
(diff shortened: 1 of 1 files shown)


laggui:
CC @ArthurBrussee thanks for the hint towards the bit checking when we discussed this a couple weeks back. Hopefully this will fix your use case too.

crates/cubecl-ir/src/processing.rs
```diff
@@ -197,6 +197,12 @@ impl ScopeProcessing {
                         sanitize_constant_scalar_ref_var(&mut op.lhs, &op.rhs);
                         sanitize_constant_scalar_ref_var(&mut op.rhs, &op.lhs);
                     }
+                    Comparison::IsNan(op) => {
+                        sanitize_constant_scalar_ref_var(&mut op.input, &inst.out.unwrap());
+                    }
+                    Comparison::IsInf(op) => {
+                        sanitize_constant_scalar_ref_var(&mut op.input, &inst.out.unwrap());
+                    }
```
wingertge: This is incorrect, you'd be converting the input to `bool` with this code. It should just be an empty body.

wingertge (changes requested):

crates/cubecl-ir/src/processing.rs
laggui: Shoot you're right. Good catch!

I don't actually recall changing this, my brain must have autofilled this while chasing the missing pattern matches highlighted by the compiler 😅

PR #938, opened 2025-09-30
refactor: Refactor quantization to use layouts and multi-dimensional block sizes

wingertge:
Refactors quantization to support non-integer quantizations and non-float scales (fp8 is not counted as a float because it doesn't support float operations).
Refactors block size to have the same dimensionality as the underlying tensor
Refactors scale indexing using layouts


crates/cubecl-quant/src/layout/scales.rs
```diff
@@ -0,0 +1,242 @@
...
+    fn to_source_pos(&self, _pos: Self::Coordinates) -> Self::SourceCoordinates {
+        0u32.runtime()
+    }
+
+    fn shape(&self) -> Self::Coordinates {
+        self.tensor_len
+    }
+
+    fn is_in_bounds(&self, pos: Self::Coordinates) -> bool {
+        pos < self.tensor_len
```
nathanielsimard, line 86: I guess it's alway in bound no?

crates/cubecl-quant/src/layout/scales.rs
```diff
@@ -0,0 +1,242 @@
...
+
+#[derive(CubeType, CubeLaunch)]
+pub struct BlockScaledLayout {
+    tensor_shape: Sequence<FastDivmod>,
+    tensor_len: u32,
+    scales_strides: Sequence<u32>,
+    #[cube(comptime)]
+    block_size: Vec<usize>,
+    #[cube(comptime)]
+    scales_line_size: u8,
```
nathanielsimard, line 111: I would add docs on why Vec is chosen instead of sequence and on all the fields here.

crates/cubecl-quant/src/layout/scales.rs
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    fn to_source_pos_checked(&self, pos: Self::Coordinates) -> (Self::SourceCoordinates, bool) {
+        (self.to_source_pos(pos), self.is_in_bounds(pos))
+    }
+}
+
+#[cube]
+impl BlockScaledLayout {
+    pub fn is_block_start(&self, pos: u32) -> bool {
```
nathanielsimard, line 154: I would add a bit of docs on top of those functions, saying why they are useful

crates/cubecl-quant/src/layout/scales.rs
```diff
@@ -0,0 +1,242 @@
...
+            is_start &= offs_local.is_multiple_of(block_size_local);
+        }
+
+        is_start
+    }
+}
+
+#[allow(unused_variables)]
+#[cube]
+fn unwrap(v: u32) -> comptime_type!(u32) {
```
nathanielsimard, line 175: That could be: `as_comptime_unchecked`

nathanielsimard: Or just : `try_as_comptime` and `as_comptime`. We don't need to `unchecked` which often times is unsafe. That could be in `cubecl-std`

crates/cubecl-quant/src/scheme.rs
```diff
@@ -78,21 +80,27 @@ impl QuantScheme {
...
 /// Data type used to represent quantized values.
 #[derive(Clone, Copy, Debug, Hash, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
 pub enum QuantValue {
     /// 8-bit quantization with full range.
     Q8F,
+    /// 8-bit floating point, e5m2 format.
+    E5M2,
+    /// 8-bit floating point, e4m3 format.
+    E4M3,
     /// 4-bit quantization with full range.
```
nathanielsimard, line 100: There's something strange to have floating point here. They are native quantize types right?

crates/cubecl-quant/src/layout/scales.rs
wingertge: It's just for completeness, for unchecked kernels this is never called anyways.

crates/cubecl-quant/src/layout/scales.rs
wingertge: Figured it was obvious, since Sequence is for runtime values and this is a comptime value.

crates/cubecl-quant/src/scheme.rs
wingertge: Yes, they're quantized floating point formats (though technically the 8-bit ones can be used for MMA without scales, it just doesn't really have a use case).
They need to be separate types because they obviously have different semantics than integer formats. We'll need a different way to calculate the scales for them too eventually, but I didn't implement that yet.
I also didn't implement packed kernels for them because the packing really relies on integer semantics rn and I didn't feel like completely overhauling that.

crates/cubecl-quant/src/layout/scales.rs
wingertge: It's kinda hard because we don't have a good way to convert an expanded constant back to its original type without hardcoding the type (i.e. u32 here). I think I'd rather investigate adding that in a separate PR and then maybe unwrapping automatically for always-unrolled loops.

nathanielsimard (approved):

PR #939, opened 2025-10-02
Upgrade wgpu

ArthurBrussee:

### Instructions


ArthurBrussee:
This is currently blocked as after https://github.com/gfx-rs/wgpu/commit/b5385225abdcf6d92dd6056bb3615ef735ec0949, buffers binding size now needs to be 4 byte aligned, which is failing some tests with eg. f16 elements in Burn.

The reasoning is here https://github.com/gpuweb/gpuweb/pull/3477 thought it might be removed after all discussed here https://github.com/gpuweb/gpuweb/issues/5203.

For now we'd need to pad the binding sizes. This is mostly ok except that for checked kernels it breaks Vulkan robustuness, so would need manual checks there after all. @wingertge proposal is to JIT different kernels depending on whether any of the inputs have padding or not. Unchecked kernels aren't affected.

laggui:
Closing for v28 update in #1119

PR #940, opened 2025-10-02
refactor: Allow accessing the underlying buffer for `TensorMap`

wingertge:
Adds the ability to access the underlying buffer on tensor maps, to allow direct reading. This is achieved by just adding a buffer of the same ID as the tensor map to the kernel params.
Also changes the virtual tensor function to get the tensor map so it now returns a `CubeOption` instead of panicking (this didn't exist when I initially added the function).

### DIFF
--- a/crates/cubecl-core/src/frontend/container/tensor/tensormap.rs
+++ b/crates/cubecl-core/src/frontend/container/tensor/tensormap.rs
@@ -250,6 +250,7 @@ macro_rules! tma_store {
     };
 }
 
+tma_store!(1, x);
 tma_store!(2, y, x);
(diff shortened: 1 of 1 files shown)


crates/cubecl-cuda/src/compute/server.rs
```diff
@@ -241,6 +249,8 @@ impl ComputeServer for CudaServer {
                     .collect();
                 let elem_stride: Vec<_> = map.elem_stride.iter().rev().map(|s| *s as u32).collect();
 
+                println!("meta: {map:?}");
```
nathanielsimard: print

nathanielsimard (approved):

PR #954, opened 2025-10-12
feat: Quantized view

wingertge:
Adds a new kind of launchable view that represents a set of quantized values and scales, and dequantizes on read. The dequantization is entirely transparent to the kernel and happens automatically.
This is the implementation with some tests to ensure the dequantization and unrolling happens correctly. I'll put the integration with matmul in a separate PR.

# CubeCL-Quant changes
 
This PR moves the quantization scheme out of `cubecl-quant` and into `cubecl-common`, reexporting in `cubecl-quant`. This is required to be able to use it in `cubecl-std`, and the quantized view needs to live there because it's tightly coupled to the launch code. It also replaces the FP8 references with constants so it works without the corresponding crates.

### DIFF
--- a/crates/cubecl-attention/src/components/global/dummy/read.rs
+++ b/crates/cubecl-attention/src/components/global/dummy/read.rs
@@ -135,9 +135,11 @@ impl<AP: AttentionPrecision, G: GlobalAttentionConfig> DummyKeyReader<AP, G> {
                                 let index_load = row_load_in_tile * tile_cols_load + col_load;
                                 let index_store = col_load * tile_rows_load + row_load_in_tile;
 
-                                slice[index_store + store_offset] = Line::cast_from(
-                                    view.read_checked(((tile_row_load, tile_col_load), index_load)),
-                                );
+                                slice[index_store + store_offset] =
+                                    Line::cast_from(view.read_checked((
+                                        (tile_row_load, tile_col_load).runtime(),
+                                        index_load,
+                                    )));
                             }
                         }
                     }
@@ -210,7 +212,7 @@ impl<AP: AttentionPrecision, G: GlobalAttentionConfig> DummyValueReader<AP, G> {
                                 let index = row_in_tile * tile_cols + col;
 
                                 slice[index + offset] = Line::cast_from(
-                                    view.read_checked(((tile_row, tile_col), index)),
+                                    view.read_checked(((tile_row, tile_col).runtime(), index)),
                                 );
                             }
                         }
--- a/crates/cubecl-attention/src/components/stage/dummy/attention.rs
+++ b/crates/cubecl-attention/src/components/stage/dummy/attention.rs
@@ -56,19 +56,15 @@ impl<
 
         let p = config.tiling_scheme().partition_size;
 
-        let mut kv = comptime![0u32];
-
         let mut max_placeholder = TA::init_max_placeholder(config.num_rows_per_unit());
         let mut sum_placeholder = TA::init_sum_placeholder(config.num_rows_per_unit());
 
         #[unroll]
         #[allow(clippy::explicit_counter_loop)]
-        for _ in 0..p.seq_kv {
-            let mut hd = comptime![0u32];
-
+        for kv in 0..p.seq_kv {
             #[unroll]
             #[allow(clippy::explicit_counter_loop)]
-            for _ in 0..p.head_dim {
+            for hd in 0..p.head_dim {
                 let key_tile = SK::tile(key_reader, (hd, kv).runtime());
 
                 TA::fill_key(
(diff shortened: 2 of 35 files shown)


crates/cubecl-std/src/quant/dequantize.rs
```diff
@@ -0,0 +1,109 @@
+use cubecl::prelude::*;
+use cubecl_common::quant::scheme::*;
+use cubecl_common::{e2m1x2, e4m3, e5m2};
+use cubecl_core::{self as cubecl, intrinsic};
+
+/// Dequantize a line of values, where `line_size * num_quants` is a power of two.
+/// Unaligned values can't be dequantized in place.
+#[cube]
+pub fn dequantize_aligned<Q: CubePrimitive, S: CubePrimitive, F: Float>(
```
nathanielsimard, line 9: Is there a reason why those functions are in cubecl-std instead of cubecl-quant?

crates/cubecl-std/src/quant/view.rs
```diff
@@ -0,0 +1,306 @@
...
+    self as cubecl,
+    ir::{ElemType, FloatKind, StorageType},
+    prelude::barrier::BarrierExpand,
+    unexpanded,
+};
+use half::{bf16, f16};
+
+/// View that dequantizes after loads. Scales layout should take values coordinates and map them
+/// to the corresponding scale.
+/// Assumes only one scale maps to a single load. Adjust line size of values or block size to ensure
```
nathanielsimard, line 29: I would put that comment inside a `# Warning` section. I would also comment on the formula that needs to be respected like `block_size % line_size == 0`

crates/cubecl-std/src/quant/view.rs
```diff
@@ -0,0 +1,306 @@
...
+        let scale = self.scales.clone().__expand_read_method(scope, pos);
+
+        dequantize_aligned::expand::<Q, S, F>(scope, value, scale, self.scheme)
+    }
+
+    fn __expand_read_checked_method(
+        &self,
+        scope: &mut Scope,
+        pos: <C>::ExpandType,
+    ) -> ExpandElementTyped<Line<F>> {
```
nathanielsimard, line 132: We should find a wait to use the `cube` macro to generate these. Maybe with a tag? `cube(dyn_trait)` ?

crates/cubecl-std/src/quant/view.rs
```diff
@@ -0,0 +1,306 @@
...
+            scales,
+            scheme,
+            _ty: PhantomData,
+        };
+        ViewExpand::new(view)
+    }
+
+    #[allow(clippy::missing_transmute_annotations)]
+    unsafe {
+        match E::as_type(&builder.scope) {
```
nathanielsimard, line 289: The scope has an element type?

crates/cubecl-std/src/quant/view.rs
```diff
@@ -0,0 +1,306 @@
...
+            StorageType::Scalar(ElemType::Float(ty)) => match ty {
+                FloatKind::F16 => t(expand_dynamic_f::<f16, C>(values, scales, scheme, builder)),
```
nathanielsimard, line 291: Is the dequantization float value in the schema? If yes, then are we casting to the asked element type from the view?

crates/cubecl-std/src/quant/dequantize.rs
wingertge: `cubecl-quant` requires `cubecl-std` and it would end up being circular. And the view needs to be in `cubecl-std` because it needs tight integration with the launch code so it can work as a normal view.

crates/cubecl-std/src/quant/view.rs
wingertge: In this case there's only the expand so I'm not sure it's even worth it, since there's no duplication. But the main reason this needs to be expanded is that `C` may or may not be clone, but we know the expand is. So there's actually no way to properly implement an unexpanded version for arbitrary coordinates without stricter trait bounds.
There are some that I'd like to have default bodies for though.

crates/cubecl-std/src/quant/view.rs
wingertge: It's a bit awkward, the float value is available because that's in the view signature (i.e. `View<f32, Coords1d>`). The scale and value types are not in the signature, so we need to reconstruct them from the schema. The quantized view also requires the view type to be a float, but the signature for `View` only requires `CubePrimitive`, so to tighten the trait bounds it needs to match on the `ElemType` and panic if it's not a float. We can't put the bound on the `LaunchArg` impl because we still want to support unquantized views on non-floats obviously.

wingertge: We can't make the signature of the quantized view much looser because it requires at least `Mul` to be supported, which it isn't for `CubePrimitive` (not supported for `bool` and the conversion-only minifloats).

PR #959, opened 2025-10-16
Refactor: Device locking

nathanielsimard:
Big refactor to centralize locking strategies and remove a lot of custom buggy code.

The reason I worked on this was a deadlock caused by fusion. The problem is that the fusion server was locked, but was waiting on a recursive function that needed the fusion server. With this refactor, the fusion server will use the same lock as the cubecl server, which resolves the issue.

### DIFF
--- a/crates/cubecl-runtime/tests/dummy/compute.rs
+++ b/crates/cubecl-runtime/tests/dummy/compute.rs
@@ -1,45 +1,50 @@
 use std::sync::Arc;
 
 use super::DummyServer;
-use cubecl_common::CubeDim;
-use cubecl_common::profile::TimingMethod;
+use cubecl_common::device::{Device, DeviceState};
 use cubecl_runtime::client::ComputeClient;
 use cubecl_runtime::logging::ServerLogger;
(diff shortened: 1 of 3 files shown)


crates/cubecl-common/src/device.rs
```diff
@@ -43,3 +43,456 @@ impl PartialOrd for DeviceId {
...
+                ptr.as_mut().unwrap()
+            };
+
+            if !map.contains_key(&key) {
+                let state_default = S::init(self.device_id);
+                let any: Box<dyn Any + Send + 'static> = Box::new(state_default);
+                let cell = RefCell::new(any);
+
+                map.insert(key, cell);
+            }
```
laggui, line 340: Had to double-check, but this will realloc after another number of items are inserted and might reference a dangling pointer. Typically the capacity grows to 4 after the first item is inserted? So at the 5th item it will change.

We could simply init the map with a fixed capacity first to guarantee that it doesn't happen for a number of device states, e.g. `HashMap::with_capacity(MAX_STATE_TYPES)`.

PR #960, opened 2025-10-16
feat: Matmul global quant

wingertge:
Implements automatic dqeuantization on global load for matmul.

### Instructions


crates/cubecl-matmul/src/components/global/memory/layout.rs
```diff
@@ -102,6 +108,11 @@ impl Layout for BatchedGlobalLayout {
         let line_size = comptime![self.config.global_line_size];
         let (mut batch, row, col) = coords;
 
+        let (row, col) = match comptime![self.config.matrix_layout] {
+            MatrixLayout::RowMajor => (row, col / self.packing),
+            MatrixLayout::ColMajor => (row / self.packing, col),
+        };
```
nathanielsimard, line 114: Would it be useful to also perform a comptime if to remove the / 1 here?

crates/cubecl-matmul/src/components/global/memory/layout.rs
```diff
@@ -175,6 +186,157 @@ impl<'a, R: Runtime> BatchedGlobalLayoutLaunch<'a, R> {
...
+
+        let scales_layout = {
+            let shape = (ScalarArg::new(rows as u32), ScalarArg::new(cols as u32));
+
+            match scheme.level {
+                QuantLevel::Tensor => BatchedGlobalScaleLayoutArgs::PerTensor { shape },
+                QuantLevel::Block(block_size) => {
+                    let [block_row, block_col] = block_size.as_dim();
+                    let mut config = config;
+                    config.global_line_size = 1;
```
nathanielsimard: I would put a comment on why we override the global_line_size here? Scales are never vectorized right?

crates/cubecl-std/src/quant/view.rs
```diff
@@ -218,6 +218,55 @@ impl<Q: CubePrimitive, S: CubePrimitive, F: Float, C: Coordinates + 'static>
...
+
+    fn execute<Q: CubePrimitive, S: CubePrimitive>(self) -> Self::Output {
+        let values = View::<Line<Q>, C>::expand(self.values, self.builder);
+        let scales = View::<S, C>::expand(self.scales, self.builder);
+        let view = QuantizedViewExpand::new(values, scales, self.scheme);
+        ViewExpand::new(view)
+    }
+}
+
+pub fn run_with_quant_type<F: RunWithQuantType>(func: F, scheme: QuantScheme) -> F::Output {
```
nathanielsimard, line 242: This would benefit from more docs.

crates/cubecl-matmul/src/components/global/memory/layout.rs
wingertge: Yeah I added a comment and also changed it to the constructor syntax to make it more obvious we're making a copy but with line size 1.

crates/cubecl-matmul/src/components/global/memory/layout.rs
wingertge: Don't think so, it's one line in the generated code and it will always get optimized out. Not worth making the actual readable code less readable.

crates/cubecl-std/src/quant/view.rs
wingertge: Done. It's mainly for fusion since that also needs to dynamically instantiate the views with the right types. But could be useful in the future.

PR #964, opened 2025-10-17
Metal: fix plane ops

louisfd:
Fix plane operations for metal dialect

### DIFF
--- a/crates/cubecl-spirv/src/bitwise.rs
+++ b/crates/cubecl-spirv/src/bitwise.rs
@@ -5,7 +5,7 @@ use cubecl_core::{
 use cubecl_core::{comptime, ir as core, prelude::*};
 use cubecl_core::{cube, ir::Bitwise};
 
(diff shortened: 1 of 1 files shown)


crates/cubecl-cpp/src/metal/dialect.rs
```diff
@@ -33,94 +33,107 @@ impl Dialect for MslDialect {
...
+        if vectorization > 1 {
+            for k in 0..vectorization - 1 {
+                f.write_fmt(format_args!(
+                    "{simd_op_prefix}{input}.i_{k}{simd_op_suffix},\n"
+                ))?;
+            }
+            f.write_fmt(format_args!(
+                "{simd_op_prefix}{input}.i_{}{simd_op_suffix}\n",
+                vectorization - 1
+            ))?;
```
nathanielsimard: Can you merge both writes?

PR #967, opened 2025-10-19
fix: Packed matmul

wingertge:
Various fixes for quantized matmul with packed storage.

### Instructions

### DIFF
--- a/crates/cubecl-core/src/runtime_tests/plane.rs
+++ b/crates/cubecl-core/src/runtime_tests/plane.rs
@@ -672,9 +672,7 @@ pub fn test_plane_shuffle<
     let mut expected = input.clone();
 
     // All lanes read from lane 0 (same as broadcast(value, 0))
-    for v in 0..vectorization as usize {
-        expected[v] = input[v]; // Lane 0's value
-    }
+    expected[..vectorization as usize].copy_from_slice(&input[..vectorization as usize]);
 
     let input: Vec<F> = input.into_iter().map(|x| F::new(x)).collect();
     let expected: Vec<F> = expected.into_iter().map(|x| F::new(x)).collect();


crates/cubecl-matmul/src/base.rs
```diff
@@ -216,6 +262,49 @@ impl<'a, R: Runtime> MatmulInputHandleRef<'a, R> {
...
+            MatmulInputHandleRef::Quantized {
+                data,
+                scale,
+                shape,
+                scheme,
+            } => {
+                let data = match scheme.store {
+                    // e2m1 has native packing so also needs to be re-packed
+                    QuantStore::Native if scheme.value == QuantValue::E2M1 => {
+                        let data = into_contiguous_packed::<R, u8>(client, data, shape, 2);
```
nathanielsimard, line 283: Why 2? Is it always the case, if so we should add a comment.

crates/cubecl-matmul/src/kernels/naive.rs
```diff
@@ -18,11 +18,40 @@ use crate::{
...
+fn load_unrolled<I: Numeric>(
+    view: &View<Line<I>, Coords3d>,
+    pos: Coords3d,
+    #[comptime] layout: MatrixLayout,
+    #[comptime] line_size: u32,
+) -> Line<I> {
+    let view_line_size = view.line_size();
+    if comptime![view.line_size() == line_size] {
+        view[pos]
+    } else {
```
nathanielsimard, line 37: I would add a comptime assert here view_line_size < line_size.

crates/cubecl-std/src/tensor/contiguous.rs
```diff
@@ -149,6 +149,92 @@ fn into_contiguous_kernel_pack<N: CubePrimitive>(
     }
 }
 
+#[cube]
+fn index_packed<N: Int>(
```
nathanielsimard, line 155: I would add some docs here.

crates/cubecl-std/src/tensor/contiguous.rs
```diff
@@ -181,6 +267,36 @@ pub fn into_contiguous_pitched<R: Runtime, E: CubePrimitive>(
...
+/// `shape` refers to the actual (unpacked) shape of the tensor, while `packing` specifies the
+/// number of elements in each storage element.
+///
+/// # Warning
+/// This assumes `u32` or `u8` packing.
+pub fn into_contiguous_packed<R: Runtime, I: Int>(
+    client: &ComputeClient<R::Server>,
+    input: &TensorHandleRef<'_, R>,
+    shape: &[usize],
+    packing: u32,
```
nathanielsimard, line 284: What does the packing value means?

crates/cubecl-matmul/src/base.rs
wingertge: because this is the native store, which uses a fixed type (`e2m1x2`). I guess I could use `scheme.native_packing()` to make it a bit more generic.

wingertge: Actually we need to assert the packed type (`u8`) so it should be specific to e2m1.

crates/cubecl-matmul/src/kernels/naive.rs
wingertge: I can do that but it's already guaranteed because `line_size = max(lhs_line_size, rhs_line_size)`.

crates/cubecl-std/src/tensor/contiguous.rs
wingertge: It's just a private function to make it a bit cleaner but I added a basic description

crates/cubecl-std/src/tensor/contiguous.rs
wingertge: the number of elements being packed together, same as `packing` in `StorageType::Packed`. It's mentioned in the doc comment too.

crates/cubecl-std/src/tensor/contiguous.rs
```diff
@@ -256,6 +374,82 @@ pub fn into_contiguous_ref<R: Runtime, E: CubePrimitive>(
...
+        rank - 1,
+    );
+    let num_vecs = num_elems / line_size as usize;
+    let num_sm = client
+        .properties()
+        .hardware
+        .num_streaming_multiprocessors
+        .unwrap_or(NUM_SM_APPROX);
+    let simul_vecs = num_sm * CubeDim::default().num_elems();
+    let mut elems_per_unit = match num_vecs as u32 / simul_vecs {
```
louisfd, line 402: Can you generalize the formula? Does `2.pow(floor(log2(ratio)))` or `1 << ratio.ilog2()` with `ratio = num_vecs as u32 / simul_vecs` do?

wingertge: maybe, when combined with a `.min(8)`. Looking at it now it really should use the output line size as well, since that's what's actually being stored. Maybe the limit should be based on the number of total 32-bit registers as well, `8` would be 32 registers total for an output that's aligned to 128 bits. We might want to do more testing to find the right heuristic, but keeping the current max would probably be fine for now.
I'll definitely put it in its own function so it doesn't need to be repeated.

wingertge: Done 👍

wingertge: I had to revert it because somehow it was causing issues with `cubecl-cpu`. I think the issue is with the bounds checks in `cubecl-cpu`, but I'll figure it out in a separate PR.

PR #978, opened 2025-10-22
fix: Quant matmul line sizes

wingertge:
Various fixes to line size handling, to fix quantized matmul with large matrices.
It's still due to the restrictions on algorithms, so for now line size is set to 1 for quantized inputs. It's significantly faster than keeping it vectorized.


crates/cubecl-matmul/src/kernels/layered/base.rs
```diff
@@ -179,14 +179,24 @@ fn launch_inner_ref<R: Runtime, MP: MatmulPrecision, A: Algorithm>(
...
     let line_sizes = A::filter_line_sizes(line_sizes);
-    let line_sizes = line_sizes
+    let mut line_sizes = line_sizes
         .filter_lhs_with_tensor(lhs.strides, lhs.shape, problem.lhs_layout)
         .filter_rhs_with_tensor(rhs.strides, rhs.shape, problem.rhs_layout)
         .filter_out_with_tensor(out.strides, out.shape)
         .pick_max()?;
 
+    // The large line size resulting from dequantizing ends up slower due to restrictions on
+    // algorithms. Use this as a quick and dirty fix.
```
louisfd, line 192: Are you working on the cleaner fix?

wingertge: I'm thinking about it 😅 
The perfect fix would involve a lot more interdependence between line size selection, the tiling algorithm and the tile size. You really want the max line size that's also compatible with the tiling scheme and tile size, and that's non-trivial to implement because the tile size may depend on line size, but the line size is also limited by hardware constraints on the tile size.

PR #997, opened 2025-10-28
Better error handling in the tuner

nathanielsimard:
It allows local tuners to not execute the fallback operation when tuning while making it available as a true fallback.


antimora:
Stale

PR #1000, opened 2025-10-29
feat: Granular math mode

wingertge:
Allows applying `FastMath` options on a per function basis, rather than a per kernel basis. This allows us to apply it only to performance-critical sections without enforcing the rules for the entire kernel. This also means CUDA now uses intrinsics dispatched when the required options are enabled, rather than the global and heavy-handed `--use-fast-math`.
Also adds `inverse_sqrt`/`rsqrt` because it has its own intrinsic, and in fact CUDA's `sqrt` is implemented as `1.0 / rsqrt(x)`. It seemed natural to add in this PR since it tightly relates to the `FastMath` changes.

### Instructions

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -103,14 +103,11 @@ tracel-llvm = { version = "20.1.4-5", features = ["mlir-helpers"] }
 # tracel-llvm = { git = "https://github.com/tracel-ai/tracel-llvm.git", branch = "fix/linux", package = "tracel-llvm", features = ["mlir-helpers"] }
 # tracel-llvm = { path = "../tracel-llvm/crates/tracel-llvm", features = ["mlir-helpers"] }
 
-# CubeCL-CUDA
-cudarc = { version = "0.17.5", features = [
+cudarc = { version = "0.17.7", features = [
     "std",
     "driver",
     "nvrtc",
-    "cuda-version-from-build-system",
-    "dynamic-loading",
-], default-features = false }
+], default-features = false } # CubeCL-CUDA
 
 # CubeCL-SPIR-V
 rspirv = { package = "tracel-rspirv", version = "0.12.0" }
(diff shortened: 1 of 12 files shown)


crates/cubecl-core/src/frontend/container/shared_memory.rs
```diff
@@ -5,7 +5,7 @@ use crate::{
     prelude::{Lined, LinedExpand},
     unexpanded,
 };
-use cubecl_ir::{Instruction, Operation, VariableKind};
+use cubecl_ir::{Marker, VariableKind};
```
nathanielsimard, line 8: What is a Marker?

crates/cubecl-cpp/src/shared/base.rs
```diff
@@ -1056,17 +1138,52 @@ impl<D: Dialect> CppCompiler<D> {
                 instructions.push(Instruction::Neg(self.compile_unary(op, out)))
             }
             gpu::Arithmetic::Normalize(op) => {
-                instructions.push(Instruction::Normalize(self.compile_unary(op, out)))
+                let op = self.compile_unary(op, out);
+                instructions.push(self.select_fast_float(
+                    out.ty,
+                    modes,
+                    FastMath::ReducedPrecision | FastMath::NotNaN | FastMath::NotInf,
```
nathanielsimard, line 1145: Why is there fast math hardcoded options here?

nathanielsimard (commented):
I would add some documentation in burn cubecl book on fast math with examples.

crates/cubecl-core/src/frontend/container/shared_memory.rs
wingertge: Currently just `Free`, but a general category for things that don't emit instructions. It's easier to ignore them and make sure they aren't moved if it's a separate category that can't be reflected.

crates/cubecl-cpp/src/shared/base.rs
wingertge: Those are the modes required for the intrinsic. CUDA doesn't have fine-grained settings like Vulkan, so `select_fast_float` checks if the required flags are set and then switches to the intrinsic, otherwise the default implementation. You can take a look at the args for `select_fast_float`.

nathanielsimard (approved):

PR #1001, opened 2025-10-29
Fix/memory usage

nathanielsimard:
Fix memory usage using sliced memory pools by reimplementing memory page.

### DIFF
--- a/crates/cubecl-runtime/src/memory_management/memory_pool/memory_page.rs
+++ b/crates/cubecl-runtime/src/memory_management/memory_pool/memory_page.rs
@@ -42,7 +42,7 @@ impl MemoryPage {
         };
         let id = *page.handle.id();
         let index = 0;
-        this.slices.insert(index, page);
+        this.slices.push(page);
         this.slices_map.insert(id, index);
 
         this


crates/cubecl-runtime/src/memory_management/memory_pool/memory_page.rs
```diff
@@ -0,0 +1,683 @@
...
+        self.tasks.push(returned);
+        task
+    }
+}
+
+#[derive(Debug)]
+enum MemoryTaskStatus {
+    Merging,
+    StartMerging,
+    Ingoring,
```
laggui: Typo

```suggestion
    Ignoring,
```

crates/cubecl-runtime/src/memory_management/memory_pool/memory_page.rs
```diff
@@ -0,0 +1,683 @@
...
+
+        let page = Slice {
+            handle: SliceHandle::new(),
+            storage,
+            padding: 0,
+        };
+        let id = *page.handle.id();
+        let index = 0;
+        this.slices.insert(index, page);
+        this.slices_map.insert(id, index);
```
laggui, line 46: Why not just push? Everything is already empty at this point. Just a minor suggestion

```suggestion
        this.slices.push(page);
        this.slices_map.insert(id, 0);
```

laggui (approved):
Some minor comments.

/edit: hmmm just saw the tests are still failing, looks like we might be reading back into the wrong handles

PR #1007, opened 2025-11-02
feat: Update tune key to enable safely tuning TMA algorithms

wingertge:
Changes required to support safely tuning TMA. Adds a new stride alignment property to the key, and fixes a few bugs with the launch code.

### Instructions

### DIFF
--- a/crates/cubecl-matmul/src/kernels/layered/algorithm/simple_tma.rs
+++ b/crates/cubecl-matmul/src/kernels/layered/algorithm/simple_tma.rs
@@ -4,8 +4,7 @@ use cubecl_core::{Runtime, client::ComputeClient};
 
 use crate::{
     components::{
-        AvailableLineSizes, MatmulElems, MatmulLineSizes, MatmulProblem, MatmulSelection,
-        MatmulSetupError,
+        MatmulElems, MatmulLineSizes, MatmulProblem, MatmulSelection, MatmulSetupError,
         batch::{PartitionedBatchMatmulFamily, RowMajorGlobalPartitionMatmul},
         global::{PlaneWriterFamily, single_stage::tma::SimpleTmaMatmulFamily},
         stage::{FilledStageFamily, PlaneMatmulFamily, StridedStageFamily},


crates/cubecl-matmul/src/components/global/args.rs
```diff
@@ -369,44 +369,57 @@ impl<Lhs: Numeric, Rhs: Numeric, EO: Numeric> ConcreteInputsFactory
...
         // with transposed shape and stride. Tensor metadata still has the normal layout.
         if matches!(problem.lhs_layout, components::MatrixLayout::ColMajor) {
-            lhs_shape.swap(lhs_rank - 1, lhs_rank - 2);
+            lhs_shape.swap(2, 1);
             lhs_strides.swap(lhs_rank - 1, lhs_rank - 2);
             lhs_transposed = true;
         }
         if matches!(problem.rhs_layout, components::MatrixLayout::ColMajor) {
-            rhs_shape.swap(rhs_rank - 1, rhs_rank - 2);
+            rhs_shape.swap(2, 1);
```
nathanielsimard, line 408: The shape is always 3D?

crates/cubecl-matmul/src/kernels/layered/algorithm/simple_tma.rs
```diff
@@ -48,4 +49,12 @@ where
     ) -> Result<MatmulSelection, MatmulSetupError> {
         plane_matmul_selection::<TMM, R>(client, problem, plane_dim, elems, Default::default())
     }
+
+    fn filter_line_sizes(_available_line_sizes: AvailableLineSizes) -> AvailableLineSizes {
+        AvailableLineSizes {
+            lhs: vec![1],
+            rhs: vec![1],
+            out: vec![1],
+        }
```
nathanielsimard: Why do we disable vectorization with TMA? I would add a comment here.

crates/cubecl-matmul/src/tune_key.rs
```diff
@@ -139,13 +169,21 @@ impl MatmulAutotuneKey {
...
+fn stride_align(strides: &[usize], exclude_dim: usize, elem: ElemType) -> u8 {
+    let max = MAX_STRIDE_FACTOR;
+    let mut strides = strides.to_vec();
+    strides.remove(exclude_dim);
+    let factor = strides
+        .iter()
+        .map(|it| (*it * elem.size_bits()) / 8)
+        .map(|it| it.trailing_zeros())
+        .min()
+        .unwrap_or(max);
```
nathanielsimard: I think we could avoid the `to_vec` here.

crates/cubecl-runtime/src/tune/base.rs
```diff
@@ -144,13 +144,16 @@ impl TunePlan {
...
+            Some(val) => *val,
             None => return indices,
         };
 
-        let (mut group_indices, cleanup) = self.group_plan_next(*priority);
+        let (mut group_indices, cleanup) = self.group_plan_next(priority);
         self.cleanup(cleanup);
-        indices.append(&mut group_indices);
+
+        if priority >= 0 {
```
nathanielsimard, line 159: I would add proper docs saying that negative priorities are not tuned.

crates/cubecl-matmul/src/components/global/args.rs
wingertge: For TMA, yes. TMA only has 1D-5D instructions, and they're different functions entirely so need to be dispatched differently. So for TMA I've decided to merge all batch dims into one, rather than have some complicated mapping with up to 3 batch dims (more can't work anyways because 5D is the max).

wingertge: The launch code treats swapped or broadcast batches as discontiguous and makes them contiguous before launching, so it's safe to do this.

crates/cubecl-matmul/src/kernels/layered/algorithm/simple_tma.rs
wingertge: I actually did a slightly nicer version of this override in a recent change, and added a comment. This one is actually incorrect because `out` isn't TMA so it can be vectorized just fine.

crates/cubecl-runtime/src/tune/base.rs
wingertge: I agree, just not sure where. The documentation is pretty sparse and very scattered.

wingertge: I added it and some other docs that I asked you about to `group`, let me know if that's good (and if my docs are correct).

crates/cubecl-matmul/src/kernels/layered/algorithm/simple_tma.rs
wingertge: To further explain, since TMA works in blocks, vectorization is not meaningful, similar to CMMA's load function. So to avoid having to constantly convert between line sizes, both the stage and global memory are treated as unvectorized. Tile matmuls can reinterpret it since TMA is CUDA-only anyways, and that supports reinterpreting line sizes.

PR #1031, opened 2025-11-12
Matmul: decouple reader config from global config

louisfd:
The goal is to use reader config in flash attention without being tied to GlobalConfig


louisfd:
A better PR encapsulating this one will come soon

PR #1035, opened 2025-11-13
feat: Swizzled shared memory

wingertge:
Adds shared memory swizzling to all stage and tile readers that can support it without any loading changes. Async readers apart from TMA are not yet supported, because they contiguously load more than 16 bytes at once and would need modfication to support 16-byte swizzle atoms.
Also adds support for alternate stage types, though this ended up not being used in the end because I managed to merge the swizzled tiles into the strided stage. Still good to have the infrastructure for future alternate stages, so I didn't want to remove it.

Swizzle modes are limited to the same modes as cutlass and the base TMA feature set introduced in CUDA 12.0, with atom size always set to 16 bytes (one 128-bit load) and swizzle sizes from 32 to 128 bytes. Larger atoms or the swapped bytes that exist in newer versions of TMA don't seem to be beneficial to matrix multiplication.

### Instructions

### DIFF
--- a/crates/cubecl-common/src/bytes/base.rs
+++ b/crates/cubecl-common/src/bytes/base.rs
@@ -1,7 +1,6 @@
 //! A version of [`bytemuck::BoxBytes`] that is cloneable and allows trailing uninitialized elements.
 
 use crate::bytes::default_controller::{self, NativeAllocationController};
-use crate::bytes::file::FileAllocationController;
 use alloc::boxed::Box;
 use alloc::vec::Vec;
 use core::alloc::LayoutError;
@@ -172,7 +171,7 @@ impl Bytes {
     #[cfg(feature = "std")]
     /// Creates bytes from a file at the given offset of the given size.
     pub fn from_file<P: Into<std::path::PathBuf>>(file: P, size: u64, offset: u64) -> Self {
-        let controller = FileAllocationController::new(file, size, offset);
+        let controller = crate::bytes::file::FileAllocationController::new(file, size, offset);
 
         Self {
             controller: Box::new(controller),
--- a/crates/cubecl-cpp/src/cuda/ptx/mod.rs
+++ b/crates/cubecl-cpp/src/cuda/ptx/mod.rs
@@ -294,7 +294,7 @@ pub fn ldmatrix_template<D: Dialect>(elem: Elem<D>, factor: u32, transpose: bool
         r#"
 inline __device__ void
 __ldmatrix_m{width}n8_{elem}_{factor}x{is_transposed}({args}) {{
-  uint64 addr = reinterpret_cast<uint64>(row_addr);
+  uint64 addr = __cvta_generic_to_shared(row_addr);
   asm volatile("ldmatrix.sync.aligned.m8n{width}.{num}{transposed_arg}.shared::cta.{ty}"
                " {placeholders_reg}, {placeholder_addr};"
                : {params_out}


crates/cubecl-matmul/src/components/global/read/strategy/sync_partial_cyclic.rs
```diff
@@ -223,10 +239,10 @@ pub(crate) fn load_and_store_line<EG: Numeric, ES: Numeric, TO: TilingOrder, G:
     let line_read = view.read_checked((tile, pos_within_tile));
 
     let tile_start = tile_index * job.num_lines_per_tile;
-    let tile_end = tile_start + job.num_lines_per_tile;
-    let mut tile_slice = stage
-        .as_slice_mut(line_size)
-        .slice_mut(tile_start, tile_end);
+    let mut tile_slice = stage.as_slice_mut(line_size);
+    let offset = tile_start + pos_within_tile / line_size;
+    let type_size = type_size::<ES>(line_size);
```
nathanielsimard, line 244: Is type size a way to retrieve the size of the element even when "fake"

crates/cubecl-matmul/src/components/stage/memory/config.rs
```diff
@@ -11,9 +11,47 @@ pub struct StageMemoryConfig {
...
+    /// Swizzle 16B chunks within 32B span
+    /// Swizzle<1,4,3>
+    B32,
+    /// Swizzle 16B chunks within 64B span
+    /// Swizzle<2,4,3>
+    B64,
+    /// Swizzle 16B chunks within 128B span
+    /// Swizzle<3,4,3>
+    B128,
+}
```
nathanielsimard, line 35: I would add a definition of swizzling atom, the chunks and the span in the SwizzleMode docs. Quite useful.

crates/cubecl-matmul/src/components/global/read/strategy/sync_partial_cyclic.rs
wingertge: It's to adjust for line size in a slightly cleaner way. It returns the size of a line of the generic type with `line_size`.

crates/cubecl-matmul/src/components/stage/memory/layout.rs
```diff
@@ -263,10 +281,13 @@ impl TilingOrder for TmaTilingOrder {
 pub trait TilingLayout: 'static + Send + Sync + Clone + Copy + TilingValidation {
```
louisfd, line 281: Do you think we could remove the trait altogether in favor of the enum?

crates/cubecl-matmul/src/components/stage/memory/config.rs
wingertge: It's defined on `Swizzle` itself, it's easier to define when you relate it to the actual address transformation with an example.

crates/cubecl-matmul/src/components/stage/memory/layout.rs
wingertge: The trait is useful because I'm implementing it for a custom layout in convolution btw, with the enum that would need to be upstreamed into `cubecl-matmul`. The enum was mainly for the TMA launch logic to determine the tile size, but then I changed the logic to use automatic selection based on swizzling so it's kinda unused. But I can definitely see a use for knowing the layout in the `ConreteInputsFactory` so I left it in for now.

PR #1041, opened 2025-11-14
Fix deadlock when copy 

nathanielsimard:
Since we're locking two contexts at the same time, deadlocks were possible. I added a device-kind lock, which fixes that problem while still keeping the convenience of having multiple mutable contexts locked at once.


crates/cubecl-common/src/device.rs
```diff
@@ -369,7 +379,10 @@ mod context {
     static GLOBAL: spin::Mutex<DeviceLocator> = spin::Mutex::new(DeviceLocator { state: None });
 
     struct DeviceLocator {
-        state: Option<HashMap<Key, DeviceStateLock>>,
+        state: Option<(
+            HashMap<Key, DeviceStateLock>,
+            HashMap<TypeId, Arc<ReentrantMutex<()>>>,
+        )>,
```
laggui: We should make this into a type, will make field accesses more explicit too

PR #1044, opened 2025-11-15
perf: MMA line size

wingertge:
Change MMA matmul to use the highest contiguous line size present in the matrix, up to 128-bit. This should speed up contiguous loads on HIP.

### Instructions

### DIFF
--- a/crates/cubecl-cpp/src/cuda/mma/ptx_wmma_compiler.rs
+++ b/crates/cubecl-cpp/src/cuda/mma/ptx_wmma_compiler.rs
@@ -575,6 +575,9 @@ pub(super) fn compile_manual_mma<D: Dialect>(
 
     let frag_a = (0..a_regs).map(|i| as_const_ty(format!("{frag_a}[{i}]"), ab_ty));
     let frag_b = (0..b_regs).map(|i| as_const_ty(format!("{frag_b}[{i}]"), ab_ty));
+
+    // C and D fragments are always vectorized for optimal stores and casts, but are taken as separate
+    // registers for f32 so we need to unpack it. f16 is taken in packed registers, so use as is.
     let frag_c = match cd_elem.size() {
         4 | 8 => (0..cd_regs)
             .map(|i| as_ty(format!("{frag_c}[{}].i_{}", i / 2, i % 2), cd_ty))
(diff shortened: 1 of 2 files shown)


crates/cubecl-cpp/src/cuda/mma/ptx_wmma_compiler.rs
```diff
@@ -575,8 +575,24 @@ pub(super) fn compile_manual_mma<D: Dialect>(
...
+    };
+    let frag_d = match cd_elem.size() {
+        4 | 8 => (0..cd_regs)
+            .map(|i| as_ty(format!("{frag_d}[{}].i_{}", i / 2, i % 2), cd_ty))
+            .collect::<Vec<_>>(),
+        2 => (0..cd_regs)
+            .map(|i| as_ty(format!("{frag_d}[{i}]"), cd_ty))
+            .collect::<Vec<_>>(),
+        other => panic!("Found unhandled accumulator elem size {other}"),
+    };
```
nathanielsimard, line 598: I would add comments on top of those code blocks.

crates/cubecl-cpp/src/hip/mma/wmma_intrinsics_compiler.rs
```diff
@@ -324,12 +324,59 @@ impl DialectWmmaCompiler<HipDialect<Self>> for WmmaIntrinsicCompiler {
         if flags.elem_bf16 {
             f.write_str("typedef __bf16 bhalf8_t __attribute__((ext_vector_type(8)));\n")?;
             f.write_str("typedef __bf16 bhalf16_t __attribute__((ext_vector_type(16)));\n")?;
+            if flags.inst_mma {
```
nathanielsimard: Same here, this is quite hard to follow

wingertge: I removed the union stuff but added a comment to the fragment constructors

crates/cubecl-cpp/src/cuda/mma/ptx_wmma_compiler.rs
wingertge: Added a quick explanation to the C/D unpacking code, A/B seem pretty self-explanatory because it's just indexing into the array normally

crates/cubecl-ir/src/runtime_properties.rs
```diff
@@ -29,6 +29,52 @@ pub struct MmaProperties {
...
+    pub fn new(func: impl Fn(MatrixIdent, Matrix) -> u32 + 'static) -> Self {
+        Self {
+            inner: alloc::rc::Rc::new(func),
+        }
+    }
+
+    pub fn apply(&self, ident: MatrixIdent, matrix: Matrix) -> u32 {
+        (self.inner)(ident, matrix)
+    }
+}
```
nathanielsimard, line 51: That seems kind of strange to have that in runtime properties.

wingertge: yeah we don't have a way to access target-specific properties at compile time right now, and line size needs to be comptime. I think at some point it might be good to have a dyn-compatible trait that can handle the same thing and is implemented by the compiler or something. But it's just one function for now so this is simpler.

PR #1048, opened 2025-11-16
Add support for hypot and reciprocal hypot.

vaijira:

### Instructions

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -59,8 +59,8 @@ web-time = "1.1.0"
 rstest = "0.25.0"
 serial_test = "3.1.1"
 
-bytes = { version = "1.10", default-features = false }
 bytemuck = "1.16.1"
+bytes = { version = "1.10", default-features = false }
 float4 = "0.1"
 float8 = { version = "0.4", default-features = false }
 half = { version = "2.5", features = [
@@ -103,7 +103,7 @@ tracel-llvm = { version = "20.1.4-5", features = ["mlir-helpers"] }
 # tracel-llvm = { git = "https://github.com/tracel-ai/tracel-llvm.git", branch = "fix/linux", package = "tracel-llvm", features = ["mlir-helpers"] }
 # tracel-llvm = { path = "../tracel-llvm/crates/tracel-llvm", features = ["mlir-helpers"] }
 
-cudarc = { version = "0.18.1", features = [
+cudarc = { version = "0.18.2", features = [
     "std",
     "driver",
     "nvrtc",
--- a/crates/cubecl-common/Cargo.toml
+++ b/crates/cubecl-common/Cargo.toml
@@ -32,7 +32,6 @@ std = [
 
 [dependencies]
 # ** Please make sure all dependencies support no_std when std is disabled **
-bytes = { workspace = true, optional = true }
 bytemuck = { workspace = true, features = ["derive"] }
 derive-new = { workspace = true }
 derive_more = { workspace = true, features = [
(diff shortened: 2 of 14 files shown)


vaijira:
Ready for review, to avoid underflow/overflow the following algorithms are:

* Hypot(lhs, rhs)
     let a = abs(lhs);
     let b = abs(rhs);
     let max_val = max(a, b);
     var max_val_safe = max_val;
     if (max_val == 0.0) {{ max_val_safe = 1.0; }}
     let min_val = min(a, b);
     let t = min_val / max_val_safe;
     return max_val * sqrt(fma(t, t, 1.0));

* Rhypot(lhs, rhs)
     let a = abs(lhs);
     let b = abs(rhs);
     let max_val = max(a, b);
     var max_val_safe = max_val;
     if (max_val == 0.0) {{ max_val_safe = 1.0; }}
     let min_val = min(a, b);
     let t = min_val / max_val_safe;
     return inverseSqrt(fma(t, t, 1.0)) / max_val;

nathanielsimard (commented):
I think it would be better to have a cubecl function for that instruction instead of re-implementing it for every compilers. Note that I'm not against having specialized implementation for a compiler, but we can use a cubecl function as default. Here's an example:

```rust
/// Returns the value at `index` in tensor within bounds.
#[cube]
pub fn read_tensor_checked<C: CubePrimitive>(
    tensor: Tensor<C>,
    index: u32,
    #[comptime] unroll_factor: u32,
) -> C {
    let len = tensor.buffer_len() * unroll_factor;
    let in_bounds = index < len;
    let index = Min::min(index, len);

    select(in_bounds, tensor.read_unchecked(index), C::cast_from(0u32))
}
```

Called as follow by compilers:

```rust
read_tensor_checked::expand::<Line<NumericExpand<0>>>(
        &mut scope,
        list.into(),
        index.into(),
        op.unroll_factor,
    )
    .expand
```

vaijira:
The problem is that for CUDA that has built-in instructions you really want to call hypotf and rhypotf.

nathanielsimard:
> The problem is that for CUDA that has built-in instructions you really want to call hypotf and rhypotf.

You can use that instruction for cuda and fallback on the default implementation for other compilers. other compiler can expand the cubecl function instead of recreating it with their own IR.

vaijira:
I think it could be reviewed.

crates/cubecl-cpp/src/shared/dialect.rs
```diff
@@ -652,6 +652,32 @@ pub trait DialectInstructions<D: Dialect> {
...
+    fn compile_instruction_rhypot(
+        f: &mut std::fmt::Formatter<'_>,
+        lhs: &str,
+        rhs: &str,
+        elem: Elem<D>,
+    ) -> std::fmt::Result {
+        match elem {
+            Elem::F32 => write!(f, "rhypotf({lhs}, {rhs})"),
+            Elem::F64 => write!(f, "rhypot({lhs}, {rhs})"),
+            _ => panic!("Unsupported type for hypot"),
```
nathanielsimard: Instead of panic, write an error in the CPP that will cause a compilation error that may be handled by a user:

```rust
write!(f, "#error Unsupported type for hypot: {elem}")
```

crates/cubecl-cpu/src/compiler/passes/trigonometries_transform.rs
```diff
@@ -0,0 +1,37 @@
...
+
+#[derive(Debug)]
+pub(crate) struct HypotTransform;
+
+impl IrTransformer for HypotTransform {
+    fn maybe_transform(&self, scope: &mut Scope, inst: &Instruction) -> TransformAction {
+        match &inst.operation {
+            Operation::Arithmetic(Arithmetic::Hypot(op)) => {
+                let mut scope = scope.child();
+                expand_hypot(&mut scope, op.lhs, op.rhs, inst.out.unwrap());
```
nathanielsimard, line 15: From the imports, I guess the `expand_hypot` function comes from the prelude. However I don't think we should include that in the prelude, a normal import would be better.

nathanielsimard (changes requested):
The refactor is nice, great job! I think only a few minor details need to be fixed before we can merge.

nathanielsimard (approved):

PR #1049, opened 2025-11-16
Fix fma so its callable from #cube functions

amfaber:
# Fix the fma instruction so it can be called from #[cube] functions
Surprisingly, the FMA function doesn't seem to have been used anywhere in either cubecl or burn. Before this PR, it is not usable in #[cube] functions, as it doesn't adhere to the "calling convention" (if it can be called that) of the macro, meaning that the "expand" version of any given function "func" should be found under a module of the same name, i.e. "func::expand".

This PR fixes that by using #[cube] and the intrinsic!{} macro to directly implement fma and its expand counterpart.

### DIFF
--- a/crates/cubecl-core/src/frontend/operation/fma.rs
+++ b/crates/cubecl-core/src/frontend/operation/fma.rs
@@ -1,27 +1,32 @@
-use cubecl_macros::{cube, intrinsic};
-
-use crate::{
-    self as cubecl,
-    ir::{Arithmetic, FmaOperator, Instruction},
-    prelude::CubePrimitive,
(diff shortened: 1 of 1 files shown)


amfaber:
The macro gave some issues when actually using it - I assume it was due to not being able to infer the generic in the expanded version. I reverted it to a manual expansion. If it is possible to use the macros in this case, I'd love to know how

amfaber:
Awesome :) Just for the record, I found the CPU backend to emit different MLIR and assembly with `a += b * c` and `a = fma(b, c, a)`

PR #1056, opened 2025-11-19
feat: Implement stmatrix and stage casting to support it

wingertge:
Implements `ldmatrix` and some background work required to support it (and TMA storing if we ever add that).
Stages are now set to the global type when the tile matmul allows casting on read/write. This also lowers shared usage for f16 output.
To support this, register types are now properly passed to the kernel instead of always being set to the stage type.

### Instructions

### DIFF
--- a/crates/cubecl-common/src/format.rs
+++ b/crates/cubecl-common/src/format.rs
@@ -1,0 +1,119 @@
+use alloc::format;
+use alloc::string::String;
+
+/// Format strings for use in identifiers and types.
+pub fn format_str(string: &str, markers: &[(char, char)], include_space: bool) -> String {
+    let mut result = String::new();
+    let mut depth = 0;
+    let indentation = 4;
+
+    let mut prev = ' ';
+    let mut in_string = false;
+
+    for c in string.chars() {
+        if c == ' ' {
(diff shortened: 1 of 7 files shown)


crates/cubecl-matmul/src/components/tile/plane_vec_mat_inner_product/setup.rs
```diff
@@ -31,6 +31,10 @@ where
         false
     }
 
+    fn can_cast_stage() -> bool {
```
nathanielsimard: Can it be false if the compiler can't change the vectorization factor of a line?

wingertge: This is the confusion I was worried about but couldn't come up with a better name at the time 😅 
I think `can_cast_stage_elem` might be clearer. It's not about casting the stage itself, but the element after loading. There's no need to change vectorization for that.

wingertge: Basically, whether registers can be separate from the stage element, or whether they need to be the same like in CMMA.

PR #1057, opened 2025-11-19
Added Plugin system in cubecl-runtime with an  example in cubecl cuda…

Q-Lucca:
Hello Lucca from Use-Ai.rs here,

Decided on branding therefore know my final account for my project. 

So, to enable the possibility to add additional features to ComputeServer which can be accessed through the ComputeClient i propose the Plugin system implemented here. For now i used it to get nccl to cubecl-cuda. The system allows to introduce and initialize additional types and execute additional functions over ComputeServer. This is used to not lose threads when building worker based Runtime for integrated production deployments. A library released next week will be in need for a implementation of that kind.

This is a draft with the first docs within the files. Before implementing full docs for cubecl i would need to know if you guys would generally accept a full proposal?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,7 @@ edition = "2024"
 license = "MIT OR Apache-2.0"
 readme = "README.md"
 rust-version = "1.88"
-version = "0.9.0-pre.2"
+version = "0.9.0-pre.3"
 
 [workspace.dependencies]
 bitflags = { version = "2.9.1", features = ["serde"] }
--- a/crates/cubecl-attention/Cargo.toml
+++ b/crates/cubecl-attention/Cargo.toml
@@ -19,12 +19,12 @@ attention_tests = []
 
 [dependencies]
 bytemuck = { workspace = true }
-cubecl-common = { path = "../cubecl-common", version = "0.9.0-pre.2", default-features = false }
-cubecl-core = { path = "../cubecl-core", version = "0.9.0-pre.2", default-features = false }
-cubecl-runtime = { path = "../cubecl-runtime", version = "0.9.0-pre.2", default-features = false }
-cubecl-std = { path = "../cubecl-std", version = "0.9.0-pre.2", default-features = false }
-cubecl-matmul = { path = "../cubecl-matmul", version = "0.9.0-pre.2", default-features = false }
-cubecl-random = { path = "../cubecl-random", version = "0.9.0-pre.2", default-features = false }
+cubecl-common = { path = "../cubecl-common", version = "=0.9.0-pre.3", default-features = false }
+cubecl-core = { path = "../cubecl-core", version = "=0.9.0-pre.3", default-features = false }
+cubecl-runtime = { path = "../cubecl-runtime", version = "=0.9.0-pre.3", default-features = false }
+cubecl-std = { path = "../cubecl-std", version = "=0.9.0-pre.3", default-features = false }
+cubecl-matmul = { path = "../cubecl-matmul", version = "=0.9.0-pre.3", default-features = false }
+cubecl-random = { path = "../cubecl-random", version = "=0.9.0-pre.3", default-features = false }
 half = { workspace = true, features = ["bytemuck"] }
 pretty_assertions = { workspace = true, optional = true }
 serde = { workspace = true }
--- a/crates/cubecl-common/src/format.rs
+++ b/crates/cubecl-common/src/format.rs
@@ -104,7 +104,7 @@ mod tests {
     #[test]
     fn test_format_debug() {
         let test = Test {
-            map: HashMap::from_iter([("Hey with space".to_string(), 8)].into_iter()),
+            map: HashMap::from_iter([("Hey with space".to_string(), 8)]),
         };
 
         let formatted = format_debug(&test);
--- a/crates/cubecl-convolution/Cargo.toml
+++ b/crates/cubecl-convolution/Cargo.toml
@@ -19,13 +19,13 @@ conv_tests = []
 
 [dependencies]
 bytemuck = { workspace = true }
-cubecl-common = { path = "../cubecl-common", version = "0.9.0-pre.2", default-features = false }
-cubecl-core = { path = "../cubecl-core", version = "0.9.0-pre.2", default-features = false }
-cubecl-matmul = { path = "../cubecl-matmul", version = "0.9.0-pre.2", default-features = false }
-cubecl-runtime = { path = "../cubecl-runtime", version = "0.9.0-pre.2", default-features = false }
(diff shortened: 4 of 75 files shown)


crates/cubecl-runtime/src/plugin.rs
```diff
@@ -0,0 +1,62 @@
...
+use crate::server::ComputeServer;
+
+/// This trait defines all types and the function trait for a `Plugin`.
+pub trait Plugin: Send + Sync + 'static {
+    /// This is a structure of some kind
+    /// representing all data the client layer needs to provide
+    /// for executing functions of the `Plugin` on a `ComputeServer`.
+    type ClientHandle: Send;
+    /// This is a type which needs to be build by a `ComputeServer`.
+    type ServerHandle: Send;
```
nathanielsimard, line 18: It is strange to have two handles here, I think it would be cleaner to have maybe 3 associated types:

1. Input (ServerHandle) the input of the function.
2. Output (ReturnVal) the output of the function.
3. State (InitType) the state of the plugin that needs an init.

I would ditch the client handle.

crates/cubecl-runtime/src/plugin.rs
```diff
@@ -0,0 +1,62 @@
...
+    /// A `Plugin` wants to add additional functions to a `ComputeServer`.
+    /// Since we know how the `ServerHandle` will look the `Plugin` will
+    /// be able to define additional functions which can be executed over the
+    /// `ServerHandle` by a `ComputeServer`
+    type Fns: FnOnce(Self::ServerHandle) -> Result<Self::ReturnVal, PluginError>;
```
nathanielsimard, line 23: Why use an associated type for a function rather than.. a function:

```rust
fn execute(input: Self::Input) -> Result<Self::Output, PluginError>;
```

I think it would be cleaner.

crates/cubecl-runtime/src/plugin.rs
```diff
@@ -0,0 +1,62 @@
...
+    /// Extension name for complex runtime operations.
+    const EXTENSION_NAME: &'static str;
+}
+
+/// Type we want for initialisation.
+pub trait PluginType {
+    type Insert: Send + Sync;
+
+    fn init(self) -> Self::Insert;
+}
```
nathanielsimard, line 39: I would put that function inside the Plugin trait:

```rust
pub trait Plugin {
...
fn init(state: Self::State);
}
```

I don't think we need a returned type here.

crates/cubecl-runtime/src/plugin.rs
```diff
@@ -0,0 +1,62 @@
...
+    /// The `ComputeServer` should be able to initialise the `Plugin`.
+    fn init_type(&mut self, plugin_type: SP::InitType, stream: StreamId) -> Result<(), PluginError>;
+
+    /// And should be able to take the `Plugin`'s fns plus `ClientHandle` and
+    /// `StreamId` to build the `ServerHandle` and execute the function.
+    fn plugin_fn<F>(
+        &mut self,
+        client_handle: SP::ClientHandle,
+        stream_id: StreamId,
+        op: SP::Fns,
```
nathanielsimard, line 52: I don't think we need to pass the function as input here, or maybe there's something I'm not understanding.

crates/cubecl-cuda/src/compute/nccl.rs
```diff
@@ -0,0 +1,790 @@
...
+            uid,
+        };
+
+        std::thread::scope(|s| {
+            let h0 = s.spawn(move || {
+                client0
+                    .plugin_init::<NcclExtension>(init0)
+                    .expect("Failed to initialize NCCL on device 0");
+                client0
+                    .plugin_fn::<NcclExtension>(handle0, op)
```
nathanielsimard, line 398: I think it's probably a bit complex for the user API. I like the idea of plugin, but the function call can use enum instead of function definitions. If for instance the `Plugin::Input` is an enum, like `AllReduce`, `Init`, `Gather`, etc. We don't have to support many functions per extension. It would be cleaner to call:

```rust
client.execute::<NcclExtension>(NcclArg::AllReduce(NcclReduce::Avg, inputs, output));
```
Do you think that would be better?

crates/cubecl-runtime/src/plugin.rs
Q-Lucca: First of all thanks for your time and feedback. :) 

Fair enough, I'm not particularly creative and consistent with my choice of words :D So yeah fully with you and will definitely improve on that in future iterations.

crates/cubecl-runtime/src/plugin.rs
Q-Lucca: In my targeted architecture the goal is to decouple cluster function orchestration from the whole cluster lifecycle. So a deployment could profit from ai research while handling production without the need of komplex re-deploynment iterations. There is an hole reseaoning for this decision, but ... (Conv0)

crates/cubecl-runtime/src/plugin.rs
Q-Lucca: (Conv0) ... fair enough, would say there is too much of my architecture leaking in here. These concepts are more for my specific case not for general use. Your suggestion is definitely a better match for CubeCL's general API.

crates/cubecl-runtime/src/plugin.rs
Q-Lucca: Thats true xD Made the type function while designing my fn orchestrator. The function use case is not necessary even with my architecture xD Unnecessary and like you said in future iterations.

crates/cubecl-cuda/src/compute/nccl.rs
Q-Lucca: With a deeper understanding of the differentiations you introduced me to this makes sense. Also had the idea to improve on the execution method of ComputeClient, but i'm carefull with reworking implementations due to the respect of the author. Coming from a corporate employe background my additions already felt illegal. :D Will adapt and make a more integrated solution to abstract complexities inline with CubeCl's design. Already hade a more profound look into it and got an idea will iterate on this. But didn't wanted to do that if you have had no intressed in the general approach.

Q-Lucca:
Ok, big thanks for the input. I have a deeper understanding of the differentiation in our project philosophies. The general goal for my next iteration will be a more straightforward usage within the client for general use while also allowing me to approach plugins with capabilities for optimizations needed in production grade multi service network deployments. The second iteration will propose a integration with the client's execution methods, to abstract complexities in sync with CubeCL while maintaining compabilitie.

crates/cubecl-cuda/src/compute/nccl.rs
nathanielsimard: No worries, I think the plugin system makes sense. It's normal to iterate over a design :)

antimora:
Stale

PR #1062, opened 2025-11-21
Fix/autotuner

nathanielsimard:
When an autotune entry was skipped due to a negative priority, autotuning stopped completely even though other tunables were still available. This led to an "unable to find a proper autotune kernel" runtime error.


laggui (approved):
Nice!

Validated that this fixes the burn test for `test_conv2d_binary_broadcasted` (which was failing on vulkan)

PR #1076, opened 2025-11-27
feat: Rewrite async loaders to make them actually useful

wingertge:
Rewrites async load strategies using lower level instructions because `async_memcpy` is very unreliable and has a tendency to trigger an extremely slow fallback.
Uses these new loaders to enable specialized matmul on older (sm_80) Nvidia cards.
Adds strides to the `MatmulProblem` and passes it to the validation function so TMA and async copy don't crash everything from unaligned  strides.
Refactors `MatmulElems` to store quantization state so TMA and async copy don't crash everything from quantized global tensors.

### Instructions


crates/cubecl-cpp/src/cuda/ptx/copy_async.cuh
```diff
@@ -0,0 +1,52 @@
+inline __device__ void
```
nathanielsimard, line 2: Is it some kind of template?

nathanielsimard: I would add comments on top of each function, probably with links to Nvidia documentation. Inline PTX is hard enough to understand with docs !

crates/cubecl-cpp/src/cuda/ptx/mma.rs
```diff
@@ -0,0 +1,370 @@
+use crate::{
```
nathanielsimard, line 1: Is it a refactor or just a file that was moved?

crates/cubecl-ir/src/barrier.rs
```diff
@@ -80,6 +89,9 @@ pub enum BarrierOps {
         arrive_count_update: Variable,
         transaction_count_update: Variable,
     },
+    CommitCopyAsync {
+        barrier: Variable,
+    },
```
nathanielsimard, line 94: Is there a reason to expose the commit instruction? I guess it's better if we create a batch of copy async calls then commit to create a bigger batch of copy operations that can be optimized by the cuda runtime. If that's the case, I would add documentation on top of the commit function in the frontend telling when to use it and why/how

crates/cubecl-cpp/src/cuda/ptx/mma.rs
wingertge: Just moved to a separate module

crates/cubecl-ir/src/barrier.rs
wingertge: I added a bit more docs but it's hard to explain without going very far into implementation details

crates/cubecl-cpp/src/cuda/ptx/copy_async.cuh
wingertge: Added a bit of documentation but this is generated code, just need to add the wrappers because Nvidia doesn't provide them with the extra arg. They only provide the one with a fixed copy size. And for arrive it's just to simplify the barrier handling at the callsite.

louisfd (approved):
LGTM

PR #1092, opened 2025-12-04
use input type for argmax output

tzemanovic:
Hi,

This PR fixes the output of ArgMax to use the given input tensor dtype. I ran into this using e.g. `max_dim_with_indices` in burn. I tried to call `to_data().into_vec()` on the tensor, which failed cause even though I was using i64 as int type, the data type was i32 (error `TypeMismatch("Invalid target element type (expected I32, got I64)")`).

### Instructions

- [x] Submit a PR in burn with your fixes and link it here: https://github.com/tracel-ai/burn/pull/4124

### DIFF
--- a/crates/cubecl-reduce/src/instructions/mixed.rs
+++ b/crates/cubecl-reduce/src/instructions/mixed.rs
@@ -49,7 +49,7 @@ impl ReduceFnConfig {
             ReduceFnConfig::ArgMax | ReduceFnConfig::ArgMin => {
                 return ReduceDtypes {
                     input: input.into(),
-                    output: i32::as_type_native_unchecked(),
+                    output: input.into(),
                     accumulation: input.into(),
                 };
             }


crates/cubecl-reduce/src/instructions/mixed.rs
```diff
@@ -49,7 +49,7 @@ impl ReduceFnConfig {
             ReduceFnConfig::ArgMax | ReduceFnConfig::ArgMin => {
                 return ReduceDtypes {
                     input: input.into(),
-                    output: i32::as_type_native_unchecked(),
+                    output: input.into(),
```
laggui, line 52: If the input is float though that would break.

Probably the `precision(...)` should take the input and output elem type as input, and then in Burn's `reduce_dim` we can provide both accordingly.

tzemanovic: ah I see, makes sense. I see this has been moved to cubek so I will re-open it there

PR #1098, opened 2025-12-06
Feat/cpu scheduler

nathanielsimard:
Add proper lazy execution + multi-streams support to the cpu runtime.


so-rose:
I was literally needing this today for a concurrent pool of `ComputeClient`s, thank you!

PR #1099, opened 2025-12-07
refactor: Tensor map

wingertge:
Refactors tensor maps to make the kind typed, since it's invalid to use a tiled tensor map with im2col and vice versa. So this should be checked at compile time, and views should only implement the appropriate version of load.
For the purposes of launching, im2col tensor maps now use a layout of `(CoordsDyn, CoordsDyn)` representing the position and offset. There's currently no checking of the dimensionality of each coordinate set, so it's up to the layout to provide the correct one (4D and 2D for example).

### Instructions

### DIFF
--- a/crates/cubecl-cuda/src/compute/server.rs
+++ b/crates/cubecl-cuda/src/compute/server.rs
@@ -716,7 +716,7 @@ fn swizzle_to_cuda(swizzle: TensorMapSwizzle) -> CUtensorMapSwizzle {
         #[cfg(cuda_12080)]
         TensorMapSwizzle::B128Atom64B => CU_TENSOR_MAP_SWIZZLE_128B_ATOM_64B,
         #[cfg(not(cuda_12080))]
-        other => unimplemented!("Swizzle atomicity requires CUDA 12.8 or higher"),
+        _ => unimplemented!("Swizzle atomicity requires CUDA 12.8 or higher"),
     }
 }
 


crates/cubecl-runtime/src/tma.rs
```diff
@@ -2,6 +2,65 @@ use alloc::vec::Vec;
...
+    /// In matmul, for example, this might be `batch x m x k`, or whatever the stage size is.
+    /// If a dimension isn't present in the tile, it should just be set to `1`.
+    ///
+    /// For CUDA, this must be a power of two and `<= 256` on each dimension.
+    pub tile_size: Vec<u32>,
+}
+
+/// Args for im2col tensor maps
+#[cfg_attr(
+    any(target_os = "windows", target_os = "linux", target_os = "macos"),
```
nathanielsimard: We should simply add a serde feature flag

crates/cubecl-std/src/tensor/view/operations/tensor_map.rs
```diff
@@ -67,7 +60,7 @@ macro_rules! impl_tensor_map {
                     _scope: &mut Scope,
                     _pos: <$coords as CubeType>::ExpandType,
                 ) -> ExpandElementTyped<bool> {
-                    unimplemented!("Can't read from tensor map");
+                    true.into()
```
nathanielsimard, line 64: We could say why it's always in-bounds

crates/cubecl-runtime/src/tma.rs
wingertge: Actually I think we can remove them entirely, think it was for an old `Serialize` bound that no longer exists.

nathanielsimard (approved):

PR #1105, opened 2025-12-14
Support const in match patterns for #[cube] macro

sepcnt:
Previously, only literals were supported in match patterns within the `#[cube]` macro.
This change enables support for constants by updating the parser to recognize `Pat::Path` and `Pat::Ident` as constants.
This grammar support is essential to reduce hard-coded constants in code.

Making it possible to compile below example:
```rust
const CASE_0: u32 = 0;
const CASE_1: u32 = 1;

#[cube(launch)]
pub fn kernel_switch_const<F: Float>(output: &mut Array<F>, case: u32) {
    if UNIT_POS == 0 {
        let value = match case {
            CASE_0 => F::new(1.0f32),
            CASE_1 => F::new(3.0f32),
            _ => F::new(5.0f32),
        };
        output[0] = value;
    }
}
```

Detailed changes:
- `crates/cubecl-macros/src/parse/branch.rs`:
    - Updated `numeric_match` to parse patterns into `Vec<Expression>` instead of `Vec<Lit>`.
    - Implemented logic to handle `Pat::Path` and `Pat::Ident` as constants.
    - Used `lit_ty` to properly type literals.
- `crates/cubecl-macros/src/expression.rs`:
    - Updated `Expression::Switch` to store cases as `Vec<(Expression, Block)>`.
- `crates/cubecl-macros/src/generate/expression.rs`:
    - Updated `Switch` generation to use `as_const` or `to_tokens` for case values.
- `crates/cubecl-macros/src/parse/expression.rs`:
    - Made `lit_ty` public for use in `branch.rs`.
- `crates/cubecl-core/src/runtime_tests/branch.rs`:
    - Added `test_switch_const` to verify the new functionality.

### DIFF
--- a/crates/cubecl-core/src/runtime_tests/branch.rs
+++ b/crates/cubecl-core/src/runtime_tests/branch.rs
@@ -41,11 +41,46 @@ pub fn kernel_switch_or_arm<F: Float>(output: &mut Array<F>, case: u32) {
     }
 }
 
+const CASE_0: u32 = 0;
+const CASE_1: u32 = 1;
+
+#[cube(launch)]
+pub fn kernel_switch_const<F: Float>(output: &mut Array<F>, case: u32) {
+    if UNIT_POS == 0 {
+        let value = match case {
+            CASE_0 => F::new(1.0f32),
+            CASE_1 => F::new(3.0f32),
+            _ => F::new(5.0f32),
+        };
+        output[0] = value;
+    }
+}
+
 #[cube(launch)]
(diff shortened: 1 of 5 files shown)


wingertge:
I never implemented that because it's actually not possible to distinguish an identifier from a constant. This is also a problem in normal Rust as a matter of fact, it's very easy to accidentally create an identifier match arm when you meant to use a constant. But compiler errors for this won't be great anyways so probably fine.

PR #1106, opened 2025-12-14
Add `tracing` instrumentation and dependencies across crates

crutcher:
### Instructions

### DIFF
--- a/crates/cubecl-core/src/compute/launcher.rs
+++ b/crates/cubecl-core/src/compute/launcher.rs
@@ -68,7 +68,6 @@ impl<R: Runtime> KernelLauncher<R> {
     /// - Contain any loops that never terminate. These may be optimized away entirely or cause
     ///   other unpredictable behaviour.
     #[track_caller]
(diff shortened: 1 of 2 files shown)


crates/cubecl-core/src/compute/launcher.rs
```diff
@@ -68,6 +68,7 @@ impl<R: Runtime> KernelLauncher<R> {
     /// - Contain any loops that never terminate. These may be optimized away entirely or cause
     ///   other unpredictable behaviour.
     #[track_caller]
+    #[tracing::instrument(level = "debug", skip(self, kernel, client))]
     pub unsafe fn launch_unchecked<K: CubeKernel>(
```
nathanielsimard, line 71: We could also add it under `launch`

nathanielsimard: But I'm wondering if we don't want all tracing calls in the client instead, which calls this function anyways.

crutcher: Pushing this down to `client.launch_inner`

PR #1115, opened 2025-12-19
feat: register external buffers

### LINKED ISSUE #291: Direct buffer passing
Hey yall I just found out about this project a few days ago, very cool stuff. Currently evaluating this for use in my team at Anduril. One of the main blockers tho is the lack of a easy way to directly pass in device buffers to kernels (I'm sure theres some not so straightforward ways) . There are several reasons to want to do this:

- You are using cubecl in conjunction with some other library, tool, langauge, etc that produces arrays/tensors that are already on the GPU. You'll have no choice but to copy it back to the host, and then have the runtime copy it back in order to call a kernel on it. 
- You want to use unified / pinned memory, GPU RDMA, etc for performance reasons. 

I've sketched out an implementation [here](https://github.com/tracel-ai/cubecl/compare/main...leizaf:cubecl:direct-buffer-passing) and I'm looking for feedback before I get too deep. I've basically just added an `ArrayArg` variant that takes the `Resource` of the associated `ComputeStorage`.  In the end you'll be able to mix it with normal cubecl arrays and it'll look something like this:
```rust
let device_buffer = CudaResource::from_device_pointer(...);

some_kernel::launch_unchecked::<...>(
    &client,
    CubeCount::Static(...),
    CubeDim::new(...),
    ArrayArg::from_raw_resource(device_buffer),
    ArrayArg::from_raw_parts(&normal_handle, ...),
);
```

mertalev:
This PR allows registering external buffers with CubeCL, which is very useful as it enables zero-copy transfer. It internally uses a `BindingMemory` enum to distinguish external vs managed buffers as suggested in #291. I tried to keep this PR as minimal and non-invasive as possible - DLPack integration and such is left for future work.

Fixes #291

### Instructions

### DIFF
--- a/crates/cubecl-runtime/src/memory_management/memory_pool/user_managed_pool.rs
+++ b/crates/cubecl-runtime/src/memory_management/memory_pool/user_managed_pool.rs
@@ -9,9 +9,9 @@ use hashbrown::HashMap;
 
 use super::{MemoryPool, Slice, SliceBinding, SliceHandle, SliceId};
 
-/// A memory pool for user-managed external buffers.
+/// A memory pool for user-managed external resources.
 ///
-/// When all references to a handle are dropped, the buffer is automatically
+/// When all references to a handle are dropped, the resource is automatically
 /// deallocated during cleanup.
 pub struct UserManagedPool {
     slices: HashMap<SliceId, Slice>,
@@ -24,9 +24,9 @@ impl UserManagedPool {
         }
     }
 
-    /// Register an external buffer.
+    /// Register an external resource.
     ///
-    /// The buffer will be deallocated from storage when all handle references
+    /// The resource will be deallocated from storage when all handle references
     /// are dropped and cleanup runs, or when explicitly released.
     pub(crate) fn register(&mut self, storage: StorageHandle) -> SliceHandle {
         let handle = SliceHandle::new();
@@ -35,11 +35,11 @@ impl UserManagedPool {
         handle
     }
 
-    /// Immediately unregister a buffer.
+    /// Immediately unregister a resource.
     ///
-    /// The caller must ensure all GPU operations using this buffer have completed before this call.
+    /// The caller must ensure all GPU operations using this resource have completed before this call.
     ///
-    /// Returns the storage handle if found, allowing the caller to retrieve the buffer.
+    /// Returns the storage handle if found, allowing the caller to retrieve the resource.
     pub(crate) fn unregister(&mut self, id: &SliceId) -> Option<StorageHandle> {
         self.slices.remove(id).map(|slice| slice.storage)
     }
--- a/crates/cubecl-wgpu/src/runtime.rs
+++ b/crates/cubecl-wgpu/src/runtime.rs
@@ -209,11 +209,13 @@ pub async fn init_setup_async<G: GraphicsApi>(
 /// - Any pending GPU operations on the buffer are complete before registration
 ///
 /// # Example
-/// ```ignore
-/// use cubecl_wgpu::{WgpuResource, register_external};
-///
-/// let resource = WgpuResource::new(my_buffer, 0, my_buffer.size());
-/// let handle = register_external(&device, resource, stream_id);
+/// ```no_run
+/// # use cubecl_wgpu::{WgpuResource, WgpuDevice, register_external};
+/// # use cubecl_common::stream_id::StreamId;
+/// # fn example(device: &WgpuDevice, my_buffer: wgpu::Buffer, stream_id: StreamId) {
+/// let resource = WgpuResource::new(my_buffer, 0, 64);
+/// let handle = register_external(device, resource, stream_id);
+/// # }
 /// ```
 pub fn register_external(
     device: &WgpuDevice,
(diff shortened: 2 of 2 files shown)


mertalev:
> The PR is good overall, but I have to think if it would be better to have a user managed memory pool instead. So all memory handles/bindings would be the same, but the underlying storage would be user managed. We already have different memory pools for different usage, so that could also work. Do you have an opinion about this?

I think either approach works, but the pool approach has the benefit that bindings and handles are simple and don't need branching whenever they're used. Let me give it a shot and see how it looks.

mertalev:
Yup, I think the pool version ends up being cleaner! One thing to note: I made it take ownership of the buffer since that's what the pool kind of leans toward with its `cleanup` etc., but I suppose dlpack integration would want borrow semantics for more flexibility. What do you think?

crates/cubecl-runtime/src/memory_management/memory_pool/user_managed_pool.rs
```diff
@@ -0,0 +1,106 @@
...
+
+use super::{MemoryPool, Slice, SliceBinding, SliceHandle, SliceId};
+
+/// A memory pool for user-managed external buffers.
+///
+/// When all references to a handle are dropped, the buffer is automatically
+/// unregistered during cleanup and deallocated from storage.
+pub struct UserManagedPool {
+    slices: HashMap<SliceId, Slice>,
+}
```
nathanielsimard, line 18: I think I would also allow manually deallocating a buffer on the user managed pool. That could be a lot faster than calling cleanup on all memory pools. It would be another way to dealloc a buffer, with fined grain control instead of relying on a "GC" like cleanup. Cleanup is not called often, since most of our pools don't need to deallocate, pretty much only when switching models to reset the memory pools allocations.

mertalev: True, I added an unregister API to allow this to let the caller either use or drop the underlying buffer.

crates/cubecl-wgpu/src/compute/mem_manager.rs
```diff
@@ -155,4 +155,29 @@ impl WgpuMemManager {
...
+    /// Ownership of the buffer is transferred to CubeCL. The buffer will be dropped
+    /// when released or when all references are dropped and cleanup runs.
+    pub(crate) fn register_external(
+        &mut self,
+        buffer: wgpu::Buffer,
+        stream_id: StreamId,
+    ) -> Handle {
+        let size = buffer.size();
+        let storage_handle = self.memory_pool.storage().register_external(buffer);
+        let slice_handle = self.memory_pool.register_external(storage_handle);
```
nathanielsimard: Registering the buffer in the storage should be done by the memory pool also. A single call to `register_external` should be necessary.

crates/cubecl-wgpu/src/compute/mem_manager.rs
```diff
@@ -155,4 +155,29 @@ impl WgpuMemManager {
...
+    }
+
+    /// Immediately unregister an external buffer.
+    ///
+    /// The caller must ensure all GPU operations using this buffer have completed before this call.
+    ///
+    /// Returns the buffer if found, allowing the caller to use or drop it.
+    pub(crate) fn unregister_external(&mut self, handle: &Handle) -> Option<wgpu::Buffer> {
+        let storage_handle = self.memory_pool.unregister_external(&handle.memory)?;
+        self.memory_pool.storage().take(&storage_handle)
```
nathanielsimard: Same here, `unregister_external` should return the storage resource rather than the storage handle.

crates/cubecl-wgpu/src/compute/server.rs
```diff
@@ -155,6 +155,29 @@ impl WgpuServer {
...
+    ///
+    /// Returns the buffer if found, allowing the caller to use or drop it.
+    pub fn unregister_external(
+        &mut self,
+        handle: &Handle,
+        stream_id: StreamId,
+    ) -> Option<wgpu::Buffer> {
+        let stream = self.scheduler.stream(&stream_id);
+        stream.mem_manage.unregister_external(handle)
+    }
```
nathanielsimard: For completness, those functions could be included in the ComputeServer trait and return/receive `ComputeStorage::Resource` instead. That would allow to have a unified API to register external buffers for all cubecl runtimes.

crates/cubecl-wgpu/src/compute/mem_manager.rs
mertalev: This code is in cubecl-wgpu while the memory pool is defined in cubecl-runtime, which has no wgpu dependency. It seems hard to make it a single call without bleeding the types.

crates/cubecl-wgpu/src/compute/server.rs
mertalev: I made this change, but only wgpu is usable in this PR - the rest will panic with `unimplemented!`

crates/cubecl-wgpu/src/compute/mem_manager.rs
mertalev: Decoupling the API from wgpu with `ComputeStorage::Resource` allowed me to turn it into a single call.

somethingelseentirely:
Anything we can do to help push this forward it would be very nice to have for our metal edge deployments, because we could just mmap from storage directly into the GPU 😅

nathanielsimard:
The memory management changed a lot since that PR was made. But the strategy of registering a buffer in a new memory pool should be used.

antimora:
Stale

PR #979, opened 2025-10-23
Chore/wgpu v27

laggui:
Forked from #939

Additional changes:
- Fix binding alignment for vulkan
- Update w/ main

## Validate your PR with burn.

It is important that you make sure that you don't introduce any bugs in burn. 

### Instructions

- [ ] Create a new branch or fork of the [burn repo](https://github.com/Tracel-AI/burn)
- [ ] Update the main [Cargo.toml](https://github.com/tracel-ai/burn/blob/40aa993fb5969351a006b560ec89da5119dc9721/Cargo.toml#L159=L161) with this PR hash.
- [ ] Fix any broken tests or compilation errors in burn.
- [ ] Submit a PR in burn with your fixes and link it here.

PR #1119, opened 2025-12-22
Upgrade to wgpu v28

laggui:
- Changes from v27 in https://github.com/tracel-ai/cubecl/pull/979
- Migrate error scope usage ([now use guards and are thread-local](https://github.com/gfx-rs/wgpu/blob/trunk/CHANGELOG.md#error-scopes-now-use-guards-and-are-thread-local))
- Migrate `enumerate_adapters` ([now async and avail on WebGPU](https://github.com/gfx-rs/wgpu/blob/trunk/CHANGELOG.md#wgpuinstanceenumerate_adapters-is-now-async--available-on-webgpu))
- Migrate `wgpu::PollType::Wait` ([now a struct w/ submission index and timeout fields](https://github.com/gfx-rs/wgpu/blob/trunk/CHANGELOG.md#wgpupolltypewait-has-now-an-optional-timeout))
- Migrate `subgroup_{min,max}_size` ([renamed + now in `AdapterInfo` not limits](https://github.com/gfx-rs/wgpu/blob/trunk/CHANGELOG.md#subgroup_minmax_size-renamed-and-moved-from-limits---adapterinfo))
- Updated MSRV to 1.92 (latest stable, but required for wgpu v28)
- Patched `sysinfo` (used in cubecl-cpu, required for burn) to be compatible with `windows` crate version for wgpu (otherwise we get [compilation errors in burn](https://github.com/tracel-ai/burn/actions/runs/20438464279/job/58725773343))

### Instructions

### DIFF
--- a/crates/cubecl-core/src/runtime_tests/index.rs
+++ b/crates/cubecl-core/src/runtime_tests/index.rs
@@ -40,6 +40,42 @@ pub fn test_kernel_index_scalar<R: Runtime, F: Float + CubeElement>(client: Comp
     assert_eq!(actual[2], F::new(123.0));
 }
 
+#[cube(launch)]
+fn shuffle_kernel(orders: &mut Array<u32>) {
+    if UNIT_POS < 2 {
+        let mut order = Array::<u32>::new(2usize);
+        order[0] = 0;
+        order[1] = 1;
+
+        let i = ABSOLUTE_POS & 1;
+        let tmp = order[i];
+        order[i] = 9;
+        order[i ^ 1] = tmp;
+
+        let base = ABSOLUTE_POS * 2;
+        orders[base] = order[0];
+        orders[base + 1] = order[1];
+    }
+}
+
+// Regression test for invalid `CopyTransform`
+pub fn test_kernel_shuffle<R: Runtime>(client: ComputeClient<R>) {
+    let handle = client.empty(4 * size_of::<u32>());
+
+    shuffle_kernel::launch::<R>(
+        &client,
+        CubeCount::Static(1, 1, 1),
+        CubeDim::new_1d(2),
+        unsafe { ArrayArg::from_raw_parts::<u32>(&handle, 4, 1) },
+    )
+    .unwrap();
+
+    let actual = client.read_one(handle);
+    let actual = u32::from_bytes(&actual);
+
+    assert_eq!(actual, [9, 0, 1, 9]);
+}
+
 #[allow(missing_docs)]
 #[macro_export]
 macro_rules! testgen_index {
@@ -52,6 +88,12 @@ macro_rules! testgen_index {
             cubecl_core::runtime_tests::index::test_kernel_index_scalar::<TestRuntime, FloatType>(
                 client,
             );
+        }
+
+        #[$crate::runtime_tests::test_log::test]
+        fn test_shuffle() {
+            let client = TestRuntime::client(&Default::default());
+            cubecl_core::runtime_tests::index::test_kernel_shuffle::<TestRuntime>(client);
         }
     };
 }
--- a/crates/cubecl-ir/src/variable.rs
+++ b/crates/cubecl-ir/src/variable.rs
@@ -169,6 +169,17 @@ impl Variable {
         )
     }
 
+    /// Is this an array type that is contained in concrete memory,
+    /// or a local array/scalar/vector?
+    pub fn is_memory(&self) -> bool {
+        matches!(
+            self.kind,
+            VariableKind::GlobalInputArray { .. }
+                | VariableKind::GlobalOutputArray { .. }
+                | VariableKind::SharedArray { .. }
+        )
+    }
+
     pub fn has_length(&self) -> bool {
         matches!(
             self.kind,
--- a/crates/cubecl-opt/src/passes/index_merge.rs
+++ b/crates/cubecl-opt/src/passes/index_merge.rs
@@ -21,7 +21,7 @@ impl OptimizerPass for CopyTransform {
                 let inst = ops.borrow()[idx].clone();
                 match &inst.operation {
                     Operation::Operator(Operator::Index(op))
-                        if op.list.is_array()
+                        if op.list.is_memory()
                             && op.list.ty == inst.ty()
                             && !is_reused(opt, &inst.out) =>
                     {
@@ -30,7 +30,7 @@ impl OptimizerPass for CopyTransform {
                         }
                     }
                     Operation::Operator(Operator::IndexAssign(op))
-                        if inst.out().is_array() && inst.ty() == op.value.ty =>
+                        if inst.out().is_memory() && inst.ty() == op.value.ty =>
                     {
                         if let Some(id) = as_versioned(&op.value) {
                             writes.insert(id, (idx, inst.out(), op.index));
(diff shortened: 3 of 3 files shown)


crates/cubecl-wgpu/src/compute/storage.rs
```diff
@@ -36,12 +36,15 @@ pub struct WgpuResource {
 impl WgpuResource {
     /// Return the binding view of the buffer.
     pub fn as_wgpu_bind_resource(&self) -> wgpu::BindingResource<'_> {
+        #[cfg(feature = "spirv")]
```
ArthurBrussee: This might need a comment why this is ok, and that it does break vulkan robustness

crates/cubecl-wgpu/src/device.rs
```diff
@@ -76,62 +76,45 @@ impl Device for WgpuDevice {
...
+            ..Default::default()
+        });
+        let adapters = enumerate_all_adapters(instance);
+        adapters.len()
     }
 }
+
+fn enumerate_all_adapters(instance: wgpu::Instance) -> Vec<wgpu::Adapter> {
+    // `enumerate_adapters` is now async & available on WebGPU
+    cubecl_common::future::block_on(instance.enumerate_adapters(wgpu::Backends::all()))
```
ArthurBrussee, line 138: block_on however, does _not_ work on WebGPU. WebGPU already has a special async creation path. I think the best bet is to save the device_count etc. (maybe just a Vec<AdapterInfo>?) on creation, which can wait on this async.

crates/cubecl-wgpu/src/runtime.rs
```diff
@@ -356,32 +357,41 @@ async fn request_adapter(
...
         #[cfg(not(target_family = "wasm"))]
-        WgpuDevice::DiscreteGpu(num) => select_from_adapter_list(
-            num,
-            "No Discrete GPU device found",
-            &instance,
-            &device,
-            backend,
-        ),
+        WgpuDevice::DiscreteGpu(num) => {
+            select_from_adapter_list(
```
ArthurBrussee, line 360: This might work on WebGPU now right? So one could allow these overrides in theory?

crates/cubecl-wgpu/src/device.rs
laggui: Whoops :sweat_smile: thanks for catching that lol.

Was just going through breaking changes to make the compiler happy first, but totally missed that.

We could save the device count on init, but does WebGPU in the browser typically expose more than one adapter?  You probably know better than I do for this point.

Otherwise, we could stick to the previous WASM cfg where we always return a fixed device count of 1.

crates/cubecl-wgpu/src/runtime.rs
laggui: For consistency, perhaps. But is more than 1 adapter typically exposed (same question as above)?

Thanks for the comments btw :pray:

laggui:
@ArthurBrussee I kept the current behavior fixed to 1 device, but lmk if supporting more than one adapter might still make sense.

Everything else has been updated, will run some tests.

crates/cubecl-wgpu/src/device.rs
ArthurBrussee: Yep you're right, it's still only 1 so not much too it. I thought the new wgpu feature ment WebGPU exposed this now but no, they just async request "the" adapter and return a vec of that.

crates/cubecl-wgpu/src/runtime.rs
ArthurBrussee: Basically but looking into it, there is one subtetly here, which is the adapter "power preference". It could be worth explicitly setting this to low when selecting an integrated GPU, and maybe explicitly to high in all other cases.

Cargo.toml
```diff
@@ -131,3 +131,7 @@ opt-level = 0
 
 [profile.dev.build-override]
 opt-level = 3
+
+# TODO: remove once sysinfo release includes windows 0.62 (required for wgpu)
+[patch.crates-io]
+sysinfo = { git = "https://github.com/GuillaumeGomez/sysinfo", rev = "b0ae22bfb2064d74e9561fb71890ad68763c7aff" }
```
nathanielsimard: We can't merge until all our dependencies are on crates.io, otherwise it blocks any release we want to do.

crates/cubecl-wgpu/src/compute/storage.rs
```diff
@@ -36,12 +36,21 @@ pub struct WgpuResource {
...
     pub fn as_wgpu_bind_resource(&self) -> wgpu::BindingResource<'_> {
+        // wgpu enforces 4-byte alignment for buffer binding sizes per the WebGPU spec.
+        // - https://github.com/gfx-rs/wgpu/pull/8041
+        //
+        // This padding is safe because:
+        // 1. In checked mode, bounds checks prevent reading beyond the logical size.
+        // 2. In unchecked mode, OOB access is already undefined behavior.
+        #[cfg(feature = "spirv")]
+        let size = self.size.next_multiple_of(4);
+        #[cfg(not(feature = "spirv"))]
```
nathanielsimard: @wingertge do you think we should have a different behavior here with the SPIRV compiler?

wingertge: I think it should be padded to the next multiple regardless of target, since WGSL now also supports `f16` (and so does Metal). If I'm understanding this code correctly.

crates/cubecl-wgpu/src/runtime.rs
laggui: Done!

Cargo.toml
laggui: I know.. this might be blocked until then. Had to apply this patch otherwise it wouldn't compile on windows targets due to conflicting `windows` types (see this [build output](https://github.com/tracel-ai/burn/actions/runs/20438464279/job/58725773343)).

wgpu requires `windows` 0.62 (used for DX12), and `sysinfo` uses the win APIs as well so it pulls `windows` but the latest release currently requires `>=0.59, <0.62`.

thomsen85:
@laggui 'sysinfo' v0.38.0 released, it uses `windows` 0.62. Would love to see this get merged :)

laggui:
> @laggui 'sysinfo' v0.38.0 released, it uses `windows` 0.62. Would love to see this get merged :)

Thanks for flagging! Will get this updated

crates/cubecl-wgpu/src/backend/base.rs
```diff
@@ -45,8 +42,8 @@ impl WgpuServer {
                 log::trace!("Using SPIR-V cache");
 
                 let repr = AutoRepresentationRef::SpirV(&entry.kernel);
-                let module = self.create_module(&entry.entrypoint_name, Some(repr), "", mode);
-                let pipeline = self.create_pipeline(&entry.entrypoint_name, Some(repr), module)?;
+                let module = self.create_module(&entry.entrypoint_name, Some(repr), "", mode)?;
+                let pipeline = self.create_pipeline(&entry.entrypoint_name, Some(repr), module);
```
laggui, line 46: Given the new local error scope, I changed `create_module` to return the error instead of `create_pipeline` (which were recently split into separate methods)

PR #1121, opened 2025-12-28
refactor: Constants

wingertge:
Small refactor to constants in the IR, to make code cleaner and allow for more effective inlining of constant lines.

Instead of combining the kind and value into the enum, makes the value typeless and uses the `Variable::ty` field to carry the type information. This allows using lined and other more complex types, and also simplifies handling since most use cases were ignoring the kind anyways, and when the type is needed it's already available from the `ty` field on the Variable.
Constants are now converted to the target type on construction, and the now redundant sanitize code is removed.

Also fixes some non-fatal validation errors with Vulkan atomics.

### Instructions

### DIFF
--- a/README.md
+++ b/README.md
@@ -233,7 +233,7 @@ You can even ship the autotune cache with your program, reducing cold start time
 
 ## Resource
 
-For now we don't have a lot of resources to learn, but you can look at the [matrix multiplication library](/crates/cubecl-matmul/README.md) to see how CubeCL can be used.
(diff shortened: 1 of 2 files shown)


crates/cubecl-ir/src/branch.rs
```diff
@@ -115,114 +115,3 @@ pub struct RangeLoop {
 pub struct Loop {
     pub scope: Scope,
 }
-
-impl If {
```
nathanielsimard, line 119: Is there a reason we don't need that now?

wingertge: It was from the ancient pre-compiler version of CubeCL, so it hadn't been used for a long time. It was for the old assembly macro.

PR #1129, opened 2026-01-05
Feat/comptime device props

nathanielsimard:
Allow kernels to fetch the device properties at compile-time, removing the need to include them in the compilation key, which is wasteful. Kernels are JIT compiled for a single specific device.

### DIFF
--- a/crates/cubecl-core/src/frontend/container/line/ops.rs
+++ b/crates/cubecl-core/src/frontend/container/line/ops.rs
@@ -8,7 +8,7 @@ use crate::{
 };
 use crate::{
     frontend::{
-        Abs, ArcCos, ArcCosh, ArcSin, ArcSinh, ArcTan, ArcTanh, Ceil, Clamp, Cos, Cosh,
+        Abs, ArcCos, ArcCosh, ArcSin, ArcSinh, ArcTan, ArcTan2, ArcTanh, Ceil, Clamp, Cos, Cosh,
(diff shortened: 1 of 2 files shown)


crates/cubecl-core/src/compute/builder.rs
```diff
@@ -156,6 +161,7 @@ impl KernelBuilder {
         } else {
             debug == 1
         };
+        println!("Here");
```
louisfd: remove print

crates/cubecl-wgpu/src/compiler/wgsl/compiler.rs
```diff
@@ -85,6 +85,7 @@ impl WgslCompiler {
         mut value: kernel::KernelDefinition,
         mode: ExecutionMode,
     ) -> Result<wgsl::ComputeShader, CompilationError> {
+        println!("Compile shader?");
```
louisfd: remove print

crates/cubecl-wgpu/src/compiler/wgsl/compiler.rs
```diff
@@ -115,6 +116,7 @@ impl WgslCompiler {
 
         self.metadata = Metadata::new(num_meta as u32, num_ext);
 
+        println!("Hash pro {}", value.body.properties.is_some());
```
louisfd: remove print

louisfd (changes requested):
Doesn't compile on metal (i can look into it). Also remove some prints

PR #1131, opened 2026-01-08
refactor: Close gaps between CubeCL and standard Rust

wingertge:
# Changes

* Traits now properly accept method syntax, even when the type is generic
* Atomics now accept method syntax
* Adds `to_bits` and `from_bits` to float types (alias to `Reinterpret::reinterpret`)

# Breaking Changes

* `log` renamed to `ln` for consistency with Rust's standard library
* Atomic functions renamed to be consistent with Rust atomics
* `bitwise_not` removed in favor of the existing `Not` trait
* Removed `Min`, `Max`, and `Clamp`. Replaced them with `Ord` and a set of freestanding functions that work for any `PartialOrd`. This is what `num_traits` does, and three of the new functions are actually just expansions of the `num_traits` functions.

### Instructions

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,13 +24,17 @@ log = { default-features = false, version = "^0.4.22" }
 
 tokio = { version = "^1.47.1" }
 
+test-log = { version = "^0.2", default-features = false }
+
+tracing = { version = "^0.1.43", default-features = false }
+tracing-opentelemetry = "^0.32.0"
+tracing-subscriber = "^0.3.20"
+
 opentelemetry = "^0.31.0"
 opentelemetry-aws = "^0.19.0"
 opentelemetry-otlp = "^0.31.0"
 opentelemetry_sdk = "^0.31.0"
-tracing = { version = "^0.1.43", default-features = false }
-tracing-opentelemetry = "^0.32.0"
-tracing-subscriber = "^0.3.20"
+
 
 bincode = { version = "2.0.1", features = [
     "alloc",
--- a/PERFORMANCE.md
+++ b/PERFORMANCE.md
@@ -1,0 +1,64 @@
+# Performance Hacking
+
+## Tracing
+
+Most of the crates support the "tracing" feature, which enables the [tracing](https://crates.io/crates/tracing) crate.
+
+When the feature is enabled, some critical path code is instrumented with tracing spans:
+
+```rust
+#[cfg_attr(
+    feature = "tracing",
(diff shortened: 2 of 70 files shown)


crates/cubecl-core/src/frontend/operation/binary.rs
```diff
@@ -231,13 +324,135 @@ macro_rules! impl_binary_func_mixed_types {
...
                 }
-            })*
+            }
+        }
+    }
+}
+
+macro_rules! impl_core_binop {
+    ($trait: ident, $method: ident, $op: expr) => {
+        paste::paste! {
```
nathanielsimard, line 343: It would be cool at some point to remove `paste::paste`. simpler macros are a win.

crates/cubecl-core/src/frontend/operation/binary.rs
```diff
@@ -288,56 +503,6 @@ impl_binary_func!(
     f32,
     f64
 );
-impl_binary_func!(
```
nathanielsimard, line 291: Why we deleted that? I guess it wasn't necessary anymore?

crates/cubecl-core/src/frontend/operation/binary.rs
wingertge: paste specifically makes them simpler, especially since expand methods have a fixed name that can't be freely changed

crates/cubecl-core/src/frontend/operation/binary.rs
wingertge: They're now handled by `Ord` (for integers), the float-specific functions (for floats), as well as the free-standing functions (for generic numeric values).

wingertge:
> I think we should take the opportunity during the refactor to add more documentation on the internals, specifically why the new traits exist and who should implement them.

Are you talking about the new `CubeAdd` etc traits? I think I'd like to eventually add a page to the contributor guide on that, it's just cause you can't add associated methods to primitive types without adding a new trait. And they shouldn't be implemented by end users, they're for extending primitives.

This honestly isn't a finished refactor, but the remaining part will require much more development so I wanted to just do the easy part and get it out quickly. I'd like to eventually replace as many traits as possible with an expand version of the `num_traits` traits. There's just some hang-ups that I need to think about how to handle properly. The main thought behind this and the remaining refactor is that wherever possible, CubeCL should create expansions for existing Rust ecosystem functions, rather than creating an alternate version.

PR #1133, opened 2026-01-09
Lift/Test `valid_strides` into a layout validation lib.

crutcher:
- work out what contract `valid_strides` was enforcing.
- write tests for that
- lift into a `Result<(), StrideError>` mechanic.
- provide a migration function.
- lift into `cubecl-zspace`.

Previously discussed with @nathanielsimard was a plan to create a public no-deps common `zspace` module; this is an incremental step towards that. I don't see use in poisoning the crate namespace until we have something with a solid surface; but we need to start pulling the parts together.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -13,6 +13,14 @@ on:
         default: false
 
 jobs:
+  publish-cubecl-zspace:
+    uses: tracel-ai/github-actions/.github/workflows/publish-crate.yml@v5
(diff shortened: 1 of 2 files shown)


nathanielsimard (commented):
The only missing part is to add the possibility to deploy the crate. The file that needs to be updated is `.github/publish.yml`

crutcher:
Rebased; and fixed `.github/publish.yml`

PR #1134, opened 2026-01-12
fix: Store packed dimension on packed quant stores so it can be correctly handled for swapped tensors

wingertge:
Stores the packed dimension on packed quantized tensors. This is required to correctly handle swaps/permutations for certain cases (i.e. a 4-wide matrix).
The packed dim is stored as an inverse index. While this is somewhat confusing, it's important because it allows us to always initialize it to `0` by default, and makes things like unsqueezing batch dims easier. It also matches the block size, which applies to the last `N` dimensions.

Also refactors debug representation of kernel args because it was bothering me too much while I was debugging an issue with block scaled matmul. Core properties that are always present, like address type and `CubeDim`, are now stored directly on the `KernelId` struct. Macro-generated kernels also generate an info struct, so the info field is properly typed and associates parameters with their corresponding function argument. I believe this also fixes the long-standing issue of too many arguments causing errors around missing implementations for tuples of size > 16.

### Instructions


crates/cubecl-macros/src/generate/kernel.rs
```diff
@@ -441,4 +454,95 @@ impl Launch {
             TokenStream::new()
         }
     }
+
+    fn info_ty(&self, name: &Ident) -> proc_macro2::TokenStream {
```
nathanielsimard, line 458: I'm a bit confused, do we need that for the kernel id to work? I though most of the info were in the `self.info` field

wingertge: that's what it goes into, it ensures info is properly formatted with names instead of as an anonymous tuple, along with fixing the problem that traits are only implemented for tuples up to a certain size.

PR #1138, opened 2026-01-13
Propagate lint rules to crates; fix outstanding violations.

crutcher:
```terminaloutput
cargo clippy --fix --allow-staged --allow-dirty
```
Did most of the cleanup after the `Cargo.toml` rules were propagated.


crutcher:
Rebase.

PR #1143, opened 2026-01-14
CudaServer::change_server_serialized simultaneous Commands.

crutcher:
Interleave src/dst Commands / Event timelines.
This is in prep for interleaved ring buffer copies.

### DIFF
--- a/crates/cubecl-cuda/src/compute/server.rs
+++ b/crates/cubecl-cuda/src/compute/server.rs
@@ -613,7 +613,9 @@ impl CudaServer {
         let handle_dst = command_dst.reserve(binding.size())?;
         let resource_dst = command_dst.resource(handle_dst.clone().binding())?;
 
-        // stream_src waits until the stream_dst buffer is sequenced.
+        // Since the `cpu_buffer` lives in the destination stream timeline,
+        // we ensure that the allocation of that copy buffer is complete
+        // in both timelines before we permit the source stream to proceed.
         Fence::new(stream_dst).wait_async(stream_src);
 
         // TODO: Interleave


crates/cubecl-cuda/src/compute/server.rs
```diff
@@ -584,76 +583,73 @@ impl CudaServer {
...
+                elem_size,
+                &cpu_buffer,
+                resource_dst.ptr,
+                stream_dst,
+            )
+        }?;
+
+        // GC the buffer when the write completes.
+        let event = Fence::new(stream_dst);
+        command_dst.streams.gc(GcTask::new(cpu_buffer, event));
```
nathanielsimard: Could we allocate the cpu buffer with the destination command instead to avoid that as it was done before?

crutcher: It was always gc'd. What changed here is that I unrolled the the `unsafe { write_to_gpu(...) }` wrapper; because I'm working toward slice rights; and that wrapper was scheduling the gc.

It is semantically more correct to allocate on the source stream, because it is the source stream that needs a sequence point on that allocation before performing any buffer copies.

It doesn't matter *much* right now, because those allocations are synchronous; but in prep for moving towards making them async, we should remove these semantic warts.

crutcher: For reference, see the relevant portion of `Command::write_to_gpu()`:
https://github.com/tracel-ai/cubecl/blob/main/crates/cubecl-cuda/src/compute/command.rs#L350-L354

nathanielsimard: `bytes` was never registered to the GC task list because the buffer is owned by the destination pinned memory pool. Since this pool enforces strict task ordering, the buffer can be reused immediately after the 'to-device' copy is registered without concurrency issues. This safety is guaranteed by the CUDA runtime’s task queue order. If we use the source pinned memory pool instead, we must wait for the destination execution queue to complete the copy before releasing the buffer. This creates unnecessary latency and resource waste compared to the destination-owned approach.

crutcher: I moved this over to dst; but I left the fence in, as it is part of the GC process in the `Command::write_to_gpu()` api, and is also a requirement for `GcTask` interface.

crates/cubecl-cuda/src/compute/server.rs
```diff
@@ -584,76 +583,74 @@ impl CudaServer {
...
+                &cpu_buffer,
+                resource_dst.ptr,
+                stream_dst,
+            )
+        }?;
+
+        // GC the buffer when the write completes.
+        // Make sure we don't reuse the pinned memory until the write to gpu is completed.
+        let event = Fence::new(stream_dst);
+        command_dst.streams.gc(GcTask::new(cpu_buffer, event));
```
nathanielsimard: That is no longer necessary, since we can reuse the same cpu buffer even if the copy isnt finish. The ordering is enforced, if you queue 5 copy in a row on the same dst server, they can all reuse the same cpu buffer.

nathanielsimard: Having the GcTask would trigger 5 more allocations instead, which would slow down the copy and increase memory usage.

crutcher: If what you're telling me about the pool allocation is correct (and I've not traced through all of that code), then we need a cross-stream fence somewhere:
* Either, we allocate the buffer in the src pool; and we fence gc on an event in the dst pool,
* Or, we allocate the buffer in the dst pool, and we fence the src>buf write on the dst pool timeline.

Now, many fences we could setup are free; if the timeline (from some other utility algorithm) already lines up a stream, then the fence becomes basically a no-op.

If we want to allocate the buf on dst, and drop it at the end of this code block; we should set a fence to make src sync on the dst stream after allocating the dst buffer; to ensure that the buffer is available in the dst timeline before src overwrites it (and by that reading, the current code is probably buggy).

Do you agree with this analysis?

crates/cubecl-cuda/src/compute/server.rs
```diff
@@ -584,76 +583,112 @@ impl CudaServer {
...
+        let mut cpu_buffer = command_dst.reserve_cpu(num_bytes, true, None);
 
-        // We create a command on the source server to retrieve the correct resource from the
-        // source memory pools. We also make sure the current stream is aligned with the stream of
-        // the binding, where the data was first allocated.
+        let handle_dst = command_dst.reserve(binding.size())?;
+        let resource_dst = command_dst.resource(handle_dst.clone().binding())?;
+
+        // stream_src waits until the stream_dst buffer is sequenced.
+        Fence::new(stream_dst).wait_async(stream_src);
```
nathanielsimard, line 619: Why do we need that sync?

crates/cubecl-cuda/src/compute/server.rs
```diff
@@ -584,76 +583,112 @@ impl CudaServer {
...
+                &mut cpu_buffer,
                 resource_src.ptr,
                 stream_src,
             )?;
         }
-        let fence_src = Fence::new(stream_src);
 
-        // We need to free the command before creating another one.
+        // stream_dst waits until the stream_src write is sequenced.
+        Fence::new(stream_src).wait_async(stream_dst);
```
nathanielsimard, line 668: Yes this one is required.

nathanielsimard (commented):
I'm unsure about the first sync, but other than that it's quite clear now!

crates/cubecl-cuda/src/compute/server.rs
crutcher: Added a comment explaining this.

crutcher:
@nathanielsimard can you take another look, I rebased and added a comment.

PR #1151, opened 2026-01-20
Fix MLIR pass ordering and add CPU barrier support

jguhlin:
I needed some changes to support some models on CPU (yes, I know it's very slow; HPC was down for the weekend, and I wanted to confirm a fix in a different repo was working, so I needed a portable kernel).

Fixes MLIR lowering failure when `scf.if` yields memref results, and adds barrier/sync_cube support for the CPU backend.

### MLIR Pass Ordering Fix
When `scf.if` yields a memref, `scf-to-cf` creates `cf.cond_br` with memref-typed block arguments. If `cf-to-llvm` runs before memrefs are converted, verification fails with llvm.cond_br operand must be LLVM dialect-compatible

Fixed by reordering passes to run `finalize-memref-to-llvm` and `reconcile-unrealized-casts` between `scf-to-cf` and `cf-to-llvm`.

### CPU Barrier Support
- Map cube-level barrier ops to `sync_cube()`
- Fix `sync_cube()` spin-loop with dedicated `BARRIER_TARGET`
- Add `rsqrt`/`rsqrtf` external functions for math-to-libm lowering
- Configurable worker stack size via `CUBECL_CPU_STACK_SIZE` / `CUBECL_CPU_STACK_MB`

### Also fixes
- Buffer position lookups in metadata visitor
- `LocalConst` variable handling

## Test Plan
- [x] MLIR regression test: `it_lowers_scf_if_memref_result_through_cf_to_llvm`
- [x] Runtime test: `test_line_cf` (Line<F> in conditionals)
- [x] Barrier tests: `barrier_smoke`, `sync_cube_magic`, `sync_cube_two_phase`, `sync_cube_all_reduce`
- [x] All 508 `cubecl-cpu` tests pass

### DIFF
--- a/crates/cubecl-core/src/runtime_tests/line.rs
+++ b/crates/cubecl-core/src/runtime_tests/line.rs
@@ -269,7 +269,9 @@ macro_rules! testgen_line {
         #[$crate::runtime_tests::test_log::test]
         fn test_line_conditional() {
             let client = TestRuntime::client(&Default::default());
-            cubecl_core::runtime_tests::line::test_line_conditional::<TestRuntime, FloatType>(client);
+            cubecl_core::runtime_tests::line::test_line_conditional::<TestRuntime, FloatType>(
+                client,
+            );
         }
 
         #[$crate::runtime_tests::test_log::test]


crates/cubecl-core/src/runtime_tests/line.rs
```diff
@@ -112,6 +112,57 @@ pub fn test_line_loop_unroll<R: Runtime, F: Float + CubeElement>(client: Compute
     }
 }
 
+#[cube(launch_unchecked)]
+pub fn kernel_line_cf<F: Float>(
```
wingertge: This name is hard to understand without knowing the MLIR compiler internals, so it should probably have a more descriptive name since it's in the general test suite.

jguhlin:
Sorry for for the delay. Our HPC was down for a few days

PR #1160, opened 2026-01-23
Dependency tweaks for building on Android

metasim:
A couple `Cargo.toml` tweaks to handle modern Android targets.

---

### Instructions

- [X] Create a new branch or fork of the [burn repo](https://github.com/Tracel-AI/burn)
- [X] Update the main [Cargo.toml](https://github.com/tracel-ai/burn/blob/40aa993fb5969351a006b560ec89da5119dc9721/Cargo.toml#L159=L161) with this PR hash.
- [X] Fix any broken tests or compilation errors in burn.
- [X] Submit a PR in burn with your fixes and link it here. ← No fixes necessary

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -58,9 +58,7 @@ serde = { version = "1.0.204", default-features = false, features = [
 serde_bytes = { version = "0.11.17", default-features = false, features = [
     "alloc",
 ] } # alloc for no_std
-serde_json = { version = "1.0.119", default-features = false, features = [
-    "alloc",
-] } # alloc for no_std
+serde_json = { version = "1.0.119", default-features = false }
 toml = "0.9"
 variadics_please = "1"
 
--- a/crates/cubecl-runtime/build.rs
+++ b/crates/cubecl-runtime/build.rs
@@ -3,8 +3,8 @@ use cfg_aliases::cfg_aliases;
 fn main() {
     // Setup cfg aliases
     cfg_aliases! {
-        // Some features like autotune caching, compilation caching, and config loading rely on files/environment variables,
-        // which are only applicable on "standard desktop platforms".
+        // Some features like autotune caching, compilation caching, and config loading
+        // require std with OS-level filesystem and environment access.
         std_io: { all(feature = "std", any(target_os = "windows", target_os = "linux", target_os = "macos", target_os = "android")) },
         exclusive_memory_only: { any(feature = "exclusive-memory-only", target_family = "wasm") },
         multi_threading: { all(feature = "std", not(target_family = "wasm")) },


crates/cubecl-runtime/Cargo.toml
```diff
@@ -61,7 +61,7 @@ web-time = { workspace = true }
 thiserror = { workspace = true }
 
 # Persistent cache deps - has to match the cfg(std_io) cfg.
-[target.'cfg(any(target_os = "windows", target_os = "linux", target_os = "macos"))'.dependencies]
+[target.'cfg(any(target_os = "windows", target_os = "linux", target_os = "macos", target_os = "android"))'.dependencies]
```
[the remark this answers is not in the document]
metasim: ~Human reviewer: Is this legit? If so, I'll commit the suggestion.~
I'll apply this and fix the merge conflicts shortly.

crates/cubecl-runtime/build.rs
```diff
@@ -5,7 +5,7 @@ fn main() {
     cfg_aliases! {
         // Some features like autotune caching, compilation caching, and config loading rely on files/environment variables,
         // which are only applicable on "standard desktop platforms".
```
laggui: Guess that comment regarding "standard desktop platforms" is no longer accurate 😅 we could clarify e.g.

```rust
// Some features like autotune caching, compilation caching, and config loading
// require std with OS-level filesystem and environment access.
```

Cargo.toml
```diff
@@ -58,7 +58,9 @@ serde = { version = "1.0.204", default-features = false, features = [
 serde_bytes = { version = "0.11.17", default-features = false, features = [
     "alloc",
 ] } # alloc for no_std
-serde_json = { version = "1.0.119", default-features = false }
+serde_json = { version = "1.0.119", default-features = false, features = [
+    "alloc",
+] } # alloc for no_std
```
metasim: This _technically_ might not be required for this PR, but I think it's a bug... before adding `target_os = "android"` to the other specifications (i.e. the build assumed android was `no_std`), I got a compilation error that was fixed by adding "alloc" here as with the other serde dependencies.

laggui: Actually, it might be better to enable the `"alloc"` feature only at the crate level that pulls `serde_json` instead (if it is indeed a missing feature).

We also removed some unconditional std feature propagation in https://github.com/tracel-ai/cubecl/pull/1184, and looks like there is an outstanding PR to make some crates truly `no_std`.

Can you validate where this is required, and fixed the comment mentioned above? :pray:

metasim: @laggui 

> Can you validate where this is required,

I tried, but I don't yet know enough about the architecture to feel confident in figuring this out, so I just removed the change. I hope that's sufficient.

laggui: Ahhh I see, I thought you made this change specifically because you were hitting a compilation error that required the change.

Thanks for reverting :slightly_smiling_face:

PR #1164, opened 2026-01-28
feat: Add explicit resource errors

wingertge:
CUDA has extremely opaque errors, so any error when requesting too many resources will simply show `CUDA_ERROR_INVALID_VALUE`. This adds a new error type for resource limit to make the errors there more readable, and to make it possible to handle the error gracefully (i.e. for matmul).

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -14,7 +14,7 @@ on:
 
 jobs:
   publish-cubecl-zspace:
-    uses: tracel-ai/github-actions/.github/workflows/publish-crate.yml@v5
+    uses: tracel-ai/github-actions/.github/workflows/publish-crate.yml@v8
     with:
       crate: cubecl-zspace
       dry-run-only: ${{ github.event_name == 'workflow_dispatch' && inputs.dry-run-only || false }}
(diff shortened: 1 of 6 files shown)


crates/cubecl-hip/src/compute/context.rs
```diff
@@ -349,4 +357,51 @@ impl HipContext {
             }
         }
     }
+
+    fn validate_shared(&self, repr: &Option<HipComputeKernel>) -> Result<(), LaunchError> {
```
nathanielsimard, line 364: Those validation functions could be shared by all runtimes since they only use hardware properties no? It could be in `cubecl-runtime`, at a similar place where pitch tensor layout is implemented.

wingertge: Shared memory is handled differently by each backend, so non-trivial to unify. Though maybe the compiled kernel should have it as a property, since all backends do implement it in one way or another. I'll try that one.
But cube dim and units can definitely be shared.

wingertge: Nvm there's no good way to get shared memory onto the compiled kernel without complicating the current abstraction. So shared memory validation will stay in each backend, but the others can be shared.

PR #1166, opened 2026-01-30
feat: Chunked binary compilation cache

wingertge:
Migrates compilation caches to a much more space and performance-efficient binary format, and also implements (currently manual) chunking support, so kernels can be loaded only where they're needed. There is no automated chunking at the moment, but this implements the infrastructure needed for tooling to implement various chunking strategies.

The cache consists of two parts:
* A single monolithic table of contents. This only contains the cache key and the path of the chunk the kernel is saved in.
* A set of chunks that are loaded on demand and inserted into an in-memory cache. Each chunk is loaded into memory at once, but different chunks can be loaded only when needed.

Reads will try to fetch the entry from the in-memory cache first, then look at the TOC to see if it's cached in an unloaded chunk. If it is, then that chunk is loaded.

# Performance
Performance is identical at worst, noticably better at best. Size of the cache is 50% smaller for SPIR-V, 70% smaller for PTX (which was 20x the size of SPIR-V to begin with).

### Instructions

### DIFF
--- a/crates/cubecl-macros/src/generate/launch.rs
+++ b/crates/cubecl-macros/src/generate/launch.rs
@@ -97,7 +97,12 @@ impl Launch {
             let address_type = prelude_type("AddressType");
 
             let kernel_doc = format!(
-                "Launch the kernel [{}()] on the given runtime",
+                "Launch the kernel [{}()] on the given runtime without bound checks.\n\n\
+                 # Safety\n\n\
+                 The kernel must not:\n\
+                 - Contain any out of bounds reads or writes. Doing so is immediate UB.\n\
+                 - Contain any loops that never terminate. These may be optimized away entirely or cause\n\
+                   other unpredictable behaviour.",
                 self.func.sig.name
             );
             let generics = &self.launch_generics;


crates/cubecl-common/src/compilation_cache.rs
```diff
@@ -0,0 +1,223 @@
...
+
+type InMemoryCache<K, V> = RefCell<HashMap<K, Box<V>>>;
+
+/// A chunked cache for compilation artifacts. Uses a human readable table of contents, with binary
+/// storage for the compiled kernel.
+#[derive(Debug)]
+pub struct CompilationCache<K: CacheKey, V: CacheValue> {
+    toc: Cache<K, String>,
+    // Box ensures values aren't moved when inserting new elements, so we don't need to keep it
+    // locked for reads
```
nathanielsimard: There is no box?

nathanielsimard: The `Cache` type also has an `in_memory_cache` is there a reason we have another one in the `CompilationCache`?

crates/cubecl-runtime/src/id.rs
```diff
@@ -332,6 +334,7 @@ trait DynKey: core::fmt::Debug + Send + Sync {
     fn dyn_type_id(&self) -> TypeId;
     fn dyn_eq(&self, other: &dyn DynKey) -> bool;
     fn dyn_hash(&self, state: &mut dyn Hasher);
+    fn dyn_hash_one(&self) -> StableHash;
```
nathanielsimard, line 337: Is there a reason that method is important? Maybe we could add docs to this trait.

crates/cubecl-common/src/compilation_cache.rs
wingertge: The comment should be moved to the type alias, good point.
The `in_memory_cache` here is for the kernels. `Cache` is hardcoded to a JSON cache.
It's only used here for the table of contents, since it's small enough to store in JSON and it's nice to have them readable.
The `in_memory_cache` is a writable map that gets filled with the kernels from chunks on demand.

crates/cubecl-runtime/src/id.rs
wingertge: Just a bit of cleanup to make the code more readable. And a way to get the hash without wrapping it in another hasher, but that's not used in the current version. So right now it's just cleanup.

This trait is implemented automatically and not used outside of `KernelId`, so don't think this necessarily needs a doc.

PR #1175, opened 2026-02-04
feat: metal backend

dcvz:
I benchmarked the native backend against running the same MSL through wgpu. One caveat to keep the comparison honest: wgpu only ever runs fast math, while native defaults to **precise** (IEEE-accurate). So the default comparison is precise-vs-fast, and even with that handicap native wins on every bandwidth-bound kernel:

| kernel (f32) | native metal | wgpu‹msl› | faster by |
|---|---:|---:|:---:|
| small copy (4K) | 0.21 ms | 1.33 ms | **6.3×** |
| small add (1M) | 0.25 ms | 1.33 ms | **5.3×** |
| big copy (33M) | 2.88 ms | 3.81 ms | **1.3×** |
| big add (33M) | 4.51 ms | 4.94 ms | **1.1×** |
| heavy cos/sin (33M) | 57.2 ms | 17.3 ms | **3.3× slower** |

The only row that flips is the heavy transcendental kernel — accurate `cos`/`sin` simply cost more than fast approximations. However, fast math is configurable per-op, so you can opt into it wherever the extra accuracy isn't worth it. If we compare fast vs fast, native is ahead everywhere:

| kernel (f32) | native metal | wgpu‹msl› | faster by |
|---|---:|---:|:---:|
| small copy (4K) | 0.12 ms | 1.33 ms | **11×** |
| small add (1M) | 0.24 ms | 1.33 ms | **5.5×** |
| big copy (33M) | 2.41 ms | 3.81 ms | **1.6×** |
| big add (33M) | 3.53 ms | 4.94 ms | **1.4×** |
| heavy cos/sin (33M) | 14.7 ms | 17.3 ms | **1.2×** |


crates/cubecl-metal/src/runtime.rs
```diff
@@ -0,0 +1,211 @@
...
+        "metal"
+    }
+
+    fn max_cube_count() -> (u32, u32, u32) {
+        (u32::MAX, u32::MAX, u32::MAX)
+    }
+
+    fn can_read_tensor(_shape: &Shape, _strides: &Strides) -> bool {
+        true
+    }
```
AdrianEddy: This needs to check the tensor layout:

```rust
    fn can_read_tensor(shape: &Shape, strides: &Strides) -> bool {
        if shape.is_empty() {
            return true;
        }

        // Check that strides match contiguous (row-major) layout
        let rank = shape.len();
        let mut expected_stride = 1usize;
        for i in (0..rank).rev() {
            if strides[i] != expected_stride {
                return false;
            }
            expected_stride *= shape[i];
        }

        true
    }
```

dcvz: Good point! updated!

crates/cubecl/Cargo.toml
```diff
@@ -47,7 +47,7 @@ cuda = ["cubecl-cuda"]
 cuda-ptx-wmma = ["cubecl-cuda?/ptx-wmma"]
 hip = ["cubecl-hip"]
 hip-rocwmma = ["cubecl-hip?/rocwmma"]
-metal = ["wgpu-msl"]
+metal = ["cubecl-metal"]
```
louisfd: This is a big silent swap for everyone using `--features metal`. I think we should keep metal = ["wgpu-msl"] and expose the native one as `metal-native` until it's well tested, then flip the default, perhaps a few weeks from now. I would start using `metal-native` personally to test it.

crates/cubecl-metal/src/compute/server.rs
```diff
@@ -0,0 +1,576 @@
...
+            return Box::pin(async move { Err(e) });
+        }
+
+        let results: Result<Vec<_>, ServerError> = descriptors
+            .iter()
+            .map(|descriptor| {
+                let (resource, offset) = resolve_origin_resource(&mut resolved, &descriptor.handle)
+                    .map_err(ServerError::from)?;
+
+                let size: usize = descriptor.shape.iter().product();
```
louisfd: This reads a flat block and never looks at `descriptor.strides` (same in write), which is wrong for pitched/padded layouts that can_read_tensor says we support. 
Cuda does a real pitched copy (`read_async`/`write_to_gpu` in `cubecl-cuda/src/compute/command.rs`). I guess most tests are contiguous so they may not catch it.

crates/cubecl-metal/src/compute/server.rs
```diff
@@ -0,0 +1,576 @@
...
+            {
+                Ok(r) => r,
+                Err(e) => return Box::pin(async move { Err(e) }),
+            };
+
+        // Flush, wait, then read.
+        let stream = resolved.current();
+        let event = MetalStreamBackend::flush(stream);
+
+        if let Err(e) = MetalStreamBackend::wait_event_sync(event) {
```
louisfd, line 250: If the buffer faults here, the event still signals and we hand back garbage as Ok? Maybe check how cuda handles it.

crates/cubecl-metal/src/runtime.rs
```diff
@@ -0,0 +1,240 @@
...
+        let mem_props = MemoryDeviceProperties {
+            max_page_size: (*metal_device).maxBufferLength() as u64,
+            alignment: 256,
+        };
+
+        let hardware_props = HardwareProperties {
+            load_width: 128,
+            plane_size_min: 32,
+            plane_size_max: 32,
+            max_bindings: 31,
```
louisfd: Generally we must have one less (30) because of the info buffer every kernel has.

louisfd (changes requested):
Before we merge this, could you make a PR in [cubek](https://github.com/tracel-ai/cubek) based on it? That will allow me to test more thoroughly. I also have a few other comments (flagged by claude tbh).

dcvz:
@louisfd I've addressed the comments! Regarding a `cubek` branch: there's not much new besides having to point to my branch:

```
cubecl = { git = "https://github.com/oxideai/cubecl", branch = "feat/native-metal-backend", default-features = false }
cubecl-common = { git = "https://github.com/oxideai/cubecl", branch = "feat/native-metal-backend", default-features = false }
```

and running benches with:
```
cargo bench -p benchmarks --features cubecl/metal-native --bench unary
```

louisfd (commented):
I'm getting No Metal device found on my M2 Pro (macOS 13.4). default_device() uses MTLCreateSystemDefaultDevice(), which returns None for me, while MTLCopyAllDevices() works fine.

I think maybe we could fall back to the first device from all_devices():
```rust
pub fn default_device() -> Option<Retained<ProtocolObject<dyn MTLDevice>>> {
      objc2_metal::MTLCreateSystemDefaultDevice()
          .or_else(|| all_devices().into_iter().next())
}
```

PR #1176, opened 2026-02-06
fix: Make TMA errors return a result instead of panicking

wingertge:
Fixes https://github.com/tracel-ai/burn/issues/4445 by making invalid TMA configs return a launch error instead of panicking. Only validate on error, instead of on every launch, since the validation is fairly expensive.

### DIFF
--- a/crates/cubecl-cuda/src/compute/server.rs
+++ b/crates/cubecl-cuda/src/compute/server.rs
@@ -26,7 +26,7 @@ use cubecl_runtime::{
     compiler::CubeTask,
     config::GlobalConfig,
     logging::ServerLogger,
-    memory_management::{MemoryAllocationMode, MemoryUsage, offset_handles, should_optimize},
+    memory_management::{MemoryAllocationMode, MemoryUsage, offset_handles, optimal_align},
     server::{self, ComputeServer},
     storage::BindingResource,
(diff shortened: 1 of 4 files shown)


crates/cubecl-cuda/src/compute/server.rs
```diff
@@ -747,84 +757,113 @@ fn oob_to_cuda(fill: OobFill) -> CUtensorMapFloatOOBfill {
...
+    ($lhs: expr, $rhs: expr, $($arg:tt)+) => {
+        if $lhs == $rhs {
+            Ok(())
+        } else {
+            Err(LaunchError::Unknown {
+                reason: format!($($arg)*),
+                backtrace: BackTrace::capture(),
+            })
+        }
+    };
```
nathanielsimard: It's a detail, but we could improve the macro:

```rust
macro_rules! launch_check {
    ($assertion: expr, $($arg:tt)+) => {
        if $assertion {
            Ok(())
        } else {
            Err(LaunchError::Unknown {
                reason: format!($($arg)*),
                backtrace: BackTrace::capture(),
            })
        }
    };
}

Also I don't think we need 2 macros, just calling `launch_check` with `lhs == rhs` is good enough.

PR #1187, opened 2026-02-13
perf: Improve performance of `MetadataBuilder`

wingertge:
Improves performance of `MetadataBuilder` with two optimizations:
* Reuse the staging buffers inside `MetadataBuilder` using `thread_local`
* Specialize the state on address type so `finish` can just be a `memcpy` instead of iterating through every element.

This gives a relative peformance improvement of ~60-70% for `MetadataBuilder::finish`.

### Instructions

### DIFF
--- a/crates/cubecl-core/src/compute/launcher.rs
+++ b/crates/cubecl-core/src/compute/launcher.rs
@@ -1,9 +1,12 @@
-use core::cell::RefCell;
 use std::{collections::BTreeMap, marker::PhantomData};
 
 use crate::prelude::{ArrayArg, TensorArg, TensorMapArg, TensorMapKind};
 use crate::{CubeScalar, KernelSettings};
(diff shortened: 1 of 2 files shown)


crates/cubecl-core/src/compute/launcher.rs
```diff
@@ -101,15 +102,19 @@ impl<R: Runtime> KernelLauncher<R> {
     }
 }
 
+thread_local! {
+    static METADATA: RefCell<MetadataBuilder> = RefCell::new(MetadataBuilder::default());
+}
```
nathanielsimard, line 111: I don't think this work on wasm, is there a way to put that behind a feature flag?

wingertge: Ah I forgot `thread_local` requires std. I can gate it, yeah.

PR #1188, opened 2026-02-13
feat: Add support for `VK_EXT_shader_long_vector`

wingertge:
Not yet in a stable driver so can't test it, but this should work.


antimora:
Stale?

wingertge:
> Stale?

Only in the sense that I'm still waiting for a driver that includes it

wingertge:
I got tired of waiting for Nvidia to get this on stable so I tested with the beta driver. That driver crashes when encountering `OpSwitch` but all the other `cubecl` tests pass, and I've inspected the SPIR-V to ensure it's generating the large vectors correctly. So I think this is ready to merge.

nathanielsimard (approved):

PR #1190, opened 2026-02-17
refactor: Metadata

wingertge:
Refactors metadata to make `Strides` opaque and migrate `Shape` to `cubecl-zspace` so it can be shared. Uses `SmallVec` with an inline size of 5, matching the hardcoded 5D limit of TMA on Nvidia cards. This covers almost every ML op, and has relatively little overhead for 99% of all cases (most tensors are 4D in modern architectures, both for vision and attention).

This also includes the changes to `supported_line_sizes` from https://github.com/tracel-ai/cubecl/pull/1188, but not the actual implementation of long vectors since that can't be tested yet.

### Instructions


nathanielsimard (approved):
LGTM

PR #1197, opened 2026-02-25
feat: add exp2 and log2 operations

jeff-hiner:
Inside tight loops for inference engines that do stuff like sigmoid functions, it's often more efficient to precalculate a constant to multiply, and then use `exp2`/`log2` instead of `exp`. This is because the `2` variants generally only require some very simple math on the exponent. (Some Stable Diffusion implementations use this trick, and I want to mirror it.)

Add Exp2 and Log2 as first-class operations in cubecl-ir, mirroring the existing Exp and Log operations. Wire them through all backends:
- cubecl-core: trait impls for all float types, Float supertrait bound
- cubecl-ir: Arithmetic enum variants with Display
- cubecl-opt: visitor and constant propagation support
- cubecl-spirv: GLSLstd450 Exp2/Log2 extension instructions
- cubecl-wgpu: exp2()/log2() WGSL builtins

This PR is purely additive, and doesn't introduce any issues in burn. (But I did fork it and check.)


crates/cubecl-core/src/frontend/element/float/typemap.rs
```diff
@@ -238,6 +238,7 @@ impl<const POS: u8> Magnitude for ElemExpand<POS> {}
 impl<const POS: u8> Recip for ElemExpand<POS> {}
 impl<const POS: u8> Erf for ElemExpand<POS> {}
 impl<const POS: u8> Exp for ElemExpand<POS> {}
+impl<const POS: u8> Exp2 for ElemExpand<POS> {}
```
louisfd: This will need to be updated

crates/cubecl-core/src/frontend/element/float.rs
```diff
@@ -23,6 +23,7 @@ pub trait Float:
     Numeric
     + FloatOps
     + Exp
+    + Exp2
     + Log
     + Log1p
```
louisfd, line 26: Is there a reason Log2 was left out here

crates/cubecl-core/src/frontend/element/float/typemap.rs
louisfd: Is there a reason Log2 was left out here

crates/cubecl-wgpu/src/compiler/wgsl/instructions.rs
```diff
@@ -640,10 +648,18 @@ impl Display for Instruction {
                 let out = out.fmt_left();
                 writeln!(f, "{out} = exp({input});")
             }
+            Instruction::Exp2 { input, out } => {
```
louisfd, line 652: It should also be implemented for the cpp compiler

louisfd (changes requested):

antimora:
Stale?

jeff-hiner:
Not abandoned, I've just been a bit busy with some personal matters the last couple weeks. Give me a moment.

jeff-hiner:
I believe everything has been resolved, but I'm not as familiar with the codebase as the owners. Is everything good here, or does it need more work?

nathanielsimard (commented):
I think the only missing piece is the unit tests. I believe they are in `cubecl-core/src/runtime_tests`

jeff-hiner:
I wasn't able to find any matching unit tests for `exp` or `log` in `cubecl-core/src/runtime_tests`, but I'm trying to keep this lined up for a potential merge.

PR #1199, opened 2026-02-27
Refactor device communication channel

nathanielsimard:
The goal of this PR is to change the way the client communicate with the server. Since the server is mutable, but not the client we need some way of given mut access to the server. Previously we used lock, which is fine, but I wanted the possibility to use a channel instead, which might be better in some scenarios. However the current design made it impossible, since for most functions, we needed to **wait** for the server return type, which break entirely the lazy channel execution style we want. So I refactored a lot fo allow true async.

CubeK PR: 
Burn PR:

### DIFF
--- a/crates/cubecl-common/src/device/handle/tests_recursive.rs
+++ b/crates/cubecl-common/src/device/handle/tests_recursive.rs
@@ -5,13 +5,14 @@ fn test_recursive_execution_same_state() {
         type_id: 0,
         index_id: 5,
     };
-    let context = DeviceHandle::<TestDeviceState<1>>::new(device_id);
-    let context_cloned = context.clone();
+    let handle = DeviceHandle::<TestDeviceState<1>>::new(device_id);
+    let handle_cloned = handle.clone();
 
-    let _count = context
+
+    let _count = handle
         .submit_blocking(move |state| {
             state.counter += 1;
-            context_cloned.submit(move |state| {
+            handle_cloned.submit(move |state| {
                 state.counter += 1;
             });
         })
--- a/crates/cubecl-core/src/compute/launcher.rs
+++ b/crates/cubecl-core/src/compute/launcher.rs
@@ -238,7 +238,7 @@ impl<R: Runtime> TensorState<R> {
         };
 
         let elem_size = array.elem_size * *vectorization;
-        let buffer_len = array.handle.size() / elem_size as u64;
+        let buffer_len = array.handle.size_in_used() / elem_size as u64;
         let address_type = self.address_type();
         self.with_metadata(|meta| {
             meta.register_array(


Cargo.toml
```diff
@@ -27,6 +27,7 @@ invalid_html_tags = "warn"
 bitflags = { version = "2.9.1", features = ["serde"] }
 derive-new = { version = "0.7.0", default-features = false }
 derive_more = { version = "2", default-features = false }
+oneshot = { version = "0.1.13", default-features = false }
```
wingertge: This is an old version, should probably use 0.2

crates/cubecl-core/src/frontend/container/array/launch.rs
```diff
@@ -22,18 +22,18 @@ pub struct ArrayCompilationArg {
 impl CompilationArg for ArrayCompilationArg {}
 
 /// Tensor representation with a reference to the [server handle](cubecl_runtime::server::Handle).
-pub struct ArrayHandleRef<'a, R: Runtime> {
-    pub handle: &'a cubecl_runtime::server::Handle,
+pub struct ArrayHandleRef<R: Runtime> {
```
wingertge: Should this still be called `Ref`?

crates/cubecl-runtime/src/server/base.rs
```diff
@@ -470,42 +478,44 @@ pub enum AllocationKind {
...
+    pub fn contiguous(shape: Shape, elem_size: usize) -> Self {
+        MemoryLayoutDescriptor::new(MemoryLayoutStrategy::Contiguous, shape, elem_size)
     }
 }
 
 /// An allocation with associated strides. Strides depend on tensor layout.
-#[derive(Debug)]
-pub struct Allocation {
+#[derive(Debug, Clone)]
+pub struct MemoryLayout<R: Runtime> {
```
wingertge, line 505: This is not a great name, it's not just a layout, it's a handle plus layout.

crates/cubecl-runtime/src/stream/event.rs
```diff
@@ -28,13 +27,18 @@ pub trait EventStreamBackend: 'static {
 
     /// Initializes and returns a new stream associated with the given stream ID.
     fn create_stream(&self) -> Self::Stream;
+    /// Returns the cursor of the given handle on the given stream.
+    fn handle_cursor(stream: &Self::Stream, handle: &Binding) -> u64;
+    /// Returns wheter the stream can access new tasks.
+    fn is_healty(stream: &Self::Stream) -> bool;
```
wingertge: typo, `healty` instead of `healthy`

crates/cubecl-runtime/src/stream/event.rs
```diff
@@ -206,20 +212,29 @@ impl<B: EventStreamBackend> MultiStream<B> {
     pub fn resolve<'a>(
         &mut self,
         stream_id: StreamId,
-        bindings: impl Iterator<Item = &'a Binding>,
-    ) -> ResolvedStreams<'_, B> {
-        let analysis = self.align_streams(stream_id, bindings);
+        handles: impl Iterator<Item = &'a Binding>,
+        enfore_healty: bool,
```
wingertge: typo

crates/cubecl-runtime/src/stream/event.rs
```diff
@@ -332,147 +361,148 @@ impl<B: EventStreamBackend> core::fmt::Debug for StreamWrapper<B> {
...
-        }
-
-        fn wait_event(_stream: &mut Self::Stream, _event: Self::Event) {}
-
-        fn wait_event_sync(_event: Self::Event) -> Result<(), ExecutionError> {
-            Ok(())
-        }
-    }
-}
+// TODO: Fix the tests
```
wingertge: TODO/dead code

crates/cubecl-runtime/src/server/base.rs
```diff
@@ -237,121 +258,107 @@ pub enum ExecutionError {
...
+    #[error("An execution error happened during profiling\nCaused by:\n  {0}")]
+    Profile(#[from] ProfileError),
+
+    /// An execution error happened during profiling
+    #[error("An execution error happened during profiling\nCaused by:\n  {0}")]
+    Io(#[from] IoError),
+
+    /// The server is an invalid state.
+    #[error("The server is in an invalid state\nCaused by:\n  {reason}")]
+    ServerUnHealty {
```
wingertge: `Unhealthy` should be a single word

crates/cubecl-common/src/device/handle/tests.rs
```diff
@@ -0,0 +1,99 @@
...
+// 
+//     let _count = context
+//         .submit_blocking(move |state| {
+//             state.counter += 1;
+//             context_cloned.submit(move |state| {
+//                 state.counter += 1;
+//             });
+//         })
+//         .unwrap();
+// }
```
wingertge: Dead code

crates/cubecl-common/src/device/handle/mod.rs
```diff
@@ -0,0 +1,105 @@
...
+
+    include!("./tests.rs");
+}
+
+// #[cfg(test)]
+// mod tests_mutex {
+//     type DeviceHandle<S> = mutex::MutexDeviceHandle<S>;
+//
+//     include!("./tests.rs");
+// }
```
wingertge: Dead code

wingertge (commented):
Leftover `profile.json.gz` in `cubecl-common`

crates/cubecl-runtime/src/server/base.rs
nathanielsimard: It's not an allocation either, nothing has been allocated, it creates a handle to memory that will exist at some point with a given layout.

PR #1252, opened 2026-03-26
Trying some buffer count validation

louisfd:

### Instructions

### DIFF
--- a/crates/cubecl-wgpu/src/compute/server.rs
+++ b/crates/cubecl-wgpu/src/compute/server.rs
@@ -169,6 +169,7 @@ impl WgpuServer {
         self.scheduler.logger.log_compilation(&compiled);
 
         self.validate_shared(&compiled.repr)?;
+        self.validate_binding_numbers(&compiled.repr)?;
 
         // /!\ Do not delete the following commented code.
         // This is useful while working on the metal compiler.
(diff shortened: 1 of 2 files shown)


crates/cubecl-wgpu/src/compute/server.rs
```diff
@@ -237,6 +238,34 @@ impl WgpuServer {
...
+    ) -> Result<(), LaunchError> {
+        let max = self.utilities.properties.hardware.max_bindings;
+
+        let num_storage_buffers = repr.as_ref().map(|repr| match repr {
+            AutoRepresentation::Wgsl(repr) => repr.buffers.len(),
+            #[cfg(feature = "msl")]
+            AutoRepresentation::Msl(repr) => repr.buffers.len(),
+            #[cfg(feature = "spirv")]
+            AutoRepresentation::SpirV(repr) => repr.bindings.len(),
+        });
```
wingertge, line 254: I think it would be easier to pass the kernel bindings in and do the check before the compilation, then there's no need to match on the representation. We need them to match the number of buffers in the kernel anyways, since they're used to fetch the wgpu resources.

crates/cubecl-wgpu/src/compute/server.rs
```diff
@@ -237,6 +238,34 @@ impl WgpuServer {
...
+
+        if let Some(num_storage_buffers) = num_storage_buffers
+            && num_storage_buffers > max as usize
```
wingertge, line 257: I think this leaves out the info buffer, so that would need to be counted separately if `info.data` isn't empty.

louisfd:
Closing as stale

PR #1259, opened 2026-03-31
Adds arena + refactor stream id

nathanielsimard:
This PR introduces two related changes:

`StreamId::executes`: a safe wrapper around the raw `StreamId::swap` pairs scattered across ChannelDeviceHandle. It uses a drop guard to restore the previous stream on both normal return and unwind, eliminating three unsafe blocks at the call sites.

arena module (behind std feature): a circular, allocation-free pool for reusable memory blocks.


laggui (approved):

Matheus-Garbelini:
Is there a high level example on how to use streaming with wgpu?

PR #1284, opened 2026-04-10
Fix performance regression rocm

nathanielsimard:
Broadcast select operations to per-element ternaries on native types when vectorized, avoiding branching that the HIP compiler emits for struct-level ternaries.


crates/cubecl-core/src/frontend/operation/branch.rs
```diff
@@ -26,10 +26,12 @@ pub fn select_many<C: Scalar, N: Size>(
     then: Vector<C, N>,
     or_else: Vector<C, N>,
 ) -> Vector<C, N> {
-    intrinsic!(|scope| select::expand(scope, condition.expand.into(), then, or_else))
+    intrinsic!(|scope| { { select::expand(scope, condition.expand.into(), then, or_else) } })
```
wingertge: Don't think these extra brackets are supposed to be here

PR #1286, opened 2026-04-13
Feat/vector sum

### LINKED ISSUE #1285: Vector Sum Instruction
Add a dedicated instruction to reduce a SIMD vector to a single scalar value via summation. 
Targeting CPU backends where wide vectorization makes this especially valuable, but implement it for every compiler.

nathanielsimard:
Add new vector sum instruction

fix #1285


louisfd (approved):
LGTM

PR #1289, opened 2026-04-13
Fix one case of unsoundness, and two other potential bugs.

ArthurBrussee:
Three independent bugs found by running miri against the test suite, each with a minimal regression test:

Miri UB:

`Task::init<F>` in the channel task module now validates both `size_of::<F>()` and `align_of::<F>()` when routing closures between inline, arena, and boxed storage. The arena buffer was also bumped from `Vec<u128>` to `Vec<SlotBuffer>` where `SlotBuffer` is `#[repr(C, align(64))]`, so every arena slot is 64-byte aligned and matches the inline slot's alignment. Closures with over-aligned captures previously produced a misaligned `ptr::write`, which is UB as caught by miri.

Two other issues found along the way:

1. `Bytes::try_enforce_runtime_align` now checks the controller's reported `alloc_align()` instead of the raw pointer's alignment offset. The raw pointer could happen to be aligned while the controller still reported a lower alignment, causing downstream `try_into_vec::<E>` to wrongly reject valid conversions.

2. `ChannelDeviceState::init` serializes on a per-key `spin::Mutex` so that `S::init` runs exactly once per `(DeviceId, TypeId)`. The previous check-drop-work-lock pattern let concurrent callers each construct their own `S` instance; the losers were immediately dropped.

Each fix has a minimal regression test.

### DIFF
--- a/crates/cubecl-common/src/device/handle/channel.rs
+++ b/crates/cubecl-common/src/device/handle/channel.rs
@@ -461,8 +461,8 @@ mod task {
         /// `ptr::write` (UB). The boxed fallback uses `Box::new`, whose allocation
         /// satisfies any alignment.
         pub fn init<F: FnOnce() -> TaskResult + Send + 'static>(&mut self, func: F) {
-            let fits_inline =
-                size_of::<F>() <= size_of::<SmallTaskData>() && align_of::<F>() <= INLINE_SLOT_ALIGN;
+            let fits_inline = size_of::<F>() <= size_of::<SmallTaskData>()
+                && align_of::<F>() <= INLINE_SLOT_ALIGN;
             let fits_arena =
                 size_of::<F>() <= size_of::<LargeTaskData>() && align_of::<F>() <= ARENA_SLOT_ALIGN;
 
@@ -1137,9 +1137,7 @@ mod tests {
         // Mirror the 64-aligned 4KB region that `TaskBuffer::new` allocates per slot.
         #[repr(C, align(64))]
         struct TestArena([u128; GLOBAL_TASK_MAX_SIZE / 16]);
-        let mut arena = alloc::boxed::Box::new(TestArena(
-            [0u128; GLOBAL_TASK_MAX_SIZE / 16],
-        ));
+        let mut arena = alloc::boxed::Box::new(TestArena([0u128; GLOBAL_TASK_MAX_SIZE / 16]));
         let arena_ptr = arena.0.as_mut_ptr() as *mut u8;
         let mut task = Task::new(arena_ptr);
 
@@ -1165,9 +1163,7 @@ mod tests {
 
         #[repr(C, align(64))]
         struct TestArena([u128; GLOBAL_TASK_MAX_SIZE / 16]);
-        let mut arena = alloc::boxed::Box::new(TestArena(
-            [0u128; GLOBAL_TASK_MAX_SIZE / 16],
-        ));
+        let mut arena = alloc::boxed::Box::new(TestArena([0u128; GLOBAL_TASK_MAX_SIZE / 16]));
         let arena_ptr = arena.0.as_mut_ptr() as *mut u8;
         let mut task = Task::new(arena_ptr);
 
--- a/crates/cubecl-common/src/device/handle/channel.rs
+++ b/crates/cubecl-common/src/device/handle/channel.rs
@@ -407,21 +407,15 @@ mod task {
 
     /// The maximum size of a closure that can be stored without heap allocation.
     pub const GLOBAL_TASK_MAX_SIZE: usize = 4096;
+
     /// The maximum size of a closure that can be stored using inlined memory.
     const INLINE_TASK_MAX_SIZE: usize = 48;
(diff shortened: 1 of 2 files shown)


crates/cubecl-common/src/device/handle/channel.rs
```diff
@@ -411,6 +420,14 @@ mod task {
     pub type LargeTaskData = [u128; GLOBAL_TASK_MAX_SIZE / 16];
     pub type SmallTaskData = [u128; INLINE_TASK_MAX_SIZE / 16];
```
nathanielsimard: Should that be updated to `SlotBuffer` ?

crates/cubecl-common/src/device/handle/channel.rs
```diff
@@ -689,20 +716,38 @@ mod custom_channel {
...
+    #[repr(C, align(64))]
+    #[derive(Clone, Copy)]
+    struct SlotBuffer {
+        data: [u128; GLOBAL_TASK_MAX_SIZE / 16],
+    }
+
+    const _: () = {
+        assert!(core::mem::size_of::<SlotBuffer>() == GLOBAL_TASK_MAX_SIZE);
+        assert!(core::mem::align_of::<SlotBuffer>() == 64);
+    };
```
nathanielsimard: That's a good way to ensure const properties!

crates/cubecl-common/src/device/handle/channel.rs
```diff
@@ -689,20 +716,38 @@ mod custom_channel {
...
     }
 
+    /// A single arena slot. `#[repr(C, align(64))]` makes every slot 64-byte aligned,
+    /// matching the inline task slot so the router can store closures with
+    /// `align_of::<F>()` up to 64 in either location. `GLOBAL_TASK_MAX_SIZE` is a
+    /// multiple of 64, so there is no per-slot padding.
+    #[repr(C, align(64))]
+    #[derive(Clone, Copy)]
+    struct SlotBuffer {
+        data: [u128; GLOBAL_TASK_MAX_SIZE / 16],
```
nathanielsimard: If we use `repr(C, align(64))` we don't need to use `u128 / 16`, we can use directly `u8`

crates/cubecl-common/src/device/handle/channel.rs
```diff
@@ -255,8 +255,12 @@ struct ChannelService {
...
-    spin::Mutex::new(None);
+/// Per-key init slot. Concurrent callers for the same `(DeviceId, TypeId)` pair
+/// serialize on the inner `spin::Mutex`, ensuring `S::init` runs exactly once per key.
+#[allow(clippy::type_complexity)]
+static CHANNELS: spin::Mutex<
+    Option<HashMap<(DeviceId, TypeId), alloc::sync::Arc<spin::Mutex<Option<ChannelDeviceState>>>>>,
+> = spin::Mutex::new(None);
 
 impl ChannelDeviceState {
     pub fn init<S: DeviceService>(
```
nathanielsimard, line 265: So now there are always 2 locks when we try to retrieve a channel device handle. Is it because we can have recursive locking if one initialization requires another service to be initialized? 

Also I'm unsure if the second mutex has to be wrapped in an Arc, should it be clonable?

ArthurBrussee: Yeah was a bit overkill, just meant you could add stuff without contending other backends. That's not really needed though, I've simplified it to a single lock again, which is now just held for longer to prevent the potential double insert race.

crates/cubecl-common/src/device/handle/channel.rs
ArthurBrussee: Sweet, done!

crates/cubecl-common/src/device/handle/channel.rs
ArthurBrussee: Renamed some things here and made it a bit simpler!

crates/cubecl-common/src/device/handle/channel.rs
```diff
@@ -398,18 +401,30 @@ mod task {
...
+    /// Closures with `align_of::<F>() <= SLOT_ALIGN` fit in either slot without
+    /// falling through to the boxed path.
+    pub const SLOT_ALIGN: usize = 64;
+
+    /// One arena slot. `#[repr(C, align(SLOT_ALIGN))]` makes every slot 64-byte
+    /// aligned on its own, so the slot alignment does not depend on the layout of any
+    /// enclosing type. `GLOBAL_TASK_MAX_SIZE` is a multiple of 64, so there is no
+    /// per-slot padding.
+    #[repr(C, align(64))]
+    #[derive(Clone, Copy)]
```
nathanielsimard: I don't think we want to make this `Copy` it's a large type

nathanielsimard: I also think I would not derive Clone. It's not safe to clone the data of a slot anyway without involving the inner type representation's Clone function.

ArthurBrussee: Oh my bad yeah we really shouldn't & don't need this, fixed.

nathanielsimard (approved):

PR #1290, opened 2026-04-13
Fix and refactor all_reduce

Charles23R:
- Fixed a but where not flushing `cudarc::nccl::result::comm_init_rank` calls could result in a deadlock.
- Added logging for `DeviceService`.
- Generalize the `CommunicationId` concept.

### Instructions

https://github.com/tracel-ai/burn/pull/4740


crates/cubecl-runtime/src/client.rs
```diff
@@ -606,12 +613,34 @@ impl<R: Runtime> ComputeClient<R> {
...
+            Some(is_init) => !*is_init,
+            None => true,
+        };
+        let is_comms_init = if query_server {
+            self.device
+                .submit_blocking(move |server| server.is_comms_init(&comms_id))
+                .unwrap()
+        } else {
+            true
+        };
```
Charles23R, line 630: This is not the cleanest, my next PR will decouple "communication initialization" into a new function on the `ComputeServer` trait. This way we don't have to deal with the initialization in the actual collective operation

crates/cubecl-runtime/src/server/base.rs
```diff
@@ -398,6 +399,28 @@ where
     fn allocation_mode(&mut self, mode: MemoryAllocationMode, stream_id: StreamId);
 }
 
+/// An ID unique to any unordered combination of devices.
+#[derive(Clone, Debug, Hash, Eq, PartialEq)]
+pub struct CommunicationId {
+    /// The ID as a `String`.
+    pub str_id: String,
```
nathanielsimard: Making String clonable can be expensive. Is it possible to use an u64 for the ID? If not we can wrap the string into and Arc.

crates/cubecl-runtime/src/client.rs
```diff
@@ -26,13 +27,15 @@ use cubecl_zspace::Shape;
...
 use cubecl_common::stream_id::StreamId;
+use hashbrown::HashMap;
 
 /// The `ComputeClient` is the entry point to require tasks from the `ComputeServer`.
 /// It should be obtained for a specific device via the Compute struct.
 pub struct ComputeClient<R: Runtime> {
     device: DeviceHandle<R::Server>,
     utilities: Arc<ServerUtilities<R::Server>>,
     stream_id: Option<StreamId>,
+    is_comms_init: HashMap<CommunicationId, bool>,
```
nathanielsimard: Cloning an hashmap can be expensive. That map could be in `utilities` and wrapped into a RW Lock

nathanielsimard (changes requested):

nathanielsimard (approved):

PR #1291, opened 2026-04-14
Fix/tuner group

nathanielsimard:
Working on an improved tuner for mamtul in burn, I've found a bug in the tuning priority algo


louisfd (approved):
LGTM

PR #1292, opened 2026-04-14
Simplify async tuning

ArthurBrussee:
Too much of tuning was async, all the execution of the tuning _has_ to be synchronous to be safe. We only need to async _wait_ on the results of the tuning. This simplifies the code a bit to move the readback to an explicit process request.

Additionally, since that spawned future requires locks anyway we can remove one lock in the SharedStateMap. This then simplifies the loop() where we handle the tuning result greatly as we dont have to worry about read/write deadlocks anymore.

Also remove an unsafe impl of Send/Sync for keys, instead we can just prove this with rust bounds.


crates/cubecl-runtime/src/tune/local.rs
```diff
@@ -141,112 +128,71 @@ where
...
+            self.checks(&operations, &inputs);
+            return operations
+                .fastest(fastest_index)
+                .execute(inputs)
+                .expect("Should run when selected by autotune.");
+        }
+
+        // Checksum validation may retroactively turn an Unchecked entry into a Hit.
+        #[cfg(std_io)]
+        if matches!(tuner.fastest(&key), TuneCacheResult::Unchecked) {
```
nathanielsimard: Should we extract `tuner.fastest(&key)` into a variable instead of calling it 3 times in a row? Would a match statement be better here?

nathanielsimard (commented):
Overall it looks good.

To fix the fusion problem, I actually created an issue where I wanted to change the tuner API to give more info to fusion: https://github.com/tracel-ai/cubecl/issues/1295

But we already have that information actually with the input generator. So instead of calling `context.clone` for the input generation in burn fusion, we could call `context.clone_for_bench` and never track newly created tensor ids and such. That could be on top of the sync execution fix.

crates/cubecl-runtime/src/tune/local.rs
ArthurBrussee: Yeah bit awkward, refactored things here. Also instead of the whole dance of reading the cache -> maybe checksum -> maybe trying tuning -> re-reading it etc. the tuner now handles this all in try_tune, bit easier.

nathanielsimard (approved):

PR #1293, opened 2026-04-15
Track tuning lifetime

ArthurBrussee:
Follow up on https://github.com/tracel-ai/cubecl/pull/1292

Removes more unsafe code in tuning and isntead enforces the lifetime of the tuning context.


crates/cubecl-runtime/src/tune/base.rs
```diff
@@ -65,12 +86,12 @@ impl<K> Clone for TuneGroup<K> {
...
 impl<K> TuneGroup<K> {
     /// Create a new group based on a priority function.
-    pub fn new<F: Fn(&K) -> i8 + Send + Sync + 'static, S: Into<String>>(name: S, f: F) -> Self {
+    pub fn new(name: &str, f: impl Fn(&K) -> i8 + Send + Sync + 'static) -> Self {
         let id = GROUP_COUNTER.fetch_add(1, Ordering::Relaxed);
 
         Self {
             id,
-            name: Arc::new(name.into()),
+            name: Arc::new(name.to_string()),
```
nathanielsimard: Into is great because it doesn't need `alloc::string::ToString` to be imported when no-std, but if the imports are OK, I'm good with `to_string` as well.

crates/cubecl-runtime/src/tune/key_generator.rs
```diff
@@ -1,83 +1,24 @@
...
 #[diagnostic::on_unimplemented(
     message = "`{Self}` is not a valid key generator",
     label = "invalid key generator"
 )]
-pub trait FunctionKeygen<K, Inputs, Marker>: 'static {
-    /// Execute this function and generate a key
-    fn execute(&self, inputs: &Inputs) -> K;
+pub trait KeyGenerator<K, I: TuneInputs>: Send + Sync + 'static {
+    /// Generate a key from a set of inputs.
+    fn generate<'a>(&self, inputs: &I::At<'a>) -> K;
```
nathanielsimard: Is the `'a` lifetime necessary here? Can you just use `I::At<'_>` instead? I'm unsure if we need `'a` when the lifetime isn't used for the output.

crates/cubecl-runtime/src/tune/local.rs
```diff
@@ -143,40 +145,37 @@ where
...
 
-        // Resolve the cache state into a `done_rx` we can wait on. Hit → run immediately;
-        // Pending → attach to the in-flight tune; Miss → kick one off.
         match fastest {
             TuneCacheResult::Hit { fastest_index } => {
                 #[cfg(feature = "autotune-checks")]
-                self.checks(&operations, &inputs);
-                return operations
+                self.checks::<I, Out>(&operations, &inputs);
+                operations
```
nathanielsimard, line 169: It's hard to see with the diff, but do we return here automstically that's why the return was deleted?

crates/cubecl-runtime/src/tune/operation.rs
```diff
@@ -1,81 +1,89 @@
...
-        checksum += op.name();
-    });
-    alloc::format!("{:x}", md5::compute(checksum))
+/// A named, type-erased tunable function stored in a [`TunableSet`]. Constructed via
+/// [`Tunable::new`](super::Tunable::new); callers don't name this type directly.
+pub struct TuneFn<I: TuneInputs, Out> {
+    pub(crate) name: String,
+    #[allow(clippy::type_complexity)]
+    pub(crate) func:
+        Arc<dyn for<'a> Fn(<I as TuneInputs>::At<'a>) -> Result<Out, AutotuneError> + Send + Sync>,
```
nathanielsimard: The lifetime is quite hard to follow. I think I would add documentation here on what the lifetime is tracking.

crates/cubecl-runtime/src/tune/operation.rs
```diff
@@ -1,81 +1,89 @@
...
-pub struct TunableSet<K: AutotuneKey, Inputs: Send + 'static, Output: 'static> {
-    tunables: Vec<Tunable<K, Inputs, Output>>,
-    key_gen: Arc<dyn KeyGenerator<K, Inputs> + Send + Sync>,
-    input_gen: Arc<dyn InputGenerator<K, Inputs> + Send + Sync>,
-    #[allow(clippy::type_complexity)]
-    checksum_override: Option<Arc<dyn Fn(&Self) -> String + Send + Sync>>,
+impl<I: TuneInputs, Out> Clone for TuneFn<I, Out> {
+    fn clone(&self) -> Self {
+        Self {
+            name: self.name.clone(),
```
nathanielsimard: The string could be expensive to clone. Is it necessary for a tune fn to impl `Clone` ? If so, when is it called? We could wrap String into an Arc if it's called often to avoid frequent allocations.

crates/cubecl-runtime/src/tune/tune_benchmark.rs
```diff
@@ -29,16 +28,21 @@ impl AutotuneOutput for () {
...
     /// Returns at least one duration, otherwise an error is returned.
     pub fn profile(self) -> Result<Vec<ProfileDuration>, AutotuneError> {
         let client = self.client.clone();
         let name = self.operation.name().to_string();
 
+        // `scoped` holds exclusive device access for the whole benchmark loop and
+        // accepts non-`'static` closures.
         client
-            .exclusive(move || self.profile_exclusive())
+            .scoped(move || self.profile_exclusive())
```
nathanielsimard: I feel like `exclusive` was easier to understand than `scoped`, but it there another is reason for the rename?

crates/cubecl-runtime/src/tune/key_generator.rs
ArthurBrussee: removed!

crates/cubecl-runtime/src/tune/local.rs
ArthurBrussee: Yeah this whole match now just returns the execution result. Clippy doesnt like binding it to a variable, kind of annoying, as I agree it's a bit hard to see

crates/cubecl-runtime/src/tune/operation.rs
ArthurBrussee: Extracted the type with a brief explanation

crates/cubecl-runtime/src/tune/operation.rs
ArthurBrussee: I think it used to be necesarry but actually we can now just pass &TuneFn in some places and not implement clone at all! The function is now held in a Box<>

Did require making TuneBenchmark just some freestanding function, otherwise storing &TuneFn in it became too messy with lifetimes again.

crates/cubecl-runtime/src/tune/tune_benchmark.rs
ArthurBrussee: scoped already existed, just used that, its the same thing just without the 'static. After this refactor noone uses exclusive anymore, so, Ive gone ahead and removed it as the old scoped should be valid for any 'static lifetime as well anyway, and renamed scoped to exclusive. Lmk if thats premature though!

crates/cubecl-common/src/device/handle/channel.rs
```diff
@@ -147,20 +130,21 @@ impl<S: DeviceService + 'static> DeviceHandleSpec<S> for ChannelDeviceHandle<S>
...
-
         let current = StreamId::current();
 
         let wrapper = move || {
             let returned = current.executes(task);
             let _ = sender.send(returned);
         };
 
-        let boxed: Box<dyn FnOnce() -> TaskResult + Send> = Box::new(wrapper);
+        let boxed: Box<dyn FnOnce() + Send> = Box::new(wrapper);
```
nathanielsimard: Currently, the use of exclusive scope should be minimized. This implementation does an allocation to provide a stronger guarantee, but most of the time we should use exclusive. So I would revert the change about exclusive and keep the previous implementation. Exclusive scope is currently used in the client in places where refactoring is needed, so it's OK to have both `exclusive_scope` and `exclusive` for now. We shouldn't try to merge them.

ArthurBrussee: I've implemented a method that doesn't rely on boxing, our stack frame is alive long enough. This merged the _scoped methods for `exclusive` and `submit`, they're now identical in performance and semantics. 

It's a bit of annoying unsafe, but, `transmute` is also heavy unsafe machinery that should be avoided.

nathanielsimard (approved):

PR #1301, opened 2026-04-20
Add expm1 unary op across frontend / IR / backends

shinaoka:
## Summary

Adds `Arithmetic::Expm1` as a real-only unary float operation, wired through the frontend trait, the IR, constant propagation, and every backend that already lowers similar unary float ops (`cubecl-cpp` [CUDA / HIP], `cubecl-spirv`, `cubecl-wgpu`, and the CPU arithmetic visitor).

## Motivation

`expm1(x) = exp(x) - 1` is numerically more accurate than `exp(x) - 1` for small `|x|`, since `exp(x) ≈ 1 + tiny` there and the subtraction loses precision.

This PR also serves as a prerequisite for upcoming complex-number contract work. In that follow-up, `expm1` is explicitly rejected for complex types by a centralized validator (there is no standard `expm1` for complex in either `std::complex` or `num_complex::Complex<T>`). The reject rule can only type-check if `Arithmetic::Expm1` exists on the IR, so the op itself has to land first.

## Lowerings

| Backend | Implementation |
|---|---|
| CUDA / HIP (`cubecl-cpp`) | `function!(Expm1, "expm1", false)` — direct call to libm's native `expm1` |
| CPU (`cubecl-cpu`) | Taylor `x + x²/2 + x³/6` for `\|x\| < 1e-5`, else `exp(x) - 1`, selected per element |
| WGSL (`cubecl-wgpu`) | Same algorithm as CPU, via `select(exp(x) - 1, taylor, abs(x) < 1e-5)` |
| SPIR-V (`cubecl-spirv`) | Same algorithm, via `GLSL.std.450 Exp` + `OpSelect`; preserves `RelaxedPrecision` decoration and uniformity |

SPIR-V has no `GLSL.std.450 ExpM1` extended instruction and WGSL has no `expm1` builtin, so those backends emulate. The Taylor expansion is only used near zero to recover the accuracy that `exp(x) - 1` loses there; outside the threshold the native `exp` path is used directly.

## Changes

- `cubecl-core`: `Expm1` frontend trait, `Float` supertrait bound, `Vector<P, N>` and `DynamicScalar<Marker>` impls
- `cubecl-ir`: `Arithmetic::Expm1` variant
- `cubecl-opt`: instruction table entry, `constant_prop` pass entry
- Backends: CUDA / HIP / CPU / WGPU / SPIR-V lowerings as summarized above
- `crates/cubecl-core/src/runtime_tests/unary.rs`: runtime coverage for `expm1`

## Scope

This PR intentionally adds only the op and its lowerings. It does not touch complex-number handling; the complex reject rule for `expm1` lands in the follow-up complex contract PR.

## Validation

Locally verified with:

- `cargo check --tests -p cubecl-core -p cubecl-cuda -p cubecl-cpu -p cubecl-wgpu -p cubecl-hip`
- `cargo test -p cubecl-cuda expm1 -- --nocapture`


louisfd (approved):
LGTM

shinaoka:
Rebased onto the origin/main

PR #1304, opened 2026-04-20
Send recv

Charles23R:
- Introduce `ServerCommunication::comm_init` to decouple communications initialization from collective operations.
- `CudaServer` now uses NCCL to send data between 2 devices. Which is what is used when doing `to_device` in Burn.

... COMING SOON ...

### Instructions


crates/cubecl-common/src/device/handle/mod.rs
```diff
@@ -54,6 +54,10 @@ impl<S: DeviceService> DeviceHandle<S> {
         }
     }
 
+    pub fn id(&self) -> DeviceId {
+        self.handle.device_id()
+    }
```
nathanielsimard: I would call it device id instead of just id.

crates/cubecl-cuda/src/compute/server.rs
```diff
@@ -502,156 +455,106 @@ impl CudaServer {
...
+                count,
+                nccl_dtype,
+                rank_dst,
+                *comm_src,
+                server_src.comm_stream as _,
+            )
+            .map_err(|e| ServerError::Generic {
+                reason: format!("NCCL send failed: {e:?}"),
+                backtrace: BackTrace::capture(),
+            })?;
```
nathanielsimard, line 468: Eventually it would be good to split those two nccl call into two server functions. The reason is that the client shouldn't have to "lock" both server at the same time:

```rust
client_src.submit_blocking(|server_src| {
    client_dst.submit_blocking(|server_dst| {
        // Bad pattern
    })
})
```

It would be better to have something like that:

```rust
let id_or_something = client_src.submit_blocking(|server| server.send(..));
client_dst.submit(|server| server.recv(id_or_something, output));
return output;
```

I don't think that's necessary for this PR, but that's a future improvement.

Charles23R: I ended up doing it for this PR

nathanielsimard (approved):

PR #1321, opened 2026-05-03
Add f64 support back for CUDA.

vaijira:

### Instructions


nathanielsimard (commented):
@wingertge we should check if the problem is corrected before merging

wingertge:
I'm pretty sure it was just alignment issues, this was before the many reworks I've done to how metadata/scalars are passed since then. I'd re-enable it and we can always go back to disabling if it causes issues somewhere downstream.
`CUDA_ERROR_INVALID_VALUE` is always because one of the parameters passed to launch violates some invariant, it's not a kernel problem.

wingertge (approved):

PR #1327, opened 2026-05-09
Add workgroupUniformLoad primitive

ArthurBrussee:
Adds a `workgroupUniformLoad`-style primitive to CubeCL. Required for WebGPU compliance when a workgroup-shared value gates control flow that contains `workgroupBarrier` calls — per WGSL spec, plain workgroup loads return a non-uniform value, so the surrounding control flow fails Tint's uniformity analysis. WebGPU's `workgroupUniformLoad` exists specifically to mark the result as uniform. Uniformity is implicit on non-WebGPU backends, so those just need barrier + load.

This PR adds two methods on shared memory, one for atomics, one for non-atomics, matching the WebGPU overloads.

Discovered while porting brush's WGSL gaussian-splat kernels to CubeCL, the rasterize forward kernel uses workgroup-shared range + early-exit atomic to gate its inner loop, and Tint rejected the generated WGSL without this primitive.

### DIFF
--- a/crates/cubecl-core/src/frontend/synchronization.rs
+++ b/crates/cubecl-core/src/frontend/synchronization.rs
@@ -1,7 +1,10 @@
 use crate::{
-    ir::{Scope, Synchronization},
+    frontend::{NativeExpand, element::Atomic},
+    ir::{Instruction, Scope, Synchronization},
+    prelude::{CubePrimitive, Numeric},
     unexpanded,
 };
+use cubecl_ir::Operation;
 
 // Among all backends, the memory order guarantee of WebGPU is the weakest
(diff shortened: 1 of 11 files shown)


wingertge:
I think this should be refactored now that we properly support references to take a `&T` instead. This matches the syntax in WGSL too.

ArthurBrussee:
Yep good timing, working on it :) 

The tricky bit is doing an impl for scalars & atomics & vectors that returns the right thing. ` <T as CubePrimitive>::Scalar` works for both atomics and not atomics, but would be wrong for vectors.

wingertge (approved):

PR #1332, opened 2026-05-15
fix: Ensure constant block args are generated in the correct block

wingertge:
No idea how this wasn't hit before, MLIR was just generating constant block args in whatever block happened to be generated last, not the source block. This fixes the issue by ensuring `block` is set to the source block for the duration of `get_block_args`.

### DIFF
--- a/crates/cubecl-cpu/build.rs
+++ b/crates/cubecl-cpu/build.rs
@@ -1,5 +1,12 @@
+use std::env;
+
 fn main() -> Result<(), Box<dyn std::error::Error>> {
     // required on macos
     tracel_llvm_bundler::config::set_homebrew_library_path()?;
+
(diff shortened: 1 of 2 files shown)


marcantoinem (commented):
Good job, that would be quite a problem and would explain strange failure. For constant value every constant is sent to the start of the MLIR module with one of the passes of  MLIR. That's why it wasn't hit before. There are also LocalMut  and GlobalScalar that would have this problem as the load instruction would be in a last generated block instead of the previous block.

wingertge:
> Good job, that would be quite a problem and would explain strange failure. For constant value every constant is sent to the start of the MLIR module with one of the passes of MLIR. That's why it wasn't hit before.

Might be a bug in the new version of MLIR then, because it crashed instantly with a null pointer dereference during the initial canonicalization pass. Either way, this fixes it.

nathanielsimard (approved):

PR #1333, opened 2026-05-16
feat(runtime): add memory_usage_total across all streams

ArthurBrussee:
`memory_usage` only reports the calling thread's stream (keyed on `StreamId::current()`), so querying it from another thread sees a near-empty pool. `memory_usage_total` sums `MemoryUsage` across every stream.

### DIFF
--- a/crates/cubecl-runtime/src/server/base.rs
+++ b/crates/cubecl-runtime/src/server/base.rs
@@ -409,6 +409,12 @@ where
     /// The current memory usage of the server.
     fn memory_usage(&mut self, stream_id: StreamId) -> Result<MemoryUsage, ServerError>;
 
+    /// Total memory usage across all streams. Default reports only the
+    /// calling stream; multi-stream backends override to aggregate.
+    fn memory_usage_total(&mut self) -> Result<MemoryUsage, ServerError> {
+        self.memory_usage(StreamId::current())
+    }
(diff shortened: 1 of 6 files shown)


crates/cubecl-runtime/src/server/base.rs
```diff
@@ -409,6 +409,12 @@ where
     /// The current memory usage of the server.
     fn memory_usage(&mut self, stream_id: StreamId) -> Result<MemoryUsage, ServerError>;
 
+    /// Total memory usage across all streams. Default reports only the
+    /// calling stream; multi-stream backends override to aggregate.
+    fn memory_usage_total(&mut self) -> Result<MemoryUsage, ServerError> {
+        self.memory_usage(StreamId::current())
```
nathanielsimard: Unsure I would provide a default implementation that is probably wrong.

nathanielsimard (changes requested):
I think we should also implement the function on the cuda and rocm runtime.

ArthurBrussee:
I've changed it to a different approach which exposes an iterator of streams, and uses it to make `memory_usage` and `memory_cleanup` in client aggreate by default for all backends.

nathanielsimard (commented):
LGTM. We can probably merge after the CI is fixed.

PR #1341, opened 2026-05-21
Feat/obfuscation

nathanielsimard:
Generate a type-erased, inline wrapper for a single value.

Expands into a private module containing an `Opaque` struct that stores one value of `type:` behind a fixed-shape representation that never names the inner type. Downstream crates that handle the wrapper never have to resolve the inner type, which breaks the transitive monomorphization chain and can visibly cut compile times for deeply generic types.

### DIFF
--- a/crates/cubecl-common/src/obfuscation.rs
+++ b/crates/cubecl-common/src/obfuscation.rs
@@ -22,6 +22,7 @@
 /// obfuscate!(
 ///     type: Tensor,
 ///     module: tensor,
+///     align: 8,
 ///     derives: [Send, Sync, Debug, PartialEq, Clone],
 /// );
 ///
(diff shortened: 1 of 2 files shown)


crates/cubecl-common/src/obfuscation.rs
```diff
@@ -0,0 +1,841 @@
...
+                    /// owner is dropped).
+                    pub(in super::super) fn into_inner(self) -> $inner {
+                        // SAFETY: read the bytes through the
+                        // correctly-typed pointer, then forget `self`
+                        // to skip our `Drop` (which would otherwise
+                        // drop the value again).
+                        let inner: $inner =
+                            unsafe { ::core::ptr::read(self.data.as_ptr() as *const $inner) };
+                        ::core::mem::forget(self);
+                        inner
```
laggui: Small nit: missed this in the burn PR, but I think `ManuallyDrop::new(self)` should be generally preferred.

laggui (approved):

PR #1343, opened 2026-05-22
fix metal compilation

louisfd:


crates/cubecl-cpp/src/metal/binding.rs
```diff
@@ -7,7 +7,16 @@ use crate::{
 };
 
 pub fn bindings(repr: &MslComputeKernel, args: &KernelArguments) -> (Vec<Visibility>, usize) {
-    let buffers = repr.buffers.iter().map(|it| it.vis);
+    // When slices are shared, it needs to be read-write if ANY of the slices is read-write,
+    // and since we can't be sure, we'll assume everything is read-write.
+    #[allow(unexpected_cfgs)]
```
wingertge: Does the CFG actually transfer up to `cubecl-cpp`? I think the real fix is to just move the `bindings` function into `cubecl-wgpu` like Vulkan.

crates/cubecl-cpp/src/metal/dialect.rs
```diff
@@ -247,7 +247,7 @@ struct alignas({alignment}) {item} {{"
             }
         }
 
-        define_array_polyfill(f, "")?;
+        define_array_polyfill(f, "thread")?;
```
wingertge, line 250: I looked it up and there was zero info on this, the only info I could find claimed all functions are device functions by default... Thanks for fixing this

crates/cubecl-cpp/src/metal/binding.rs
louisfd: it was just a patch, i will move it

louisfd:
@wingertge is this what you meant?

wingertge (approved):

crates/cubecl-cpp/src/metal/binding.rs
wingertge: I've wanted to move that for a while but I keep forgetting, this is a good chance to finally get that done 😅

PR #1344, opened 2026-05-22
refactor: Add a lifetime to views

wingertge:
Adds a lifetime to `View`, so it borrows the source buffer for the duration of the view when it's created from a reference.
I also looked into adding one to `VirtualTensor` but that's used for much more hacky and niche cases, so it didn't seem worth it.

### Instructions

### DIFF
--- a/crates/cubecl-core/src/runtime_tests/barrier.rs
+++ b/crates/cubecl-core/src/runtime_tests/barrier.rs
@@ -8,7 +8,7 @@ use num_traits::Zero;
 #[cube(launch)]
 pub fn async_copy_test<F: Float, N: Size>(input: &[Vector<F, N>], output: &mut [Vector<F, N>]) {
     let barrier = Barrier::local();
-    let mut smem = SharedMemory::<Vector<F, N>>::new(1usize);
+    let mut smem = Shared::<Vector<F, N>>::new_array(1usize);
 
     let source = &input[2..3];
     let destination = &mut smem[..1];
@@ -50,7 +50,7 @@ pub fn test_async_copy<R: Runtime, F: Float + CubeElement>(client: ComputeClient
 
 #[cube(launch)]
 fn one_load<F: Float, N: Size>(lhs: &Tensor<Vector<F, N>>, output: &mut Tensor<Vector<F, N>>) {
-    let mut lhs_smem = SharedMemory::<Vector<F, N>>::new(4usize);
+    let mut lhs_smem = Shared::<Vector<F, N>>::new_array(4usize);
 
     let barrier = Barrier::shared(CUBE_DIM, UNIT_POS == 0);
     sync_cube();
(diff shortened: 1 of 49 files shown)


crates/cubecl-core/src/runtime_tests/barrier.rs
```diff
@@ -102,8 +102,8 @@ fn two_independent_loads<F: Float, N: Size>(
     output: &mut Tensor<Vector<F, N>>,
     #[comptime] num_data: usize,
 ) {
-    let mut lhs_smem = SharedMemory::<Vector<F, N>>::new(num_data);
-    let mut rhs_smem = SharedMemory::<Vector<F, N>>::new(num_data);
+    let mut lhs_smem = Shared::<Vector<F, N>>::new_array(num_data);
```
nathanielsimard: new_array is a bit strange as a name for shared memory. Is there a reason why it's not new?

crates/cubecl-ir/src/arena.rs
```diff
@@ -0,0 +1,76 @@
...
+/// let mut pool = DropBump::new();
+/// let w = pool.alloc(Wrapper(&mut external));
+/// drop(external);
+/// pool.reset(); // Runs `drop_in_place::<Wrapper>`
+/// ```
+///
+/// So don't do that.
+///
+#[derive(Default, Debug, TypeHash)]
+pub struct DropBump {
```
nathanielsimard, line 37: What kind of values are dropped? Is it inside a kernel?

nathanielsimard (commented):
I think this is a good cleanup, but looking at the cubek update, the amount of new lifetime added is a bit overwhelming. I mean we don't need the lifetime for real, since it's JIT compiled code, but it's true that it better fits the mental model of Rust.

Is there a way we could have a new view type `OwnedView` where the lifetime would go away? Inside the kernel, the buffer is still valid during the whole computation, so we don't need a lifetime to prevent a use after drop bug.

crates/cubecl-core/src/runtime_tests/barrier.rs
wingertge: Ideally it would be a `Shared<Array<T, N>>` but I still need to figure out if a generic size is worth it on array. `SharedMemory` was just an alias now, you're essentially creating a `Box<[T]>` except with `Shared` as the smart pointer type.

crates/cubecl-ir/src/arena.rs
wingertge: expands like `Sequence`, etc. Layouts are stored in here so this is very necessary, but it was already prone to leaking before. I kind of forgot about it in my initial implementation

wingertge:
> I think this is a good cleanup, but looking at the cubek update, the amount of new lifetime added is a bit overwhelming. I mean we don't need the lifetime for real, since it's JIT compiled code, but it's true that it better fits the mental model of Rust.

It's not about use after drop, it's about properly borrowing the buffer while the view exists. Same reason you can't do
```rust
let x = vec![1, 2, 3, 4];
let y = &x[..];
let z = &mut x[..];
```
 
> Is there a way we could have a new view type `OwnedView` where the lifetime would go away? Inside the kernel, the buffer is still valid during the whole computation, so we don't need a lifetime to prevent a use after drop bug.

It's just `View<'static, T>`. I only added lifetimes where they were trivial to add, it just makes it a bit more checked. The ones with `VirtualTensor` just use a static view.

crates/cubecl-core/src/runtime_tests/barrier.rs
wingertge: I just checked the docs for `Box` and it's called `new_uninit_slice` there. I think we can drop the `uninit` because shared memory is always uninit, but I think `new_slice` is better.

nathanielsimard (approved):

PR #1345, opened 2026-05-22
Improve CPU backend by removing need to flush, add atomics test and add atomic to cubecl-cpu

marcantoinem:
It's quite a big PR, a following PR for optimizing the threadpool with a work steal scheduler will come later

### Instructions

### DIFF
--- a/crates/cubecl-cpu/src/compiler/visitor/operation/synchronization.rs
+++ b/crates/cubecl-cpu/src/compiler/visitor/operation/synchronization.rs
@@ -1,7 +1,350 @@
-use cubecl_core::ir::Synchronization;
-use tracel_llvm::mlir_rs::{dialect::func, ir::attribute::FlatSymbolRefAttribute};
+use std::ops::Deref;
 
-use crate::compiler::visitor::prelude::*;
+use cubecl_core::ir::Synchronization;
+use tracel_llvm::mlir_rs::{
+    dialect::{
+        arith::{self, AtomicRMWKind, CmpiPredicate},
+        cf, func,
+        llvm::{self, LoadStoreOptions, attributes::AtomicOrdering},
+        memref,
+    },
+    ir::{
+        Block, BlockRef, Identifier, Location, Region,
+        attribute::{FlatSymbolRefAttribute, IntegerAttribute, StringAttribute, TypeAttribute},
+        r#type::{FunctionType, IntegerType, MemRefType},
+    },
+};
+
+use crate::compiler::{
+    mlir_data::{
+        SYNC_BARRIER_COUNTER_INDEX, SYNC_BARRIER_TARGET_INDEX, SYNC_STOPPED_COUNTER_INDEX,
+    },
+    visitor::prelude::*,
+};
+
+pub fn add_sync_cube_function<'a>(
+    context: &'a Context,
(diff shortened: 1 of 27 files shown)


crates/cubecl-ir/src/features.rs
```diff
@@ -108,10 +108,16 @@ impl TypeUsage {
 pub enum AtomicUsage {
     /// Atomic loads and stores
     LoadStore,
+    /// Atomic exchange
+    Swap,
```
wingertge: Is there a reason `Swap` was split out from `LoadStore`? It's always supported if `load` and `store` are supported on all platforms. It's just a load and store in one operation.

marcantoinem: I was not sure if cmpxchg work everywhere in MLIR, I will readd it back in LoadStore

marcantoinem: Nevermind I was thinking it was compare exchange yeah, swap should be supported as long as loadstore is supported

wingertge: yeah I initially thought the same about compare exchange, but actually it's true it shouldn't be supported on floats because it doesn't follow the proper semantics of float comparison, it just does bitwise comparison. So it should only be supported on integers.

crates/cubecl-cpu/src/compiler/visitor/operation/synchronization.rs
```diff
@@ -1,7 +1,350 @@
...
+};
+
+use crate::compiler::{
+    mlir_data::{
+        SYNC_BARRIER_COUNTER_INDEX, SYNC_BARRIER_TARGET_INDEX, SYNC_STOPPED_COUNTER_INDEX,
+    },
+    visitor::prelude::*,
+};
+
+pub fn add_sync_cube_function<'a>(
```
nathanielsimard, line 34: Ok so now the sync_cube method uses atomics from the cpu compiler instead of a function handled by Rust right?

crates/cubecl-cpu/src/compiler/visitor/operation/synchronization.rs
```diff
@@ -1,7 +1,350 @@
...
+
+            done.append_operation(func::r#return(&[], location));
+            region
+        },
+        &[(
+            Identifier::new(context, "sym_visibility"),
+            StringAttribute::new(context, "private").into(),
+        )],
+        location,
+    ));
```
nathanielsimard, line 97: That function is pretty big. I think it's worth documenting it or/and spliting it into a few smaller functions.

crates/cubecl-cpu/src/compiler/visitor/operation/synchronization.rs
marcantoinem: Exactly, it allow to pass more easily code, it would still be possible to pass a Rust function instead,  if we want to make it easier to change sync_cube

crates/cubecl-cpu/src/compiler/visitor/operation/synchronization.rs
marcantoinem: Fixed

nathanielsimard (approved):

PR #1346, opened 2026-05-22
Refactor/wgpu compilers

nathanielsimard:
Avoid the conflicts of backend with wgpu with different compiler. AutoCompiler keeps the same behavior as before.


louisfd (approved):
LGTM

PR #1347, opened 2026-05-23
fix: Fix more compilation issues on metal

wingertge:
Also gates `renderdoc` so rust-analyzer works with `--all-features` on MacOS


nathanielsimard (approved):

louisfd (commented):
LGTM

PR #1348, opened 2026-05-24
fix: short-circuit || and && in #[cube] code

LucaCappelletti94:
`||` and `&&` inside `#[cube]` evaluated both operands unconditionally, breaking the canonical `idx == 0 || arr[idx - 1] != x` guard. On `idx == 0` the right operand still ran, `0usize - 1` underflowed to `usize::MAX`, and `arr[usize::MAX]` was loaded.

This patch lowers `a || b` as `if a { true } else { b }` and `a && b` as `if a { b } else {false }` via the existing `if_else_expr_expand` plumbing. New regression test in `crates/cubecl-core/src/runtime_tests/short_circuit.rs`.

Standalone reproducer: https://github.com/LucaCappelletti94/cubecl-or-and-shortcircuit (`cargo run --release` shows the bug, `--features fixed` shows it gone).

### DIFF
--- a/crates/cubecl-core/src/runtime_tests/mod.rs
+++ b/crates/cubecl-core/src/runtime_tests/mod.rs
@@ -26,6 +26,7 @@ pub mod plane;
 pub mod properties;
 pub mod saturating;
 pub mod sequence;
+pub mod short_circuit;
 pub mod slice;
 pub mod stream;
 pub mod synchronization;
@@ -159,6 +160,8 @@ macro_rules! testgen_untyped {
 
         cubecl_core::testgen_to_client!();
         cubecl_core::testgen_all_reduce!();
+
+        cubecl_core::testgen_short_circuit!();
     };
 }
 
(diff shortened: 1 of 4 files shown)


wingertge:
This *will* cause performance regressions, I would prefer if the new codegen is special cased so it only triggers if one of the ops in either operand is not guaranteed to be pure (function call, index, etc). Would just need a recursive `is_always_pure` method on `Expression`. Things like binops are always pure if the operands are pure.

LucaCappelletti94:
While adding the `is_always_pure`, I found two more bugs. I will open PRs for those as well. A colleague said that I really have [the brown touch](https://www.urbandictionary.com/define.php?term=The+brown+touch)

LucaCappelletti94:
@wingertge added the `is_always_pure`, I will make 2 other PRs for the other 2 bugs. One is quite apparent and it will be immediate, the other one I only have found for now.

wingertge (approved):

PR #1350, opened 2026-05-25
fix: CubeOr expand methods computed AND instead of OR

LucaCappelletti94:
`CubeOr` was a copy of `CubeAnd` with the names switched to "or", but the method bodies and trait bound still referenced "and", so it emitted a logical AND.

This error was entirely latent today: `||` codegen always operates on `NativeExpand<bool>`, which uses the correct `OrExpand` impl, so the buggy `bool` path is never reached. Codegen can't exercise that path to my understanding at this time, but I thought it was a timebomb to be left as it was.

### DIFF
--- a/crates/cubecl-core/src/frontend/operation/binary.rs
+++ b/crates/cubecl-core/src/frontend/operation/binary.rs
@@ -330,18 +330,18 @@ impl<T: CubeAnd + CubePrimitive> AndExpand for NativeExpand<T> {
     }
 }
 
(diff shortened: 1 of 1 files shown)


wingertge:
That code path exists because the macro *doesn't* always correctly convert to a `NativeExpand` (i.e. a struct with a comptime `bool` field). So that was a ticking time bomb indeed 😬

wingertge (approved):

PR #1351, opened 2026-05-25
fix: value-form if/else with literal arms ignored the runtime condition

LucaCappelletti94:
A value-form if/else with literal-constant arms (`x = if cond { 1 } else { 0 }`) ignored the runtime condition and always produced the else value, emitting empty branches. It hit every numeric type and every backend (confirmed on CUDA, CPU, and wgpu). Using a constructor (`F::new` / `from_int`) for the arms worked, which is what kept it hidden. It is pre-existing on main, not a regression from a recent change.

Root cause: a literal arm expands to a comptime primitive, so `if_else_expr_expand` is instantiated with `C = <primitive>` and resolves to the blanket `Assign for CubePrimitive` impl, which does plain Rust assignment and emits no IR. The select then happens at comptime and the runtime condition is dropped. The fix forces both arms to runtime via `into_expand`. It lives in the value-form `If` arm that only fires for runtime conditions (comptime ones match the earlier `condition.is_const()` arm), so it cannot turn a genuinely comptime if-expression, such as one feeding an array size, into a runtime one.

The regression test is split by backend type support: a portable f32/i32/u32 case runs on every backend through `testgen_branch`, and an all-types case (i8..u64, f32, f64) runs only on CUDA and CPU via `testgen_value_form_literals!`, because wgpu rejects i8/i16/u8/u16/f64 as element types.

Generated CUDA for `output[0] = if cond != 0 { 30i32 } else { 50i32 }`:

On main:

```c
const bool l_3 = info.scalars_uint32[0] != uint32(0);
if (l_3) {
} else {
}
...
*l_12 = uint32(50);   // both branches empty, always the else value
```

With this PR:

```c
int32 l_mut_4;
const bool l_3 = info.scalars_uint32[0] != uint32(0);
if (l_3) {
    l_mut_4 = int32(30);
} else {
    l_mut_4 = int32(50);
}
...
*l_13 = l_mut_4;
```

Verified on CUDA (full lib suite, no new failures), CPU, and wgpu.

### DIFF
--- a/crates/cubecl-core/src/runtime_tests/branch.rs
+++ b/crates/cubecl-core/src/runtime_tests/branch.rs
@@ -3,6 +3,72 @@ use alloc::{vec, vec::Vec};
 use crate::{self as cubecl, as_bytes};
 use cubecl::prelude::*;
 
+// Regression: a value-form if/else with literal-constant arms must honor the runtime
+// condition (it used to select at comptime and always return the else value). One kernel
+// per type, plus a checker asserting the two conditions produce different output.
+macro_rules! lit_value_form {
+    ($kernel:ident, $check:ident, $ty:ty, $a:expr, $b:expr) => {
+        #[cube(launch)]
+        pub fn $kernel(output: &mut [$ty], cond: u32) {
+            if UNIT_POS == 0 {
+                let flag = cond != 0u32;
+                output[0] = if flag { $a } else { $b };
+            }
+        }
+
+        fn $check<R: Runtime>(client: &ComputeClient<R>) {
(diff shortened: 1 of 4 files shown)


wingertge:
There's an actual issue with the implementation of `init_mut` here, it should be returning a mutable runtime variable but I need to think a bit more about the right signature so it still works with custom structs. I'll see if I can do a more fundamental fix before I merge this band-aid solution.

LucaCappelletti94:
I have also added the test for the `match` variant of the same bug, so that we do not forget about it. It is currently a red test. Hopefully it will help you understand what could be a more general less patchy solution.

PR #1354, opened 2026-05-26
fix: Fix atomics on CUDA again

wingertge:
This is still very hacky, we might want to set a minimum CUDA version at some point and use a more modern API.

### DIFF
--- a/crates/cubecl-cpp/src/shared/dialect.rs
+++ b/crates/cubecl-cpp/src/shared/dialect.rs
@@ -1,11 +1,12 @@
 use std::{collections::HashSet, fmt::Debug};
 use std::{fmt::Display, hash::Hash};
 
-use cubecl_core::ir::{ConstantValue, Processor};
+use cubecl_core::ir::Processor;
 
 use crate::shared::{
     FmtLeft, IndexedVariable, MmaShape, SupportedMmaCombinations, SupportedScaledMmaCombinations,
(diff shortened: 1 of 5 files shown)


crates/cubecl-cpp/src/shared/dialect.rs
```diff
@@ -485,8 +486,42 @@ pub trait DialectInstructions<D: Dialect> {
...
+            4 => Item::Scalar(Elem::<D>::U32),
+            8 => Item::Scalar(Elem::<D>::U64),
+            // Hacky, but it's CUDA only for now. We should really migrate to a more modern API in
+            // general
+            16 => {
+                let out_tmp = Variable::tmp(out_item);
+                writeln!(f, "{};", out_tmp.fmt_left())?;
+                writeln!(
+                    f,
+                    "__nv_atomic_load({input}, &{out_tmp}, __NV_ATOMIC_RELAXED);"
```
nathanielsimard, line 508: This code is in the shared dialect. Would it be possible to extract it into the cuda specific dialect?

wingertge: Not without a lot of duplication since everything other than 16 byte loads is shared. Only CUDA supports loads larger than 64 bit so I just used the CUDA intrinsic for now, and I want to eventually replace all this with std::atomic if I can make it with with ROCm. But this is a quick fix for now.

nathanielsimard (approved):

PR #1361, opened 2026-05-28
fix(wgpu): bound live timestamp query sets per device (Metal counter-sample-buffer exhaustion)

AdrianEddy:
## Problem

On Metal, concurrent cold autotune across many streams crashes:

```
Failed to create counter sample buffer: "Cannot allocate sample buffer" (MTLCounterErrorDomain)
-> Buffer with 'CubeCL gpu -> cpu map buffer' label is invalid
-> Should have handle for tensor / Ordering is bigger than operations
```

Each GPU-timestamp `QuerySet` is backed by an `MTLCounterSampleBuffer`, and Metal caps the
number of **live** counter sample buffers at **32 per device** (measured: 32 OK, 33 fails —
an undocumented driver limit, not exposed via `wgpu::Limits` or any `MTLDevice` property).
`QueryProfiler` allocated query sets per stream with no device-wide accounting, so under
concurrent cold autotune the total blew past 32. Metal then failed the allocation, the
invalid query set poisoned the profiling map buffer, and wgpu's default uncaptured-error
handler escalated it to a fatal panic that took down the process.

## Fix

Introduce a per-device, lock-free **`TimestampQuerySetBudget`** that bounds the total live
timestamp query sets on a device, shared (`Arc`) by every stream/`QueryProfiler` on it:

- **`metal()`** caps at `DEFAULT_MAX_METAL_TIMING_QUERY_SETS = 28` (32 minus headroom),
  overridable via `CUBECL_MAX_METAL_TIMING_QUERY_SETS`.
- **`unbounded()`** for every other backend → unchanged behavior.
- Acquire/release is a single atomic (`compare_exchange` / `fetch_sub`) — **no locks, no
  hot-path synchronization**.

Each `WgpuStream` reserves a slot at creation to decide device-vs-system timing; if the
device's budget is full it falls back to the (fully supported) system timer instead of
exhausting the hardware. A `QueryProfiler` allocates new query sets while it can acquire
budget, reuses its own round-robin once the budget is exhausted (safe — a device's streams
share one ordered queue, so each set's timestamp-write + resolve stay atomic and in order),
and releases its slots on `Drop`.

## Behavior / compatibility

- **Non-Metal backends are unaffected** (unbounded budget = prior behavior).
- No new public API on the hot path; the budget type is `pub` but its methods are `pub(crate)`.
- Under genuine budget pressure on Metal, the only degradation is that a profile spanning
  more compute passes than its profiler has query sets may under-measure — which autotune
  tolerates — instead of crashing.

## Testing

Reproduced and verified on macOS / Metal with a 36-thread concurrent-autotune harness
(distinct input shapes per thread → distinct autotune keys → genuine concurrent profiling
across >32 streams):

- **Default budget:** 36/36 threads complete, run finishes, zero crash signatures, exit 0.
- **Budget disabled** (`CUBECL_MAX_METAL_TIMING_QUERY_SETS=9999`): reproduces the original
  `Cannot allocate sample buffer` cascade, confirming the budget is load-bearing.

`cargo check -p cubecl-wgpu --features msl` passes clean.

## Out of scope / follow-ups

- The 32 limit isn't queryable (not in `wgpu::Limits`, not on `MTLDevice`); a hardcoded
  constant + env override matches existing wgpu conventions. A startup probe could discover
  it dynamically if preferred.
- wgpu's default error handler still panics on *any* uncaptured profiling error; making the
  device-time path fully fallible (so all profiling hiccups degrade to `ProfileError`) is a
  larger, separate change touching `ProfileDuration`'s API.

---

🤖 Co-authored with [Claude Code](https://claude.com/claude-code) (Claude Opus 4.7).


nathanielsimard:
If you change the number of maximum streams, does that fix the problem? Because it's something we could be introducing: a maximum number of streams. That would be simpler than the current solution.

AdrianEddy:
I tried that at first but it didn't fix the issue - each stream was using more than 1 timestamp counter and it wasn't really deterministic. I ended up with this solution which I think is the most proper fix


For context, I'm using 5 different models in my app and they are working concurrently so that's probably why this never shows in tests or testing with one model, but is a showstopper for my app

crates/cubecl-wgpu/src/compute/timings.rs
```diff
@@ -10,13 +13,84 @@ use wgpu::{QUERY_SIZE, QuerySet, QuerySetDescriptor, QueryType};
...
+/// Bounds the live timestamp [`QuerySet`]s on a single device.
+///
+/// Each timestamp query set is backed by a Metal `MTLCounterSampleBuffer`, capped at 32 live
+/// per device; exceeding it fails the allocation, poisons profiling, and panics. Shared (via
+/// [`Arc`]) by every [`QueryProfiler`] on the device so the total stays under the limit no
+/// matter how many streams profile at once. Lock-free. Non-Metal backends use an
+/// [unbounded](Self::unbounded) budget and are unaffected.
+#[derive(Debug)]
+pub struct TimestampQuerySetBudget {
+    live: AtomicUsize,
```
nathanielsimard: I think we could use AtomicU32, if I recall AtomicUsize isn't well supported on all platforms.

crates/cubecl-wgpu/src/compute/timings.rs
```diff
@@ -10,13 +13,84 @@ use wgpu::{QUERY_SIZE, QuerySet, QuerySetDescriptor, QueryType};
...
     init_tokens: Vec<ProfilingToken>,
     query_set_pool: Vec<QuerySet>,
     query_sets: HashMap<QuerySetId, QuerySetItem>,
+    /// Shared device budget of live timestamp query sets (see [`TimestampQuerySetBudget`]).
+    budget: Arc<TimestampQuerySetBudget>,
+    /// Budget slots held, released on drop. Starts at 1 — the caller reserves one before `new`.
+    held: usize,
+    /// Query sets allocated so far; reused round-robin once the budget is exhausted.
+    created_query_sets: Vec<QuerySet>,
+    reuse_idx: usize,
```
nathanielsimard: Those new fields  could be grouped into a new struct with better docs.

nathanielsimard (approved):

PR #1363, opened 2026-05-29
fix(cuda): route F16 ↔ BF16 casts through special_cast

ogghead:
## Summary

CUDA has no direct `__nv_bfloat16(__half)` or `__nv_bfloat162(__half2)` constructor. When a kernel casts between F16 and BF16 (e.g. on quantized dequantize paths where an F16 intermediate feeds a BF16 output), the `Instruction::Assign` path emits an invalid constructor call that fails nvcc compilation.

This adds:
- `is_f16_bf16_cross()` in `base.rs` to detect F16↔BF16 cross-casts and route them through `SpecialCast` instead of `Assign`
- `cast_half_to_bfloat()` in `convert.rs` to emit the correct round-trip through F32 using `__half2float` / `__float2bfloat16_rn` (scalar) and `__half22float2` / `__float22bfloat162_rn` (packed)

## Motivation

Discovered while implementing NVFP4 E2M1 weight-only quantization. The `QuantizedView` dequant path reads e2m1x2 values, produces an F16 intermediate, and needs to output BF16. Without this fix, the generated CUDA code tries to construct `__nv_bfloat162` directly from `__half2`, which doesn't exist — nvcc rejects the kernel.

The same issue would affect any kernel that mixes F16 and BF16 operands through a `Cast` operator.

## Changes

- `crates/cubecl-cpp/src/shared/base.rs`: Add `is_f16_bf16_cross()` and extend the `SpecialCast` guard to include F16↔BF16 cross-casts
- `crates/cubecl-cpp/src/cuda/convert.rs`: Add `cast_half_to_bfloat()` emitting F32 round-trip, and a branch in `special_cast()` to call it

### DIFF
--- a/crates/cubecl-cpp/src/shared/base.rs
+++ b/crates/cubecl-cpp/src/shared/base.rs
@@ -1585,7 +1585,9 @@ impl<D: Dialect> CppCompiler<D> {
             }),
             // Needs special conversion semantics
             gpu::Operator::Cast(op)
-                if (is_fp4_fp6_fp8(op.input.elem_type()) || is_fp4_fp6_fp8(out.elem_type()))
+                if (is_fp4_fp6_fp8(op.input.elem_type())
+                    || is_fp4_fp6_fp8(out.elem_type())
+                    || is_f16_bf16_cross(op.input.elem_type(), out.elem_type()))
                 // Trivial broadcast shouldn't use special cast logic
                     && op.input.elem_type() != out.elem_type() =>
             {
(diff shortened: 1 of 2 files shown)


crates/cubecl-cpp/src/shared/base.rs
```diff
@@ -2080,6 +2082,19 @@ fn is_fp4_fp6_fp8(elem: gpu::ElemType) -> bool {
...
+fn is_f16_bf16_cross(a: gpu::ElemType, b: gpu::ElemType) -> bool {
+    let is_half_or_bhalf = |e: gpu::ElemType| {
+        matches!(
+            e,
+            gpu::ElemType::Float(FloatKind::F16) | gpu::ElemType::Float(FloatKind::BF16)
+        )
+    };
+    is_half_or_bhalf(a) && is_half_or_bhalf(b) && a != b
+}
+
```
wingertge, line 2097: This would match both f16 -> bf16 and bf16 -> f16, but it seems like the implementation is only for f16 -> bf16.

ogghead: Ah good catch, in my case I only needed F16 -> BF16 but this leaves a gap. Do you think it would be better for me to narrow this to only catch F16 -> BF16 or to add an implementation for BF16 -> F16 casting?

wingertge: I checked the docs and [it seems like `__half` has a constructor for `__bfloat16`](https://docs.nvidia.com/cuda/cuda-math-api/cuda_math_api/struct____half.html#_CPPv4N6__half6__halfEK13__nv_bfloat16) so I think narrowing the check is the right thing here.

ogghead: After you mentioned this, I checked the CUDA docs again and found that there totally is an F16 -> BF16 cast method built in. I think my driver version must have been the real issue... digging into it now but I _think_ this PR is unnecessary.

Apologies for taking your time 😭

PR #1374, opened 2026-06-06
Don't panic when const-folding an unfoldable const expression

LucaCappelletti94:
As with #1372 and #1373, this turns a fuzzer-found panic in the `#[cube]` macro into correct behavior, on the way to differential fuzzing once the front end is stable (see #1364).

The binary and unary `const-folding` paths in parse/expression.rs call `as_const(context).unwrap()` after checking `is_const()`. That assumes every const expression can be rendered back as a `const` token stream, but some cannot: a compiler intrinsic such as `vectorization_of` reports `is_const()` as `true` while `as_const()` returns `None`, so the macro aborted the user's build with an opaque panic. The fix falls back to a runtime expression when folding is not possible.

Reproducer: `vectorization_of(x) + vectorization_of(x)`.

image


nathanielsimard (approved):
Just need to fix merge conflicts

PR #1376, opened 2026-06-06
Fix: returning panic payload in device channel tasks to caller and re-raising panic for blocking APIs

### LINKED ISSUE #5025: [cubecl] Device handle custom channel reduces all failures to opaque `CallError`
Panics happening inside device channel tasks are currently swallowed by the custom channel implementation and converted into a generic `CallError, which then often gets `.unwrap()`ed higher up in CubeCL/Burn. This loses the original panic payload/backtrace and makes debugging extremely difficult.

Example:
```
thread 'DSD-0-0' (213305) panicked at /home/crutcher/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/cubecl-runtime-0.10.0/src/tune/tune_benchmark.rs:109:31:
called `Result::unwrap()` on an `Err` value: CallError
```
... for  a little bit. it does crash eventually
Lots of this in the log:
```
2026-05-29T16:54:51.812959Z ERROR burn_train::learner::application_logger: PANIC => panicked at /home/crutcher/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/burn-cubecl-fusion-0.21.0/src/engine/codegen/ir.rs:500:63:
index out of bounds: the len is 1 but the index is 1
2026-05-29T16:54:51.813139Z  WARN cubecl_common::device::handle::channel::task: Task failed: Any { .. }
2026-05-29T16:54:51.813151Z ERROR burn_train::learner::application_logger: PANIC => panicked at /home/crutcher/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/burn-fusion-0.21.0/src/stream/execution/ordering.rs:49:13:
Ordering is bigger than operations
```

The other device threads eventually also panic after the original error, producing additional downstream failures that obscure the root cause before the poisoned state eventually propagates.

Potential improvements:
- Preserve the original panic payload/backtrace instead of erasing it into `CallError`
- Re-panic on the caller side for blocking APIs (`submit_blocking`, `exclusive`, autotune/profile paths)
- Poison the device/channel after a background task panic, and have the poisoned state re-surface the original panic instead of only returning a generic `CallError` (?)

Andy2887:
Fixed tracel-ai/burn#5025

### Problem

For tasks that failed on the device thread, no useful debugging information is presented to the developers.

### Cause of the Problem

1. Logging `err` failed to produce the real error message because `err` is a pointer to the error payload, and printing the pointer gives us "Any { .. }" instead of the error message.
https://github.com/tracel-ai/cubecl/blob/a1aff5e5673f81dfb54df30ea8b43a32a57de162/crates/cubecl-common/src/device/handle/channel.rs#L463-L469

2. On the caller side, a `CallError` is returned if task failed on the device side. However, this `CallError` contains no information about the panic on the device side. Therefore, developers have no useful debugging information about the real cause of the failure.
https://github.com/tracel-ai/cubecl/blob/a1aff5e5673f81dfb54df30ea8b43a32a57de162/crates/cubecl-common/src/device/handle/channel.rs#L166-L171

### Solution
**Preserve the original panic payload in CallError**
`CallError` now has a field "payload" that carries the panic payload on the device thread.
Screenshot 2026-06-21 at 7 50 28 AM

The new design is we wrap the task inside `catch_unwind`, and we give the wrapper to "send" function. In this way, if the task failed, the panic is captured, sent to caller, and wrapped in `CallError`.

Screenshot 2026-06-21 at 8 06 46 AM

**Re-raise the panic payload on caller side for blocking APIs**

Additionally, an extension trait that re-raise the payload within `CallError` is created for the returned `Result<R, CallError>` of `submit_blocking`, `run_scoped` and `exclusive` functions.

The `unwrap_or_resume` function will unwrap `Result<R, CallError>`. If `CallError` is found, it re-raise the panic payload inside `CallError`.

Screenshot 2026-06-21 at 7 53 02 AM

We use `unwrap_or_resume` for the APIs in `ComputeClient`. Below is an example:
Screenshot 2026-06-21 at 8 49 04 AM

Note: `ComputeClient::exclusive` does not re-raise a panic payload if received. This is because `ComputeClient::exclusive` is used for benchmarking, and if a candidate kernel failed, we should just skip it instead of crashing the benchmark.
`ComputeClient::profile` only re-raise the panic in the profiled closure. A start profile or end profile error already returns `Result<_, ServerError>`, and we will keep the original design.


crates/cubecl-runtime/src/client.rs
nathanielsimard: I don't think this comment is up to date

nathanielsimard (approved):
A very small comment, but otherwise looks very good!

Andy2887:
Hi @nathanielsimard  I fixed the error log. I think the code is ready to be merged.

CLAUDE.md
```diff
@@ -0,0 +1,91 @@
+# CLAUDE.md
```
nathanielsimard: I would not commit that file for now.

Andy2887:
Sorry about that! It was not my intention to push the CLAUDE.md file. That was a mistake. I'll be more careful next time.

@nathanielsimard

PR #1378, opened 2026-06-11
refactor: Simplify `Variable` to align it with existing IRs

wingertge:
Simplifies the `Variable` type and renames it to `Value` to align it with existing IR representations and simplify the optimizer.
* `Builtin` and `Scalar` are now instructions (`ReadBuiltin` and `ReadScalar`) that output a normal value
* `ConstArray` is removed because it was broken and inconsistent the whole time, and didn't work the way I expected it to intially. If it comes back, it would be in the form of a const-initialized function local array (that's how it was implemented in Vulkan).
* `Matrix`, `BarrierToken` and `Aggregate` are now types, which they should've been the whole time
* `Versioned` has been redundant since switching to a global atomic count for the allocator
* `GlobalBuffer` and `TensorMap` are now normal values passed as inputs to the kernel function. This is still a bit messy because of the mixing between `KernelBuilder` and `Scope`, it should probably be refactored further so the scope itself keeps track of them.
* `LocalMut` and `Shared` have been replaced with `DeclareVariable`, which allocates the corresponding backing memory and returns a pointer handle to it. This is a bit awkward to generate in C++ and WGSL, but matches the way it works in all IR formats (including the ones used by the compilers for CUDA/Metal/Naga).

The one noteworthy remaining variant is `Constant`. This should be replaced with a `Constant` operation eventually, but that change makes const folding much harder without poperly tracking def-use chains. This will be done in a potential future migration to Pliron, which has native support for tracking them.

Also fixes an existing bug in HIP.

### Instructions


crates/cubecl-core/src/post_processing/unroll.rs
```diff
@@ -138,6 +125,73 @@ impl UnrollVisitor {
...
+                        return TransformAction::Ignore;
+                    }
+                }
+                Operation::DeclareVariable {
+                    value_ty,
+                    addr_space: AddressSpace::Shared,
+                    ..
+                } => {
+                    if value_ty.vector_size() > self.max_vector_size {
+                        todo!()
```
nathanielsimard: Should we fix the todo?

crates/cubecl-ir/src/variable.rs
```diff
@@ -557,70 +421,47 @@ impl Display for ConstantValue {
...
-            | VariableKind::GlobalScalar(id)
-            | VariableKind::LocalMut { id, .. }
-            | VariableKind::Versioned { id, .. }
-            | VariableKind::LocalConst { id, .. }
-            | VariableKind::ConstantArray { id, .. }
-            | VariableKind::Shared { id, .. }
-            | VariableKind::Matrix { id, .. } => Some(id),
-            _ => None,
+            ValueKind::Value { id, .. } => id,
+            _ => panic!(),
```
nathanielsimard: We could leave a better comment in the panic I believe

crates/cubecl-core/src/post_processing/unroll.rs
wingertge: Actually I realized I overthought it, now that both shared and local variables are the same type just with different address spaces we can merge the code. So that code can just go.

crates/cubecl-ir/src/variable.rs
wingertge: Done

nathanielsimard (approved):

PR #1381, opened 2026-06-13
Keep CPU scheduled buffers alive

### LINKED ISSUE #1380: CPU buffer bindings can be reused before queued execution finishes
## Summary

`cubecl-cpu` can reuse a buffer allocation while a queued CPU write or kernel still holds a raw `BytesResource` pointer to that allocation. This shows up as nondeterministic CPU test failures under parallel test execution, especially when many test threads create stream ids that fold into the same scheduler stream slot.

## Observed behavior

A high-parallel local run on `upstream/main` reproduced several unrelated-looking CPU test failures with stale, zero, or overwritten values. One deterministic reproducer is:

1. Launch a CPU kernel on stream `0` using an input buffer.
2. Drop the input handle after launch, before the queued kernel executes.
3. Create a replacement buffer on stream `max_streams`, which maps to the same scheduler slot.
4. Read the output from stream `0`.

Before the fix, the queued kernel can read the replacement data instead of its original input.

## Expected behavior

Queued CPU tasks should keep the managed memory binding alive for every scheduled resource until the task has finished executing, so the memory pool cannot reuse the allocation while the raw CPU resource is still in use.

## Validation

The failing reproducer was added as a regression test and passes after retaining the managed binding in scheduled CPU resources. The full `cubecl-cpu` lib tests also pass with `--test-threads=64` after the fix.

shinaoka:
Fixes #1380.

## Summary

- keep managed memory bindings alive for CPU scheduled writes and kernel bindings
- add mutable access for `ManagedResource` so queued CPU writes can retain ownership while mutating the resource
- add a deterministic regression test for scheduler-slot stream reuse overwriting a queued kernel input

## Root cause

`cubecl-cpu` scheduled tasks stored `BytesResource` values without holding the associated managed memory binding. When a launch consumed an input handle and the scheduler had not executed the queued task yet, the memory pool could consider that allocation free. A later allocation on a stream that maps to the same scheduler slot could reuse and overwrite the memory while the queued CPU kernel still held the old raw pointer.

## Validation

- `cargo +stable test -p cubecl-cpu queued_cpu_kernel_keeps_buffer_bindings_alive_until_execution --lib -- --nocapture --color=always`
- `cargo +stable test -p cubecl-cpu --lib -- --test-threads=64 --color=always`
- `cargo +stable xtask check audit`
- `cargo +stable xtask check format`
- `cargo +stable xtask check lint`
- `typos`

### DIFF
--- a/crates/cubecl-cpu/src/compiler/mlir_data.rs
+++ b/crates/cubecl-cpu/src/compiler/mlir_data.rs
@@ -105,7 +105,7 @@ impl MlirData {
         };
 
         for resource in resources {
-            let (ptr, len) = resource.get_write_ptr_and_length();
+            let (ptr, len) = resource.resource().get_write_ptr_and_length();
             let line_memref = LineMemRef::new(ptr, len);
             push_undirected(line_memref);
         }
--- a/crates/cubecl-cpu/src/compute/queue.rs
+++ b/crates/cubecl-cpu/src/compute/queue.rs
@@ -8,7 +8,10 @@ use crate::{
 };
 use cubecl_common::bytes::Bytes;
 use cubecl_core::{CubeDim, stream_id::StreamId};
-use cubecl_runtime::{logging::ServerLogger, storage::BytesResource};
+use cubecl_runtime::{
+    logging::ServerLogger,
+    storage::{BytesResource, ManagedResource},
+};
 use std::collections::HashMap;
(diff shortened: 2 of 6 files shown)


marcantoinem (approved):
LGTM, thanks for finding that, it's the kind of thing that is hard to find.

shinaoka:
GPT5.5 (xhigh) located the root cause in a few minutes :)

PR #1382, opened 2026-06-15
Fix handling one tuple

akiradeveloper:
CubeCL compiler can't handle one tuple `(T,)`. This PR add a test which fails the case and a subsequent fix.

```
cargo test -p cubecl-cpu test_assign_one_tuple --no-run

error[E0609]: no field `0` on type `element::base::NativeExpand<F>`
--> crates/cubecl-core/src/runtime_tests/assign.rs:17:26
|
17 |         output[0] = item.0;
|                          ^ unknown field
|
= note: available fields are: `expand`, `_type`
```


wingertge (approved):
So the whole time the the difference between Rust treating a single-element tuple as a tuple vs a parenthesized type was the comma... Good to know. I removed those because of Rust being annoying about it, but I guess a trailing comma is the fix.

PR #1384, opened 2026-06-16
fix(memory): resolve stale page lookups instead of panicking

nicolaslara:
## Summary

While trying to benchmark on burn-lm's continuous-batching server (tracel-ai/burn-lm#57), I ran into an issue where as soon as that server handles two or more concurrent requests, the cubecl-cuda dispatch thread (`DSD-0-0`) panics with `couldn't find resource for that handle: Memory page N doesn't exist` and takes the inference worker down with it. It made batched serving unusable at concurrency >= 2.

It looks like a concurrency bug but isn't: a device's compute server and all its per-stream memory managers run on one runner thread (`update_shared_bindings` / `find` / `cleanup` are all `&mut self` there), so there's nothing to race. It's a single-threaded stale-page-index bug that batched serving just happens to trigger.

## The issue

* A memory binding caches a page index in its `ManagedMemoryDescriptor`. 
* `ExclusiveMemoryPool::cleanup` deallocates free pages and renumbers the survivors, but writes the new index only into each page's current `slice.handle`. 
* A binding whose descriptor has diverged from that handle (which the cross-stream cursor lookup in batched serving keeps alive) is left holding a stale index, so `ExclusiveMemoryPool::find` then either 
    - indexes out of bounds (-> `NotFound` -> cubecl-cuda's `handle_cursor` `.unwrap()` aborts the dispatch thread) or
    -  lands in-bounds on the wrong survivor (-> the `assert_eq!` in `MemoryManagement::find` panics). 
 * In normal use this can't happen: the handle you keep alive is the very one the pool renumbers, so every binding made from it picks up the corrected page index for free. The crash only appears when a binding is left pointing at an older descriptor the pool has since stopped updating.

## The fix

- `ExclusiveMemoryPool::find` now treats the cached page as a hint, validate it by `ManagedMemoryId`, and return a clean `NotFound` for a stale/diverged binding instead of indexing blindly (one `usize` compare on the hot path).
- `MemoryManagement::find` replaces the `assert_eq!` with a graceful `NotFound`, so a stale binding can never resolve to a different buffer or panic there.
- cubecl-cuda `handle_cursor` doesn't `.unwrap()` a failed lookup (that's what aborts the dispatch thread). Its value only decides whether one stream must wait for another, so on a miss we return `u64::MAX`, which always errs toward adding the wait: the worst case is a redundant synchronization, never a skipped one that would let the GPU read a half-written buffer.

## How it was reproduced and proven

- Deterministic, no GPU: two single-threaded `cubecl-runtime` unit tests reproduce both panic shapes on clean `main` and pass after the fix (full suite + `cargo check -p cubecl-cuda` stay green).
- Live A/B on a Modal T4. I deployed the real burn-lm server (Llama-3.2-1B) twice — unfixed `ba103c7`, and the same tree with this fix on top, nothing else changed, and ran the identical concurrent chat load against both. The crash only reproduced reliably once all the load was pinned to a single GPU container; left to autoscale, the platform spread it thin enough to mask the bug. Counting dispatch-thread panics in each run's logs:

| | Unfixed `ba103c7` | Fixed |
|---|---|---|
| page-race assert panics (`memory_manage.rs`, every one a `pool:0 page:0` descriptor) | 161 | 0 |
| downstream out-of-memory panics (`server.rs` `reserve().unwrap()`) | 794 | 0 |
| total dispatch-thread panics | ~955 | 0 |

Same code, same load, only the fix differs. The 794 out-of-memory panics are a knock-on effect of the same bug: once the pool's page bookkeeping is corrupted it mis-accounts free space and asks for impossible allocations (several requesting ~1.5 GB for a 1B model), so fixing the lookup clears those too, and the server runs measurably faster for no longer thrashing a broken pool.

### DIFF
--- a/crates/cubecl-common/Cargo.toml
+++ b/crates/cubecl-common/Cargo.toml
@@ -19,6 +19,11 @@ workspace = true
 
 
 [features]
+# Enables runtime backtrace capture outside of `cfg(test)` (see build.rs).
+# Implies `std` since the `backtrace` crate rides on it. Used by diagnostic
+# builds (e.g. cubecl-runtime's `debug-mem`) that need real producer/consumer
+# stacks in a non-test binary.
+backtrace-capture = ["std"]
 cache = ["std", "serde_json", "dirs", "sanitize-filename"]
 compilation-cache = ["cache", "dep:ciborium"]
 default = ["std"]
--- a/crates/cubecl-common/build.rs
+++ b/crates/cubecl-common/build.rs
@@ -12,6 +12,7 @@ fn main() {
         // Filesystem and environment access for config loading and caching.
         std_io: { all(feature = "std", any(target_os = "windows", target_os = "linux", target_os = "macos", target_os = "android")) },
         // TODO: We can't yet activate it for everything because of how error handling is done in matmul.
-        backtrace: { all(test, feature="std") },
+        // `backtrace-capture` widens this to non-test diagnostic builds (e.g. `debug-mem`).
+        backtrace: { all(feature="std", any(test, feature="backtrace-capture")) },
     }
 }
--- a/crates/cubecl-cuda/Cargo.toml
+++ b/crates/cubecl-cuda/Cargo.toml
@@ -23,6 +23,10 @@ default = [
     "cubecl-core/default",
 ]
 ptx-wmma = []
+# Forwards to `cubecl-runtime/debug-mem`; diagnostic builds only. See that
+# feature for details. Enable on the Modal/burn-lm patch build to capture the
+# orphaning-bind backtrace behind the "Memory page N doesn't exist" crash.
+debug-mem = ["cubecl-runtime/debug-mem"]
 std = [
     "cubecl-runtime/std",
     "cubecl-common/std",
--- a/crates/cubecl-cuda/src/compute/stream.rs
+++ b/crates/cubecl-cuda/src/compute/stream.rs
@@ -148,10 +148,23 @@ impl EventStreamBackend for CudaStreamBackend {
     }
 
     fn handle_cursor(stream: &Self::Stream, binding: &Binding) -> u64 {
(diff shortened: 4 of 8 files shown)


nathanielsimard (commented):
I'm unsure I understand the problem. A binding whose ID isn't valid should not happen, even with multiple streams/threads. Cleanup should consider the number of bindings, not just the handle, that's what I mean.

nicolaslara:
So, I actually don't understand what the underlying cause of the bug is (and this only fixes the symptoms) but I was running into the `Memory page N doesn't exist` error when trying to run burn-lm (on modal), so I think somewhere in the stack the equivalent of [this test](https://github.com/tracel-ai/cubecl/pull/1384/changes#diff-d0c4e23b50bf5ee18251061c695d712f3aeff1dd090030a4760e911dc9d30244R1148-R1175) is happening. 

The idea of this PR is that if that does happen, then we fail gracefully. 

I can try to debug the actual cause, but it may take a bit longer.  My initial thought is getting a minimal repro on the server and getting a stack trace where the error happens (though it might not be enough to know where the binding was aquired)

nathanielsimard:
I prefer to fix the root cause of the problem rather than hide the error. Let's keep the panic until we understand the root cause, then maybe we can relax how we fail.

nicolaslara:
Makes sense! I'll try to isolate it.

Note that the idea here is not hiding the error, but forcing a recovery. But yeah, much better to find the root cause.

nicolaslara:
I couldn't replicate this no matter how hard I tried (though it was consistently happening when I opened the PR). 

I'll close this, but leave my debugging logs in in ([e8ac549](https://github.com/tracel-ai/cubecl/pull/1384/commits/e8ac549c709d2f8c1346cba14912fff48201c44d)) case someone ends up here in the future.

PR #1387, opened 2026-06-19
Enable min, max, bitwise, and compare-exchange atomics for i32 and u32

LucaCappelletti94:
cubecl gates each atomic operation behind a per-backend list of what that backend allows. For integers, the WebGPU (`wgsl`) and shared C++ backends only listed loads, stores, and addition, so calling `min`, `max`, a bitwise op, or compare-and-exchange on a `u32` failed at compile time. The backends could already generate that code and the hardware already runs it, so the only thing missing was the entry in the list.

This adds those operations for the 32-bit integer types `i32` and `u32` on both backends. The WebGPU side needs no device feature for it, and the single C++ change covers both `cubecl-cuda` and `cubecl-hip`.

The known-answer tests for these operations already existed but were being skipped on both backends, so they now run and pass. I also added a test where many threads race a `max` into the same cell.

### DIFF
--- a/crates/cubecl-core/src/runtime_tests/atomic.rs
+++ b/crates/cubecl-core/src/runtime_tests/atomic.rs
@@ -257,6 +257,40 @@ pub fn test_kernel_atomic_int<R: Runtime, I: Int + CubeElement>(
     }
 }
 
+#[cube(launch)]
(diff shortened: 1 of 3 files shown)


wingertge (approved):

LucaCappelletti94:
Ping?

PR #1389, opened 2026-06-19
burn element trait related changes for `e4m3`, `e5m2`

skewballfox:

[related PR for burn](https://github.com/tracel-ai/burn/pull/5096)

### Instructions


crates/cubecl-common/src/float/fp8/fp8_e4m3.rs
```diff
@@ -68,6 +98,12 @@ impl e4m3 {
         self.to_f64() as f32
     }
 
+    /// check if an [`e4m3`] value is Nan
+    #[inline]
+    pub fn is_nan(self) -> bool {
+        self.0 == 0x7Fu8 || self.0 == 0xFFu8
```
ThierryCantin-Demers: Could directly use `F8E4M3::is_nan`

crates/cubecl-common/src/float/fp8/fp8_e5m2.rs
```diff
@@ -59,6 +89,12 @@ impl e5m2 {
         self.0
     }
 
+    /// check if an [`e4m3`] value is Nan
```
ThierryCantin-Demers: Comments here refer to `e4m3` instead of `e5m2`

crates/cubecl-common/src/float/fp8/fp8_e5m2.rs
```diff
@@ -76,6 +112,60 @@ impl e5m2 {
     pub const fn to_f64(self) -> f64 {
         F8E5M2::from_bits(self.0).to_f64()
     }
+
+    /// Compares [`e4m3`] values
```
ThierryCantin-Demers: `e4m3` -> `e5m2`

crates/cubecl-common/src/float/fp8/fp8_e4m3.rs
```diff
@@ -76,6 +112,58 @@ impl e4m3 {
     pub const fn to_f64(self) -> f64 {
         F8E4M3::from_bits(self.0).to_f64()
     }
+
+    /// Compares [`e4m3`] values
+    pub fn total_cmp(self, other: Self) -> Ordering {
```
ThierryCantin-Demers, line 115: Could be `#[inline]` like it is for `e5m2`

crates/cubecl-common/src/float/fp8/fp8_e4m3.rs
```diff
@@ -18,10 +20,38 @@ pub struct e4m3(u8);
...
+    pub const MIN_POSITIVE: Self = Self::from_bits(F8E4M3::MIN_POSITIVE.to_bits());
+
+    ///Approximate number of significant digits in base 10
+    pub const DIGITS: u32 = F8E4M3::DIGITS;
+
+    ///Number of mantissa digits
+    pub const MANTISSA_DIGITS: u32 = F8E4M3::MANTISSA_DIGITS;
+
+    ///Positive infinity ∞
+    pub const INFINITY: Self = Self::from_bits(F8E4M3::INFINITY.to_bits());
```
ThierryCantin-Demers: The `e4m3` spec doesn't seem to actually support infinity and -infinity and the fact that `float8` has it looks like a quirk of the lib. I don't know if we want to reexpose this quirk, is it currently needed?

crates/cubecl-common/src/float/fp8/fp8_e5m2.rs
```diff
@@ -59,6 +89,12 @@ impl e5m2 {
         self.0
     }
 
+    /// check if an [`e4m3`] value is Nan
+    #[inline]
+    pub fn is_nan(self) -> bool {
+        [0x7D, 0x7E, 0x7F, 0xFD, 0xFE, 0xFF].contains(&self.0)
```
ThierryCantin-Demers, line 92: These are not the same values that `float8` has, but I guess it is a problem with the lib itself (https://github.com/EricLBuehler/float8/issues/8) so I think this is good

crates/cubecl-common/src/float/fp8/fp8_e4m3.rs
```diff
@@ -76,6 +112,58 @@ impl e4m3 {
...
     }
+
+    /// Compares [`e4m3`] values
+    pub fn total_cmp(self, other: Self) -> Ordering {
+        F8E4M3::total_cmp(&self.into(), &other.into())
+    }
+}
+impl Zero for e4m3 {
+    fn zero() -> Self {
+        Self::from_bits(F8E4M3::ZERO.to_bits())
```
ThierryCantin-Demers: Would be good to have a `ZERO` and `ONE` constants for `e4m3` so these could just return the values instead of reconstructing everytime. Same for `e5m2`.

crates/cubecl-common/src/float/fp8/fp8_e4m3.rs
```diff
@@ -76,6 +112,58 @@ impl e4m3 {
...
+    fn one() -> Self {
+        Self::from_bits(F8E4M3::ONE.to_bits())
+    }
+
+    fn is_one(&self) -> bool {
+        self == &Self::one()
+    }
+}
+impl Num for e4m3 {
+    // You can use the standard library's ParseFloatError
```
ThierryCantin-Demers: Seems to be leftover AI comments, would be better if removed

ThierryCantin-Demers: Same for `e5m2`

crates/cubecl-common/src/float/fp8/fp8_e4m3.rs
```diff
@@ -76,6 +112,58 @@ impl e4m3 {
...
+        F8E4M3::total_cmp(&self.into(), &other.into())
+    }
+}
+impl Zero for e4m3 {
+    fn zero() -> Self {
+        Self::from_bits(F8E4M3::ZERO.to_bits())
+    }
+
+    fn is_zero(&self) -> bool {
+        self == &Self::zero()
```
ThierryCantin-Demers: I don't know if `e4m3` and `e5m2` also have a negative zero value representation, but it should be checked here if so

ThierryCantin-Demers (changes requested):
Nice job, just a few minor comments. Would also be nice to make sure the spacing between impl blocks is consistent with the rest

crates/cubecl-common/src/float/fp8/fp8_e4m3.rs
skewballfox: so looking it up online, for e5m2 it's in the spec. for e4m3, it seems mixed. There seems to be 2 variants of the spec: FN (used by nvidia and intel) and UZ. The UZ variant interprets negative 0 as NaN. Some places online seem to implicitly reference the UZ variant when talking about e4m3

actually, this might be a bigger issue. [rocm](https://rocm.docs.amd.com/en/docs-6.3.3/reference/precision-support.html) seems to support only the UZ variant, and they have different representable ranges. There might need to be some kind of mapping layer where we expose FN but convert to UZ for a target that has UZ but not FN.

skewballfox:
@ThierryCantin-Demers I think this is ready for another review round.

ThierryCantin-Demers (approved):

crates/cubecl-common/src/float/fp8/fp8_e4m3.rs
ThierryCantin-Demers: I think for now we can have it like this and open an issue to change it later on

wingertge (approved):

PR #1390, opened 2026-06-20
Fix `into_contiguous` zeroing packed output when the repacked dim vectorizes

LucaCappelletti94:
`into_contiguous_packed_ref` passes `elems_per_unit` to `copy_kernel_packed`, but the kernel expects `num_elems_per_unit`, which is `vector_size * elems_per_unit`. The kernel divides that by `vector_size` to get `vectors_per_thread`, so when `vector_size > elems_per_unit` the count truncates to zero, the write loop never runs, and the freshly allocated output is left all zeros. When it does not truncate it still under-writes by a factor of `vector_size`. The helper was only correct when the output vector size is 1.

This affects any `into_contiguous` of a `PackedU32` or `PackedNative` tensor whose repacked last dim vectorizes while the launcher picks a small `elems_per_unit`, so packed layouts can be silently wrong. It was first noticed while carrying codebook-quantized operands through matmul, but it is general.

The fix passes `num_elems_per_unit` and drops the dead `elems_per_unit` reduction step. The added cpu tests repack a `PackedU32` tensor and compare against a reference packing across the relevant regimes, including `vector_size 1` as a no-regression guard, the original `vector_size` greater than `elems_per_unit` zeroing case, the halving reduction of `num_elems_per_unit`, and a large layout where `vectors_per_thread` exceeds 1 so the multi-vector write loop is exercised. All but the `vector_size 1` guard fail without the fix.

### DIFF
--- a/crates/cubecl-cpu/src/lib.rs
+++ b/crates/cubecl-cpu/src/lib.rs
@@ -18,6 +18,7 @@ mod tests {
     cubecl_core::testgen_all!(f32: [f16, f32, f64], i32: [i8, i16, i32, i64], u32: [u8, u16, u32, u64]);
     cubecl_std::testgen!();
     cubecl_std::testgen_tensor_identity!([f16, f32, u32]);
+    cubecl_std::testgen_tensor_into_contiguous!();
     cubecl_std::testgen_quantized_view!(f32);
 
     #[cube(launch)]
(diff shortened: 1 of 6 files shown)


wingertge (approved):

LucaCappelletti94:
Ping?

ThierryCantin-Demers (approved):

PR #1395, opened 2026-06-25
refactor: replace `dirs` with `etcetera` (drops MPL-2.0 `option-ext`) = fully permissive stack

shimwell:
## What

Replace the `dirs` dependency with `etcetera` in `cubecl-common` and `cubecl-runtime`.

## Why

`dirs` → `dirs-sys` → **`option-ext`**, which is licensed **MPL-2.0** (weak copyleft). `dirs` and `dirs-sys` are themselves MIT/Apache, but pulling in `dirs` drags `option-ext` into the tree transitively (it's an unconditional dependency of `dirs-sys`). Downstream consumers that want a strictly permissive (MIT/Apache-2.0) dependency tree inherit this through cubecl.

[`etcetera`](https://crates.io/crates/etcetera) (MIT OR Apache-2.0) provides the same user-directory lookups without `option-ext`. Its only dependencies are `cfg-if` (and `windows-sys` on Windows).

## Changes

Three first-party call sites:

| Crate | Before | After |
|---|---|---|
| `cubecl-runtime` (`config/cache.rs`) | `dirs::cache_dir()` | `choose_base_strategy().cache_dir()` |
| `cubecl-runtime` (`config/cache.rs`) | `dirs::config_local_dir()` | `choose_base_strategy().config_dir()` |
| `cubecl-common` (`cache.rs`) | `dirs::home_dir()` | `etcetera::home_dir()` |

Plus the matching workspace/feature wiring (`dirs` → `etcetera`, kept optional under the same `cache` / `std` features).

## Behaviour

- **Linux**: identical — `~/.cache`, `~/.config`, `$HOME`.
- **Windows**: cache dir unchanged (`LocalAppData`); the `Global` config dir moves from `LocalAppData` to `RoamingAppData` — both standard per-user config locations.
- **macOS**: `choose_base_strategy` uses the XDG convention (`~/.cache`, `~/.config`) rather than `dirs`' native `~/Library/...`. This is etcetera's documented CLI/dev-tool convention; happy to switch to `choose_native_strategy()` if the native macOS directories are preferred.

`etcetera` 0.11 requires Rust 1.87 — the workspace MSRV (1.95) already satisfies this.

## Verification

- `cargo check` / `cargo clippy` clean on `cubecl-common --features cache` and `cubecl-runtime`.
- `cubecl-common` cache unit test (`cache::tests::test_cache_simple`) passes.
- `cargo fmt --check` clean.

## Validation against burn

validated against `burn` :heavy_check_mark:


nathanielsimard (approved):

shimwell:
Many many thanks, much appreciated. Thanks so much for making this package and delighted we can make it fully permissive.

PR #1399, opened 2026-06-29
Explicit Stream Creation

Andy2887:
Related to tracel-ai/burn#4991

**This is the 1st of the 2 PRs to fix the issue.**

## Problem
Threads are automatically assigned to a stream on creation. Currently, there is no way to explicitly specify what stream id a thread should use.

## Solution
Allow the user to explicitly set the stream id of a thread.

Example:

Thread A is on stream A. User wants to perform some tensor operations on stream B instead.
1. Thread A calls the function that sets itself to target stream B **(the function we need to implement)**.
2. Perform tensor operations.

## Implementation in CubeCL
1. Added `StreamId::new()` function that return a `StreamId` object with a globally-unique stream id.

Screenshot 2026-06-30 at 4 22 28 PM

The key idea is, if the user wants a new stream, `StreamId::new()` will generate the id for a new stream. 

Note: the user could also manually construct a `StreamId` object.

```rust
let stream = StreamId { value: 0 };
```

This is not recommended. Theoretically, users should only create `StreamId` through `new()` - it is the legit constructor - and clone it and pass it to other threads if multiple threads wish to share the same stream.

3. Changed `StreamId::swap(StreamId)` to `StreamId::set_current(StreamId)` and dropped `unsafe`.
Semantically, the previous `swap` function sets the stream id of a thread. Changing it to `set_current` makes the intent more explicit.

`unsafe` is dropped. Previously, `unsafe` is added because it was assumed a thread's stream shouldn't change. Since we want explicit stream selection for threads, it is no longer considered "unsafe" to set the stream for a thread.

Screenshot 2026-06-30 at 8 22 10 PM

The same logic applies to `ComputeClient` class. `unsafe` keyword is dropped for `set_stream` function.

Screenshot 2026-06-30 at 9 37 57 PM

4. The logic of `StreamId::current()` does not change. 

5. Steam id storage design on each thread is refactored
Previously, `ID`, which stores stream id, only existed under `#[cfg(multi_threading)]`. For any other target (like no_std, wasm), there was no stream id and counter at all. Thus it was not possible to support multiple streams on wasm targets.
https://github.com/tracel-ai/cubecl/blob/ac24631b40347bd962e5584cc0ad1ede9ce2d087/crates/cubecl-common/src/stream_id.rs#L12-L18
Functions like `current()` are essentially no-ops on single threaded targets.
https://github.com/tracel-ai/cubecl/blob/ac24631b40347bd962e5584cc0ad1ede9ce2d087/crates/cubecl-common/src/stream_id.rs#L48-L56
The refactored stream id storage (see `mod ambient`) enables single threaded targets to also store stream id. This enables multiple stream support for wasm.

## Example Flow
Assume the user wants to have 2 threads accessing 1 stream. The user first call `StreamId::new()` object to generate a StreamId object (we call it `stream`). Then, on each of the thread, `StreamId::set_current(stream)` is called to set the stream id of that thread to equal to the stream id we just created.


nathanielsimard:
unsafe IS the right move here, changing the stream of an existing stream may break a dependency cycle and lead to data corruption. Setting a stream doesn't change that! I agree that having more control over how streams are created is a valuable feature, I just don't think this is the right implementation. The way we should support that is having a stream global runtime where you can spawn streams that may be backed by threads, sequential execution or tokio workers. Right now what I would use for your use case is the following:

```rust
let stream = StreamId { value: MY_STREAM_ID };
stream.execute(|| {
   // my code 
});
```

This is the current way of manually setting a stream for a block of compute.

PR #1414, opened 2026-07-09
refactor(benchmark): make Benchmark::sync fallible

dcvz:
On asynchronous backends, kernel launches are fire-and-forget: compilation/validation errors and GPU faults often only become observable at synchronization. Since `sync()` returned `()`, an implementor's only option was `.unwrap()`. That means a single benchmark whose kernel legitimately can't run on the current device (e.g. requesting 40 KiB of shared memory against Apple's 32 KiB limit) aborts the entire benchmark binary instead of failing that one bench. In practice this made full cubek benchmark sweeps impossible to complete on macOS.

`sync` now returns `Result<(), String>`, propagated through `profile_full` and `run()`. Two `run()` changes close holes the signature alone wouldn't fix:
- warmup errors are propagated instead of discarded. Previously a queued JIT-compilation error was consumed invisibly by a warmup sync, and the benchmark reported timings for a kernel that never ran;
- the stream is drained after the samples. Previously, under device timing (where `profile()` overrides bypass the trait's `sync`), a fault from the final samples surfaced at the _next_ benchmark's leading sync and failed the wrong one.

The first commit fixes Metal fault recording during sync/profiling (faults recorded by completion handlers during the fence wait were dropped or misattributed). Standalone fixes, but without them the trait's guarantee wouldn't hold on Metal.

Migration is one line per impl: `block_on(client.sync()).unwrap() → block_on(client.sync()).map_err(|e| format!("{e}"))`. The getting-started examples are updated.

### DIFF
--- a/crates/cubecl-metal/src/compute/server.rs
+++ b/crates/cubecl-metal/src/compute/server.rs
@@ -529,9 +529,25 @@ impl ComputeServer for MetalServer {
             Ok(r) => r,
             Err(e) => return Box::pin(async move { Err(e) }),
         };
-        let fence = MetalStreamBackend::flush(resolved.current());
+        let stream = resolved.current();
+        let fence = MetalStreamBackend::flush(stream);
+        let error_sink = stream.errors.clone();
 
-        Box::pin(async move { MetalStreamBackend::wait_event_sync(fence) })
+        Box::pin(async move {
+            MetalStreamBackend::wait_event_sync(fence)?;
+            // Completion handlers may record faults between the pre-wait drain above
+            // and the fence resolving; surface them on this sync rather than leaving
+            // them to poison a later, unrelated call.
+            let errors = core::mem::take(&mut *error_sink.lock().unwrap());
+            if errors.is_empty() {
+                Ok(())
+            } else {
+                Err(ServerError::ServerUnhealthy {
+                    errors,
+                    backtrace: cubecl_common::backtrace::BackTrace::capture(),
+                })
+            }
+        })
     }
 
     fn flush(&mut self, stream_id: StreamId) -> Result<(), ServerError> {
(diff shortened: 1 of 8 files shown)


crates/cubecl-metal/src/compute/server.rs
```diff
@@ -529,9 +529,25 @@ impl ComputeServer for MetalServer {
...
         };
-        let fence = MetalStreamBackend::flush(resolved.current());
-
-        Box::pin(async move { MetalStreamBackend::wait_event_sync(fence) })
+        let stream = resolved.current();
+        let fence = MetalStreamBackend::flush(stream);
+        let error_sink = stream.errors.clone();
+
+        Box::pin(async move {
+            MetalStreamBackend::wait_event_sync(fence)?;
```
nathanielsimard, line 537: If `wait_event_sync` fails, the recorded faults stay in the sink and poison a later unrelated call. The drain should happen regardless of the wait's outcome.

crates/cubecl-metal/src/compute/server.rs
```diff
@@ -595,6 +612,17 @@ impl ComputeServer for MetalServer {
             buffer.waitUntilCompleted();
         }
 
+        // The completion handlers have now all run; surface any fault they recorded
+        // after the sync above drained the sink, so a mid-window GPU fault fails this
+        // profile instead of leaking into the next call on the stream.
+        let errors = self.flush_errors(stream_id);
+        if !errors.is_empty() {
+            return Err(ProfileError::Unknown {
```
nathanielsimard, line 620: `errors` is a `Vec<ServerError>`, and `ProfileError::Server(#[from] Box<ServerError>)` exists at base.rs:71 for exactly this. It should be `ProfileError::Server(Box::new(ServerError::ServerUnhealthy { errors, backtrace }))`. The pre-existing `flush_errors` at :160 does the same stringify, so both could be fixed together.

nathanielsimard (commented):
That's a good improvement. Is there a way to use the ServerError struct ins

PR #1422, opened 2026-07-13
Feat/autotune throughput

SamuelBelanger:
## Key changes

- Added/Fix throughput methods to retrieve operations per second for compute (direct or cmma), memory (direct global memory) and launch overhead.
- Added `AutotuneBound` to early exit autotuning on a problem when a strategy is under the threshold limit set by the user.


crates/cubecl-runtime/src/tune/tuner.rs
```diff
@@ -198,6 +198,17 @@ impl<K: AutotuneKey> Tuner<K> {
             _ => None,
         };
 
+        // The slowest median duration still considered close enough to peak throughput.
```
louisfd: Confusant

crates/cubecl-std/src/throughput/runners/compute_direct.rs
```diff
@@ -44,13 +49,27 @@ pub fn compute_direct_throughput<I: Numeric, N: Size>(
...
+        c.insert(lane, tid + offset);
+
+        s0.insert(lane, offset + I::cast_from(1));
+        s1.insert(lane, offset + I::cast_from(2));
+        s2.insert(lane, offset + I::cast_from(3));
+        s3.insert(lane, offset + I::cast_from(4));
+    }
 
     for _ in 0..n_iter {
         if use_fma {
```
louisfd: Could be cleaner encapsulate in one ops call

louisfd (approved):
LGTM

PR #1423, opened 2026-07-13
Disable persistent tune cache option

SamuelBelanger:
## Key Changes

- Added a `CUBECL_AUTOTUNE_CACHE` environment variable to explicitly bypass cache persistent reading and writing.
- Uses `in_memory_cache` per process.


louisfd (approved):
LGTM

PR #1424, opened 2026-07-13
fix(cpu): enforce max_units_per_cube loudly and fix sync_cube scheduling

louisfd:
The CPU runtime caps max_units_per_cube at the machine's core count (units are pinned one per worker thread so the sync_cube spin barrier can make progress). Three bugs surrounded that contract:

- Launches exceeding the cap were rejected into the stream error sink and never surfaced: reads returned the stale (usually zeroed) buffer, so downstream tests failed with silent wrong results instead of an error (misdiagnosed in cubek as a wgpu/Vulkan atomics regression). The server now validates cube dims at launch like every other backend, logs the refusal, and read drains the stream and returns any recorded error.

- sync_cube did not set needs_parallelism, so barrier kernels went through the load-balancing scheduler and two units of the same cube could queue behind each other on one worker, deadlocking the barrier whenever the pool wasn't idle.

- The shared runtime tests hardcoded 16-unit cube dims, which violates the (machine-dependent) cap on hosts with fewer than 16 cores; they now clamp to max_units_per_cube. The barrier wait loops also yield between iterations (only when actually waiting) so waiting units don't starve the rest of the process.

### DIFF
--- a/crates/cubecl-cpu/src/compiler/visitor/operation/synchronization.rs
+++ b/crates/cubecl-cpu/src/compiler/visitor/operation/synchronization.rs
@@ -425,6 +425,9 @@ impl<'a> Visitor<'a> {
     pub fn visit_synchronization(&mut self, synchronization: &Synchronization) {
         match synchronization {
             Synchronization::SyncCube => {
+                // The spin barrier only completes when every unit of the cube runs on its
+                // own thread; the scheduler must not queue two units behind each other.
+                *self.needs_parallelism = true;
                 let func_name = FlatSymbolRefAttribute::new(self.context, "sync_cube");
                 self.block.append_operation(func::call(
                     self.context,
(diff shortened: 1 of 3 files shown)


crates/cubecl-cpu/src/compiler/visitor/operation/synchronization.rs
marcantoinem, line 1: There's no need to add yielding to sync_cube, the goal of the scheduler is to always put sync_cube in parallel, that's the goal of needs_parallelism

crates/cubecl-cpu/src/compiler/external_function.rs
marcantoinem, line 1: Same things no need to add yielding

crates/cubecl-cpu/src/compute/server.rs
marcantoinem, line 1: It is normally launched as an error later, so the error handling is duplicated, maybe add the log to the other check?

marcantoinem (changes requested):

PR #1428, opened 2026-07-15
fix(msl): compile scalar reductions and atomic refs

ppodolsky:
## Summary

- emit scalar Metal SIMD reductions as direct scalar assignments instead of aggregate initialization
- preserve the `device` address space when assigning an indexed atomic buffer reference to a local
- add focused regression tests for both generated MSL forms

## Failure mode

A scalar plane reduction generated invalid MSL such as:

```metal
const float result = float {simd_sum(value)};
```

An indexed global atomic could also lose its address space when copied:

```metal
device atomic_float* indexed = &buffer[index];
atomic_float* local = indexed;
```

The patch emits a direct scalar `simd_sum` assignment and keeps `device atomic_float*` on the local declaration.

## Validation

- `cargo test -p cubecl-cpp`
- downstream Burn/CubeCL selective-scan forward and autodiff backward kernels compile and match an NdArray reference on Metal


ppodolsky:
Closing at downstream request. CubeCL main already contains the corrected scalar reduction and pointer address-space model; Hermes will evaluate moving the complete Burn/CubeCL stack forward instead of carrying a release/0.10 backport.

PR #1433, opened 2026-07-17
Add address_type=auto

akiradeveloper:
Add support for `#[cube(..., address_type = "auto")]`.

Automatic mode selects the concrete `AddressType` at launch time from the
largest indexable buffer length used by the kernel:

- `U32` when every buffer length fits in `u32`
- `U64` when at least one buffer exceeds `u32::MAX`

The resolved concrete address type is included in the existing kernel settings
and cache key.


wingertge:
I don't agree this is the correct way to go, because indices can easily go beyond the vectorized buffer length and cause overflow. A lot of kernels will multiply the scalar position by the strides and then only divide by vectorization at the end because it's a lot simpler. So this is a massive footgun.

PR #1435, opened 2026-07-17
CubeCL Environment: multiplatform HPC runtime API

nathanielsimard:
Introduces `cubecl-environment`, a compatibility layer centralizing sync/async/stream/config and cache persistence across std, no-std, wasm and tokio. Migrates `cubecl-runtime` and `cubecl-wgpu` onto it and unifies the tune/compilation caches on a single `Store`.

### Highlights
- Named environments with a key-value `Store` synced to SQLite (native), IndexedDB (wasm) or memory (no-std).
- Shippable bundles (`export`/`import`) of pre-warmed autotune and compilation caches.
- Config-driven environment activation and stream policy.


crates/cubecl-environment/src/bundle/manifest.rs
```diff
@@ -0,0 +1,213 @@
...
+    pub name: String,
+    /// The cubecl version the bundle was exported with. Entries are only
+    /// visible to the same version, exactly like local caches.
+    pub cubecl_version: String,
+    /// Creation time as seconds since the unix epoch, informational.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub created_unix_secs: Option<u64>,
+    /// The environments the bundle was captured on, informational in v1.
+    #[serde(default, rename = "environments")]
+    pub environments: Vec<EnvironmentInfo>,
```
laggui, line 101: `rename = "environments"` is redundant no?

laggui (commented):
Really nice generalization! Bundling and switching should be pretty useful

Just a minor comment I noticed among the big sea of changes 😄

PR #1440, opened 2026-07-18
feat(cuda): add resilient native and grouped GEMM dispatch

ppodolsky:
## Summary

This PR adds resilient native GEMM paths to CubeCL's CUDA runtime:

- reclaim unused cached pages and retry once after `BufferTooBig`
- add backend-optional single and grouped GEMM descriptors with capability flags
- implement asynchronous BF16 GEMM with FP32 accumulation through cuBLASLt
- implement heterogeneous grouped GEMM through `cublasGemmGroupedBatchedEx` on CUDA 12.5+
- validate layouts, dimensions, batch strides, binding ranges, overlap, zero-K behavior, and stream ownership before launch
- cache cuBLASLt descriptors and algorithms by shape, with separate workspaces per CUDA stream

The allocation retry was previously proposed separately in #1438. It is consolidated here so the CUDA accelerator integration has one review thread.

## Behavior and compatibility

Both GEMM primitives are optional runtime capabilities. Their server trait methods have default fail-loud implementations, so other backends require no changes. Callers can independently inspect `features.matmul.accelerated_gemm` and `features.matmul.accelerated_grouped_gemm` before dispatch.

CUDA advertises single BF16 GEMM only on SM80+. Grouped BF16 GEMM additionally requires CUDA 12.5. The grouped descriptor supports independent M, N, K, transpose state, and binding offsets per problem. Host metadata uses a bounded pinned staging ring guarded by CUDA events, so repeated and multi-stream launches remain asynchronous without overwriting in-flight arguments.

The allocator cleanup path runs only after storage reports `BufferTooBig`. It rechecks the selected pool after cleanup before attempting one new device allocation.

## Dependency validation

- [x] CubeK PR tracel-ai/cubek#428 pins this PR head and compiles against it.
- [x] Burn PR tracel-ai/burn#5190 pins both dependency PR heads.
- [x] Upstream `main` is the direct base of this branch.
- [x] Downstream Cargo manifests reference only `tracel-ai/*` repositories.

## Validation

- allocator retry regression test
- `cargo check -p cubecl-runtime`
- `cargo check -p cubecl-cuda --examples`
- A100 checks for padding, offsets, all transpose pairs, regular and broadcast batches, heterogeneous grouped shapes, repeated queued launches, asynchronous error propagation, and multi-stream ordering
- downstream CUDA linear forward and `dx`/`dW`/`db` parity under fusion, autotune, and autodiff
- 10,000 queued single GEMMs followed by one drain, confirming no per-call host synchronization

## Performance

On an A100-SXM4 40GB, BF16 native single-GEMM dispatch beat or matched PyTorch matmul on 14 of 17 transformer-training shapes. A heterogeneous grouped expert projection measured 0.290 ms versus 0.438 ms for the prior padded batched path; activation time is excluded from both measurements.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -154,6 +154,7 @@ cudarc = { version = "0.19.6", features = [
     "driver",
     "nvrtc",
     "cublas",
+    "cublaslt",
     "nccl",
     "fallback-dynamic-loading",
     "cuda-version-from-build-system",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -153,6 +153,7 @@ cudarc = { version = "0.19.6", features = [
     "std",
     "driver",
     "nvrtc",
+    "cublas",
     "nccl",
     "fallback-dynamic-loading",
     "cuda-version-from-build-system",
--- a/crates/cubecl-common/src/device/handle/tests.rs
+++ b/crates/cubecl-common/src/device/handle/tests.rs
@@ -28,21 +28,6 @@ fn test_concurrent_increment() {
     let count = context.submit_blocking(move |state| state.counter).unwrap();
     assert_eq!(count, thread_count);
 }
-
-#[test]
-fn test_flush_queue_dispatches_partial_batch() {
-    let device = TestDevice::<3>::new(0);
-    let context = DeviceHandle::<TestDeviceState<3>>::new(device.to_id());
-    let (sender, receiver) = std::sync::mpsc::channel();
-
-    context.submit(move |_state| sender.send(()).unwrap());
-    context.flush_queue();
-
-    receiver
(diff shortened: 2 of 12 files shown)


nathanielsimard:
Agreed that grouped GEMM is a gap in the cubecl/cubek/burn stack. That said, our goal is to avoid depending on pre-built binaries like cuBLAS wherever we can.

To that end, we're working on a tile API that should simplify cross-platform compatibility, cover all dtypes including quantized tensors, and compose with Burn's automatic fusion (both on-read and on-write).

We're happy to keep this PR around as a baseline to benchmark other implementations against, but we're trying hard to avoid the combinatorial explosion of kernels per platform × dtype × fusion group.

PR #1442, opened 2026-07-22
Explicit stream creation

Andy2887:
Related to tracel-ai/burn#4991

## Summary

Streams can now be created explicitly, and work can be pinned to a chosen stream.

Here are some examples:

```rust
use cubecl_common::stream::{Stream, StreamBacking};

// Example 1: fresh stream, run synchronously on the calling thread.
Stream::spawn(StreamBacking::Sequential, || { /* ... */ }, None).join().unwrap();

// Example 2: fresh stream, run on a dedicated worker thread.
let handle = Stream::spawn(StreamBacking::Thread, || model.forward(batch), None);
let output = handle.join().unwrap();

// Example 3: stream number 5, run synchronously on the calling thread.
// so several threads can deliberately share one pool.
Stream::spawn(StreamBacking::Sequential, || { /* ... */ }, 5).join().unwrap();
```

## Details

`cubecl-common/src/stream.rs` (new)

- `Stream::spawn(backing, f, stream)` — mimic the thread `spawn` API to pass work to a stream. 
- `StreamBacking` — `Sequential` runs the closure on current thread. `Thread` runs it on a fresh worker thread. `Sequential` is explicitly designed for non-multithreaded targets, and `Thread` allows a caller to fire and forget. 
- `StreamHandle<T>` with `join()` — mirrors `thread::JoinHandle` in thread API. 

`cubecl-common/src/stream_id.rs`

**The stream counter is no longer gated behind multi_threading.** Previously `STREAM_COUNT` only existed on std, non-wasm targets. Now non-threaded builds get real stream ids too — a global static `ID: AtomicU64`.
- `StreamId::fresh()` — mints a brand-new id from the global counter, instead of
returning the ambient one like current() does.
- `StreamId::from_number(n)` — maps a user-chosen number to a stream id.
_Note: there is a chance of collision with the stream id assigned by the counter._
```rust
// Thread A starts, touches the client, lazily gets implicit id 5.
std::thread::spawn(|| { client.create(&data); });  // -> StreamId { value: 5 }
// Somewhere unrelated, a user picks stream 5 for its own reason.
Stream::spawn(StreamBacking::Sequential, || client.create(&data), 5);  // -> StreamId { value: 5 }
```
- Previously `swap_single_thread` and `current` were no-ops for non-multithreaded build. Since now Non-threaded targets get a real stream id global slot (AtomicU64) instead of hardcoding stream 0, `swap`, `current` actually takes effect.

`stub.rs` and `Cargo.toml`

Since `AtomicU64` is now used both on multithreaded and non-multithreaded targets:

- `stub.rs`: re-export `AtomicU64`, falling back to portable-atomic where it's missing. It needs its own `target_has_atomic = "64"` gate rather than reusing the existing "ptr" one, because a target can have pointer-width atomics and still lack 64-bit ones (riscv32, armv7).
- `crates/cubecl-common/Cargo.toml`: add portable-atomic under `cfg(not(target_has_atomic = "64"))`.
- `Cargo.toml`: enable portable-atomic's fallback feature.

## Next Step

Update Burn repo to include relevant documentations on how to use the Stream API, and update burn to use this version of CubeCL


nathanielsimard:
With the new CubeCL environment (https://github.com/tracel-ai/cubecl/pull/1435) there is a new way of controlling how streams operate. There is a stream policy that can be configured. So if you want each async task to run on its own stream, you could:

```rust
use cubecl_environment::stream::{StreamPolicy, set_policy};

set_policy(StreamPolicy::PerTask);

tokio::spawn(async move {
    // This executes on a new stream.
});
```

`StreamId::current().executes` still works as before.

PR #1444, opened 2026-07-23
fix(wgpu/vulkan): deny cmma on Intel Xe3 GPUs (wrong results on Mesa ANV)

ThierryCantin-Demers:
Cooperative matrix matmuls compute wrong results on Intel Xe3 GPUs with Mesa ANV: coopmat loads from shared memory at distinct offsets all return the offset-0 tile, so every N-tile of a tiled matmul comes out as A x B_tile0. The generated SPIR-V, the driver's NIR and the final ISA are all correct, so this is an ANV-codegen-or-hardware issue, tracked at https://gitlab.freedesktop.org/mesa/mesa/-/issues/15932 (standalone reproducer attached there).

Add a device-id denylist (all Panther Lake SKUs, confirmed on Arc B390, plus Wildcat Lake which shares the Xe3 architecture) and skip cmma registration on those devices. This is a correctness-only gate: perf concerns on other Intel parts are left to autotune, which benchmarks cmma against the plain matmul and drops it when it loses. Xe2 (Battlemage, Lunar Lake) keeps cmma.

### Instructions

### DIFF
--- a/crates/cubecl-wgpu/src/backend/vulkan.rs
+++ b/crates/cubecl-wgpu/src/backend/vulkan.rs
@@ -394,7 +394,33 @@ fn register_features(
         props.memory.max_page_size = max_page_size;
     }
 
-    if extended_feat.cooperative_matrix.is_some() {
+    // VK_KHR_cooperative_matrix computes wrong results on Intel Xe3 GPUs with
+    // Mesa ANV: coopmat loads from shared memory at distinct offsets all
+    // return the offset-0 tile, so every N-tile of a tiled matmul comes out
+    // as A x B_tile0 (observed as garbage LLM output). The generated SPIR-V,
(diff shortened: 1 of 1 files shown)


ThierryCantin-Demers:
I will actually do a bit more investigating as it might be possible to do a less nuclear option than to completely disable CMMA. It could be that only some shapes of the available ones fail so a more surgical approach might be possible.

ThierryCantin-Demers:
The bug in mesa seems to be fixed with the following PR based on the issue I created: https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/43181. In the future, when a new release is made, the bug will be fixed and the temporary "fix" that this current `cubecl` PR introduces will be deprecated an even be a performance regression. I will close this PR and just wait for the new mesa release.

PR #1445, opened 2026-07-25
feat: add fallible dimension index conversion

mattisonchao:
## Motivation

Burn needs to validate negative and positive dimension indices without panicking immediately so invalid dimensions can retain its structured `TensorCheck` error formatting. `AsIndex` currently only exposes a panicking dimension-wrapping helper, even though `IndexWrap` already has a fallible API.

## Modifications

- Add a `try_dim_index` default method to `AsIndex`.
- Leave the existing `expect_elem_index` and `expect_dim_index` implementations unchanged.
- Cover positive, negative, referenced, and out-of-bounds dimension conversions.

## Testing

- `cargo +nightly test -p cubecl-zspace --lib`
- `cargo +nightly fmt --all -- --check`

### DIFF
--- a/crates/cubecl-zspace/src/indexing/type_conversion.rs
+++ b/crates/cubecl-zspace/src/indexing/type_conversion.rs
@@ -4,6 +4,7 @@
 //! various indexing operations.
 
 use super::wrapping::IndexWrap;
+use crate::errors::BoundsError;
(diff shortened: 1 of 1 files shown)


mattisonchao:
Hi @laggui 

could you please help review this?

laggui (approved):

PR #1457, opened 2026-07-31
Feat/dry run environment

nathanielsimard:
It gives a way to skip execution of kernels that aren't autotuning, so that we can pre-compile kernels in advance. With this PR you can dry-run a workflow before efficiently.


louisfd (approved):
LGTM

PR #1462, opened 2026-07-31
feat(tune): roofline bounds from a Work amount and per-resource thresholds

SamuelBelanger:
Autotune bounds are currently assembled by each consumer: measure the peaks, pick thresholds, build the two `AutotuneBound`s by hand. This moves the shared half into cubecl.

`Work` holds what a problem costs, in compute operations and bytes. `Thresholds` holds the target fraction of peak, one per resource, so a caller can hold its memory bound at peak while giving compute slack. `calculate_bounds` takes both instead of seven positional arguments.

`roofline_bounds` in cubecl-std measures the compute and memory peaks plus the launch overhead and assembles the `Bounds`, so a kernel family only has to state what its problem costs and which instruction it computes with.

Per key or per group threshold policies need nothing further: a `BoundsGenerator` already receives the autotune key and the inputs, so the thresholds it passes can be any function of them.

Note that the counts in `Work` must be under-estimates. The tuner keeps the first candidate at or below the time limit, so overstating a count buys a mediocre kernel an early exit, while understating one only costs a full tuning round.

Two new unit tests: thresholds land on the matching bound, and the default is 1.0.

### DIFF
--- a/crates/cubecl-runtime/src/tune/bounds_generator.rs
+++ b/crates/cubecl-runtime/src/tune/bounds_generator.rs
@@ -57,26 +57,62 @@ pub struct AutotuneBound {
     pub ops_count: usize,
 }
 
+/// Work required by a problem, specified in minimum compute operations and byte transfers.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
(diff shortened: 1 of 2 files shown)


ThierryCantin-Demers (approved):
LGTM

Issue #1459, opened 2026-07-31
Broadcasting a scalar half into a minifloat vector passes a scalar to a packed intrinsic


ThierryCantin-Demers:
`Vector::<e4m3, N>::cast_from(scalar)` emits CUDA that does not compile when the source is a scalar `f16` or `bf16`. This is the encode-direction mirror of #1458 and is independent of it: it reproduces on main and is unaffected by either fix there.

Found while probing the blast radius of #1458. It is confirmed **not** on the UE4M3 quantization path, because the value being quantized is always a vector, so it does not block that work.

## Repro

Add to `crates/cubecl-core/src/runtime_tests/minifloat.rs`:

```rust
#[cube(launch_unchecked)]
pub fn kernel_fp8_encode_broadcast<F: Float, N: Size>(input: &[F], out: &mut [Vector<u8, N>]) {
    if ABSOLUTE_POS == 0 {
        // A scalar float widened into every lane of a minifloat vector.
        out[0] = Vector::reinterpret(Vector::<e4m3, N>::cast_from(input[0]));
    }
}

pub fn test_fp8_encode_broadcast<R: Runtime, F: Float + CubeElement>(
    client: ComputeClient<R>,
    vector_size: VectorSize,
) {
    if !e4m3::supported_uses(&client).contains(TypeUsage::Conversion) {
        println!("Unsupported, skipping");
        return;
    }

    let data = as_type![F: 1.75];
    let handle_in = client.create_from_slice(F::as_bytes(&data[..]));
    let handle_out = client.empty(vector_size * size_of::<u8>());

    unsafe {
        kernel_fp8_encode_broadcast::launch_unchecked::<F, R>(
            &client,
            CubeCount::Static(1, 1, 1),
            CubeDim::new_1d(1),
            vector_size,
            BufferArg::from_raw_parts(handle_in.clone(), 1),
            BufferArg::from_raw_parts(handle_out.clone(), vector_size),
        )
    };

    let actual = client.read_one_unchecked(handle_out);
    // 0b0_0111_110 is 1.75 in e4m3.
    let expected = vec![0b0_0111_110u8; vector_size];
    assert_eq!(u8::from_bytes(&actual), &expected[..]);
}
```

Wired into `testgen_minifloat` at sizes 1, 2 and 4.

```
cargo test -p cubecl-cuda -- fp8_encode_broadcast
```

Size 1 passes. Size 2 fails for `f16` and `bf16` only; `f32` and `f64` pass.

[...]

PR #1463, opened 2026-07-31
fix(cuda): decode single minifloats on their own so scalar broadcasts compile

### LINKED ISSUE #1458: Broadcasting a scalar minifloat into a vector emits uncompilable CUDA: undefined intrinsic and undeclared temporaries
Broadcasting a scalar minifloat into a vector emits CUDA that does not compile. Two separate defects in `cast_minifloat_to_half`, both hit by the same expression, both present at main (`d8f07fd8`).

The practical effect is that **UE4M3 quantization cannot compile on CUDA at all**, at any quantization level, because `quantize_symmetric` divides by `Vector::cast_from(scale)` where the block scale is a scalar minifloat.

## Repro

Pure cubecl, no downstream. Add this to `crates/cubecl-core/src/runtime_tests/minifloat.rs`, alongside the existing minifloat tests:

```rust
#[cube(launch_unchecked)]
pub fn kernel_fp8_broadcast<F: Float, N: Size>(input: &[e4m3], out: &mut [Vector<F, N>]) {
    if ABSOLUTE_POS == 0 {
        // A scalar minifloat widened into every lane of a vector.
        out[0] = Vector::<F, N>::cast_from(input[0]);
    }
}

pub fn test_fp8_broadcast<R: Runtime, F: Float + CubeElement>(
    client: ComputeClient<R>,
    vector_size: VectorSize,
) {
    if !e4m3::supported_uses(&client).contains(TypeUsage::Conversion) {
        println!("Unsupported, skipping");
        return;
    }

    // 0b0_0111_110 is 1.75 in e4m3.
    let handle_in = client.create_from_slice(&[0b0_0111_110u8]);
    let handle_out = client.empty(vector_size * size_of::<F>());

    unsafe {
        kernel_fp8_broadcast::launch_unchecked::<F, R>(
            &client,
            CubeCount::Static(1, 1, 1),
            CubeDim::new_1d(1),
            vector_size,
            BufferArg::from_raw_parts(handle_in.clone(), 1),
            BufferArg::from_raw_parts(handle_out.clone(), vector_size),
        )
    };

    let actual = client.read_one_unchecked(handle_out);
    let actual = F::from_bytes(&actual);
    let expected = vec![F::from_int(0); vector_size]
        .iter()
        .map(|_| as_type![F: 1.75][0])
        .collect::<Vec<_>>();

    assert_eq!(actual, &expected[..]);
}
```

and wire it into `testgen_minifloat` in the same file:

```rust
#[$crate::runtime_tests::test_log::test]
fn test_fp8_broadcast() {
    let client = TestRuntime::client(&Default::default());
    for vector_size in [1, 2, 4] {
        cubecl_core::runtime_tests::minifloat::test_fp8_broadcast::<TestRuntime, FloatType>(
            client.clone(),
            vector_size,
        );
    }
}
```

```
cargo test -p cubecl-cuda -- fp8
```

```
test tests::f16_ty::test_fp8_broadcast  ... FAILED
test tests::bf16_ty::test_fp8_broadcast ... FAILED
test tests::f32_ty::test_fp8_broadcast  ... FAILED
test tests::f64_ty::test_fp8_broadcast  ... FAILED
test tests::f16_ty::test_fp8            ... ok
test tests::bf16_ty::test_fp8           ... ok
test tests::f32_ty::test_fp8            ... ok
test tests::f64_ty::test_fp8            ... ok
```

The existing `test_fp8` passes for all four float types; only the broadcast kernel fails. Vector size 1 passes, size 2 already fails.

Emitted source at size 2:

```cuda
__half2 _tmp_1 = __half2(__nv_cvt_fp8_to_halfraw2(val_13, __NV_E4M3));
__half_2 _tmp_0 = reinterpret_cast<__half_2&>(_tmp_1);
```

```
error: identifier "__nv_cvt_fp8_to_halfraw2" is undefined
error: identifier "__half_2" is undefined
```

## Defect 1: the intrinsic name does not exist

`cast_minifloat_to_half` (`cubecl-cpp/src/cuda/convert.rs`, ~line 273) derives `in_ty` from `in_opt.elem()` and `out_ty` from `out_opt.elem()` in two independent matches, then composes `__nv_cvt_{in_ty}_to_{out_ty}`. A scalar source with a 2-wide destination composes `__nv_cvt_fp8_to_halfraw2`.

`/usr/local/cuda/include/cuda_fp8.hpp` defines exactly two of these:

```
__nv_cvt_fp8_to_halfraw
__nv_cvt_fp8x2_to_halfraw2
```

So the only valid pairings are scalar to scalar and x2 to 2-wide. The scalar to 2-wide combination is not representable as written, and a broadcast is precisely that case. Converting once and then broadcasting would avoid it.

## Defect 2: the temporary vector types are never declared

`handle_unroll` (same file, ~line 339) builds `Value::tmp(out_opt)` in the optimized item type and then `reinterpret_cast`s to `out.item()`. Neither type is registered for declaration, so the emitted program uses `__half_2`, and at size 4 `__half2_2` and `__half_4`, without ever declaring the structs. Every other composite type in the same program (`float_4`, `uint32_4`, `array`, `info_st`) is declared normally at the top.

This is independent of defect 1 and would still apply to a genuine vector to vector minifloat cast where the optimized temporary type is not otherwise declared.

Note the two mask each other depending on width: at size 4 the parse error on the undeclared `__half2_2` aborts before nvrtc resolves the call name, so only defect 2 is reported. At size 2 `__half2` is a real CUDA type, the declaration parses, and both errors surface.

## Environment

- cubecl `d8f07fd8` (main)
- RTX 4070 Ti SUPER, driver 610.43.03, CUDA 13.3.1

## Downstream

This is the root cause behind tracel-ai/burn#5270, where UE4M3 quantization fails on cube backends. A failed launch there also poisons the shared per-device server, so unrelated tests on aliased streams fail afterwards.

ThierryCantin-Demers:
Closes #1458.

## The bug

`Vector::<F, N>::cast_from(scalar_e4m3)` does not compile. `special_cast` built the f16/bf16 intermediate at the destination's width regardless of the source's, so a scalar source with a wider destination composed `__nv_cvt_fp8_to_halfraw2`. CUDA declares the decode direction only in matched pairs, scalar to scalar and x2 to x2, so that symbol does not exist. The temporary vector type it decoded into was never declared either.

The practical effect is that UE4M3 quantization cannot compile on CUDA, since `quantize_symmetric` divides by `Vector::cast_from(scale)` with a scalar minifloat scale. The same asymmetry reaches fp4, fp6 and e8m0 through the same line.

## The fix

`cuda/convert.rs` decodes a single unpacked source on its own and lets the existing trailing assign broadcast it, one intrinsic call instead of N where every lane holds the same value:

```cuda
__half _tmp_0 = __half(__nv_cvt_fp8_to_halfraw(val_13, __NV_E4M3));
const float_4 val_14 = float_4{ float(_tmp_0), float(_tmp_0), float(_tmp_0), float(_tmp_0) };
```

A packed source already supplies one value per lane of the x2 intrinsic, so it keeps the destination's width and its emitted code is unchanged.

`shared/base.rs` also registers f16 and bf16 at the destination width. `special_cast` synthesizes `Item`s inside a `Display` impl, after the declared type set is frozen, so they cannot be registered where they are created.

## Tests

`test_fp8_broadcast`, the issue's repro, at vector sizes 1, 2, 4 across f16, bf16, f32, f64.

Verified on an RTX 4070 Ti SUPER (sm_89). With the fix, `cubecl-cuda` is 774 passed, 0 failed, 24 ignored. Reverting only the two source files and keeping the test gives 4 failures naming both defects:

```
error: identifier "__nv_cvt_fp8_to_halfraw2" is undefined
error: identifier "__half_2" is undefined
```

sm_89 is the only CUDA hardware available here, so fp4, fp6 and e8m0 conversion is unsupported and `test_fp4`, `test_fp6` and `test_scale` skip. Those paths get no hardware coverage. If you have Blackwell access, checking values there is the useful thing to do.

## Known gaps

#1459 covers two further shapes with the same root cause, a scalar float into a packed minifloat vector and a packed minifloat scalar into a narrower destination. Both are present on main and independent of this change.

### Instructions

The two unchecked boxes have nothing to submit: neither repo needs a source change. Both already pin cubecl at `05cac673`, this branch's base, so the only edit either needs is the routine rev bump after merge. `cubek-quant` on CUDA is 25 passed, 0 failed, identical to its own pin, and `burn-cuda` builds clean.

Worth stating plainly: a green cubek does not validate this fix. Its quant tests cover `q2f`, `q2s`, `q4f`, `q4s`, `q8f`, `q8s` only, every scheme built `.with_param(QuantParam::F32)`, with no `E4M3`, `E5M2`, `UE4M3` or `E2M1` in the test tree, so it cannot reach this code. `test_fp8_broadcast` is the meaningful check.

### DIFF
--- a/crates/cubecl-core/src/runtime_tests/minifloat.rs
+++ b/crates/cubecl-core/src/runtime_tests/minifloat.rs
@@ -90,74 +90,6 @@ pub fn test_fp8_broadcast<R: Runtime, F: Float + CubeElement>(
     assert_eq!(actual, &expected[..]);
 }
 
-/// The shape of a symmetric quantize/dequantize round trip with a minifloat scale, as used by
-/// UE4M3 quantization: a scalar scale broadcast into a vector in both directions, with the
-/// quantized values themselves stored as a minifloat vector.
-#[cube(launch_unchecked)]
-pub fn kernel_fp8_quantize<F: Float, N: Size>(
-    input: &[Vector<F, N>],
-    scale_in: &[F],
-    out_scale: &mut [u8],
-    out_q: &mut [Vector<u8, N>],
-    out_dq: &mut [Vector<F, N>],
-) {
-    if ABSOLUTE_POS == 0 {
-        let scale = e4m3::cast_from(scale_in[0]);
-        out_scale[0] = u8::reinterpret(scale);
-
-        let value = input[0];
-
-        let quantized = Vector::<e4m3, N>::cast_from(value / Vector::cast_from(scale));
-        out_q[0] = Vector::reinterpret(quantized);
-        out_dq[0] = Vector::cast_from(scale) * Vector::<F, N>::cast_from(quantized);
-    }
-}
-
-#[allow(clippy::unusual_byte_groupings, reason = "Split by float components")]
-pub fn test_fp8_quantize<R: Runtime, F: Float + CubeElement>(
-    client: ComputeClient<R>,
-    vector_size: VectorSize,
-) {
-    if !e4m3::supported_uses(&client).contains(TypeUsage::Conversion) {
-        println!("Unsupported, skipping");
-        return;
-    }
-
-    let data = as_type![F: 2.0, 4.0, 6.0, 8.0];
-    let scale = as_type![F: 2.0];
-    let handle_in = client.create_from_slice(F::as_bytes(&data[..vector_size]));
-    let handle_scale_in = client.create_from_slice(F::as_bytes(&scale[..]));
-    let handle_scale = client.empty(size_of::<u8>());
-    let handle_q = client.empty(vector_size * size_of::<u8>());
-    let handle_dq = client.empty(vector_size * size_of::<F>());
-
-    unsafe {
-        kernel_fp8_quantize::launch_unchecked::<F, R>(
-            &client,
-            CubeCount::Static(1, 1, 1),
-            CubeDim::new_1d(1),
-            vector_size,
-            BufferArg::from_raw_parts(handle_in.clone(), vector_size),
-            BufferArg::from_raw_parts(handle_scale_in.clone(), 1),
-            BufferArg::from_raw_parts(handle_scale.clone(), 1),
-            BufferArg::from_raw_parts(handle_q.clone(), vector_size),
-            BufferArg::from_raw_parts(handle_dq.clone(), vector_size),
-        )
-    };
-
-    // 0b0_1000_000 is 2.0 in e4m3.
-    let actual_scale = client.read_one_unchecked(handle_scale);
-    assert_eq!(u8::from_bytes(&actual_scale), &[0b0_1000_000u8]);
-
-    // 1.0, 2.0, 3.0 and 4.0 in e4m3.
-    let expect_q: Vec<u8> = vec![0b0_0111_000, 0b0_1000_000, 0b0_1000_100, 0b0_1001_000];
-    let actual_q = client.read_one_unchecked(handle_q);
-    assert_eq!(u8::from_bytes(&actual_q), &expect_q[..vector_size]);
-
-    let actual_dq = client.read_one_unchecked(handle_dq);
-    assert_eq!(F::from_bytes(&actual_dq), &data[..vector_size]);
-}
-
 #[allow(clippy::unusual_byte_groupings, reason = "Split by float components")]
 pub fn test_fp8<R: Runtime, F: Float + CubeElement>(
     client: ComputeClient<R>,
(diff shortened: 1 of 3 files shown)


ThierryCantin-Demers:
Superseded by #1402, which fixes this bug as part of the Pliron rewrite.

The CUDA cast lowering in `crates/cubecl-cpp/src/cuda/convert.rs` is now a Pliron `LowerOp<Cuda> for CastOp`, and its `intermediate_for_ty` sizes the f16/bf16 intermediate from the **source** type's `vector_size`. That is the same correction this PR made, since the old code sized the intermediate from the destination and so composed a decode intrinsic that does not exist for a scalar source.

Verified on current main (b4907093) on an RTX 4070 Ti SUPER: the repro from this PR passes for f16, bf16, f32 and f64 at vector sizes 1, 2 and 4, with no "Unsupported, skipping".

One loose end worth noting: `test_fp8_broadcast` did not land anywhere, so there is still no regression test for this shape on main. Happy to open a test only PR for it if that is wanted.

PR #1469, opened 2026-08-03
feature(cubecl-cuda): add cuda graph capture and replay

michaelvsinko:
Mirrors **#1415 (`Feat/hip captured graph`)** onto the CUDA backend, plus the memory-node fix that makes a captured graph reliably relaunchable.

## Design

A capture records a fixed-shape launch sequence once and replays it as a single dispatch. The hard constraint is that **the recording window must allocate nothing** — an allocation mid-capture becomes a memory node, and a graph holding one cannot be relaunched. The design is mostly about making that true by construction.

The stream carries a three-state lifecycle, `NoCapture → Prepare → Capture → NoCapture`:

- **`graph_prepare`** arms the persistent pool (exact-fit slices, no bucket padding — a graph's shapes are static) and starts recording which slices are handed out. The caller then runs its workload eagerly a few times as warmup.
- **`start_capture`** closes warmup and opens the window (`cuStreamBeginCapture`, global mode). Launches from here are recorded, not executed.
- **`stop_capture`** ends recording, rejects the graph if it holds memory nodes, instantiates it, and hands back a `CudaGraph` that owns the executable.

Three supporting mechanisms keep the window clean:

* **Pool priming.** Warmup exists to leave the pool able to serve the recorded run without allocating. Retention makes that reliable.
* **Buffer pinning.** `capture_end` returns a retained handle for every slice the window touched, and the graph owns them for its lifetime. The pool can never hand graph memory to a later allocation, which a replay would otherwise corrupt. Slices the window never touched are left alone, so pre-existing live buffers keep their reuse and in-place semantics.
* **Deferred flushes and a pinned info cache.** A fenced flush is a host sync, and a host sync aborts a recording capture — so flushes are deferred while a capture is prepared or recording, and drained at `end_capture`. The metadata info cache runs in capture mode across both `Prepare` and `Capture`, caching every buffer and evicting none, so recorded launches find their info buffers warm and the graph gains no memcpy nodes for them.

Replay is fire-and-forget on the capture stream; inputs are refreshed in place with `client.write` between replays, which is what makes a captured decode step reusable.

## The memory-node fix

A stream-ordered allocation issued while recording is captured as a memory node, and CUDA refuses to relaunch a graph holding an allocation node it never frees: the first `cuGraphLaunch` succeeds and every later one fails with `CUDA_ERROR_INVALID_VALUE`. Nothing downstream noticed — instantiation succeeds and `cuGraphUpload` returns `CUDA_SUCCESS` — so the caller only found out on its second replay, far from the cause. On a 40-layer decode capture this hit ~2-in-3 runs.

The window allocated because warmup recycles its own slices, so the pool only grows to a pass's transient *peak* — and that peak depends on how far the host runs ahead of the server thread, so it can land below what the recorded run asks for. Measured: failing runs opened the window with 44 slices of the 512 B class where the record needed 45; passing runs, 45.

1. **Priming retention** — every slice a warmup pass touches is retained instead of recycled, and released as the window opens. The pool then holds the pass's full distinct *working set*, an upper bound on any peak, so the recorded run reuses rather than allocates.
2. **Reject at `stop_capture`** — a capture that still recorded a memory node is rejected, so a future regression is a caller-visible error rather than a graph that dies on replay.

Priming lives in shared runtime code, so both backends must release it at the same point; that is the 7-line change to `cubecl-hip`.

Verified on a 35B decode capture: 20/20 runs capture and replay with byte-identical output, against ~2-in-3 failing beforehand.

## Tests

`crates/cubecl-cuda/tests/graph.rs` — 5 tests. Three mirror #1415's HIP tests (capture/replay, input rewrite, intermediate recycling); `capture_replay` replays **twice**, which is the exact signature of the bug above. Two are new: a memory-node rejection test that forces pool growth mid-window, and a many-launch dynamic-metadata test.

`crates/cubecl-runtime` — 3 new unit tests for priming retention, which had no coverage. Mutation-checked: removing `primed.push()` fails 3, neutering `capture_priming_end` fails 2, while every pre-existing capture test stays green under both — i.e. the suite previously could not see this regression. 139 unit tests pass.


nathanielsimard:
The first wave of code review is Claude: 

```md
Code review — PR #1469 (CUDA graph capture)

I diffed the CUDA additions against the HIP implementation on main file-by-file (graph.rs, stream.rs, server.rs, command.rs, tests/graph.rs), and built/tested the branch. graph.rs and the four ComputeServer graph methods are essentially a clean transliteration — same state machine, same error strings, same ordering, same comments. The divergences are concentrated in three places.

1. Bug: dead disjunct in the CUDA write path (regression vs. main)

crates/cubecl-cuda/src/compute/command.rs:442

if should_flush && !current.capturing.defers_flushes()
    && (should_flush || current.drop_queue.should_flush())

should_flush && (should_flush || X) ≡ should_flush, so drop_queue.should_flush() is unreachable. main had if should_flush || current.drop_queue.should_flush(). Outside capture, the threshold-driven flush on this path is now gone — the queue only drains via the launch-path site at command.rs:553. HIP has the correct shape:

// cubecl-hip/src/compute/command.rs:388
if (should_flush || current.drop_queue.should_flush()) && !current.capturing.is_recording()

The CUDA line should be the same modulo the gate.

2. defers_flushes() exists only on CUDA — and looks redundant

CUDA adds StreamCaptureState::defers_flushes() (Prepare | Capture) and uses it at all three command.rs flush sites. HIP gates the same three sites on is_recording() (Capture only). So the two backends behave differently during warmup.

The doc justifying the Prepare case argues warmup must hold its staging slices so the pool ends up owning the capture run's full working set. But that guarantee is already provided by the priming retention this PR adds: capture_touch clones every handle into CaptureState::primed while priming is set (memory_manage.rs:907/924/1024), and priming is true for exactly the Prepare window. A drop-queue flush during Prepare returns the handle but the slice stays pinned by primed, so the pool can't recycle it either way. Deferring during Prepare only postpones fences and holds host Bytes longer.

My recommendation: drop defers_flushes() and gate CUDA on is_recording() like HIP. That fixes #1 and restores symmetry in one move. If you'd rather keep it, it needs to be backported to HIP's StreamCaptureState and all three HIP sites — but then the doc comment should credit primed as the actual guarantee, since as written it describes a mechanism that's now redundant.

3. Two capture-safety features are CUDA-only

- Memory-node rejection. count_memory_nodes + the end_capture rejection (server.rs:56-95, ~400) have no HIP counterpart. HIP has hipGraphGetNodes, hipGraphNodeGetType, and hipGraphNodeTypeMemAlloc/MemFree, and the same "graph with an unfreed alloc node can't be relaunched" restriction. This is the case the PR itself argues is undiagnosable without the check — HIP currently hands back a graph that dies on its second replay.
- cuGraphUpload pre-stage. hipGraphUpload exists; either mirror it or note why it's skipped on HIP.

Minor: count_memory_nodes returns 0 when either cuGraphGetNodes call fails, so a query failure reads as "no memory nodes" and the guard silently disarms. Worth a log::warn! at least; the doc comment doesn't mention it.

4. Test coverage asymmetry

cubecl-cuda/tests/graph.rs adds two tests cubecl-hip/tests/graph.rs lacks:

- cuda_graph_capture_growing_the_pool_is_rejected — depends on #3.
- cuda_graph_many_launches_dynamic_metadata — 150 launches through the dynamic-metadata staging + info-cache path, past the 64-push flush threshold. This is backend-agnostic and is the only test that actually exercises the deferred-flush/priming interaction from #2. It should be ported to HIP verbatim.

The CUDA versions of the shared tests also drop the stale gfx1151 / "approach B" s — worth syncing back.
                                                                                                                                                                                            5. Info-buffer caching: same logic, two shapesCUDA factors it into a free info_buffer(command, words) (server.rs:105-138); HIP structure in launch_checked (server.rs:718-744). CUDA also uses ? where HIP uses.unwrap() on create_with_data. CUDA's is the better shape — HIP should adopt the ) key clone is unavoidable on the grid-constants tail path where HIP can moveinfo.data; the grid-constants split itself is a genuine CUDA-only concern, so thefor info_cache is fine.)

6. Example is CUDA-only despite being generic

examples/graph_capture/src/lib.rs is written against <R: Runtime>, but Cargo.tomll/cuda"] and the two examples/*.rs only have #[cfg(feature = "cuda")] arms.examples/throughput/Cargo.toml:20 shows the pattern (hip/rocm). The module doc also opens "Runnable CUDA graph capture and replay examples" — should be backend-neutral.
```

michaelvsinko:
@nathanielsimard thanks for the thorough first pass. All points should now be addressed

examples/graph_capture/examples/basic.rs
```diff
@@ -0,0 +1,4 @@
+fn main() -> Result<(), Box<dyn core::error::Error>> {
```
nathanielsimard: I would remove that example

examples/graph_capture/examples/stateful.rs
```diff
@@ -0,0 +1,4 @@
+fn main() -> Result<(), Box<dyn core::error::Error>> {
```
nathanielsimard: I would remove that example.

examples/graph_capture/src/lib.rs
```diff
@@ -0,0 +1,436 @@
+//! Runnable graph capture and replay examples.
```
nathanielsimard: Not sure this example is also valuable.

examples/graph_capture/examples/basic.rs
michaelvsinko: I removed examples

examples/graph_capture/examples/stateful.rs
michaelvsinko: I removed examples

examples/graph_capture/src/lib.rs
michaelvsinko: I removed examples

nathanielsimard (approved):

PR #1471, opened 2026-08-03
feat: add dp4a intrinsic for packed int8×4 DOT accumulate

saivishwak:
## Summary
- Add `dp4a(a, b, c)` frontend/IR op for packed signed int8×4 DOT + accumulate (Q8 MMVQ).
- CUDA → `__dp4a`; Vulkan/SPIR-V → `OpSDot` + add; other backends → signed-byte polyfill.
- Includes constant folding and GVN support.

## Test plan
- [x] `cargo check` on core/ir/cpp/cpu/opt/spirv/cuda (rustc 1.95)
- [ ] CI
- [ ] Spot-check generated CUDA contains `__dp4a`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3255,7 +3255,7 @@ dependencies = [
 [[package]]
 name = "pliron-spirv"
 version = "0.14.1+sdk-1.4.357.0"
-source = "git+https://github.com/tracel-ai/tracel-rspirv.git?rev=e52c4c918585bc02fa413eb6ee4f0b34e1109ae0#e52c4c918585bc02fa413eb6ee4f0b34e1109ae0"
+source = "git+https://github.com/tracel-ai/tracel-rspirv.git?rev=b3153342f82e504369b982582641b933c45c6950#b3153342f82e504369b982582641b933c45c6950"
 dependencies = [
  "derive-new",
  "derive_more",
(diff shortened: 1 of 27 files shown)


wingertge:
This needs to be rewritten for the new IR. Polyfills now go in `cubecl-core/frontend/polyfills.rs` in `#[cube]` function format. There are plenty of examples you can reference for implementation details in each backend, both polyfills and native ops are used in lots of places.

saivishwak:
> This needs to be rewritten for the new IR. Polyfills now go in `cubecl-core/frontend/polyfills.rs` in `#[cube]` function format. There are plenty of examples you can reference for implementation details in each backend, both polyfills and native ops are used in lots of places.

@wingertge  thanks for the feedback. Will work on that.

saivishwak:
@wingertge , Made the changes. Let me know your thoughts on it.

Thanks

wingertge (approved):
Looks good in its current form, the one thing I'm wondering is whether unsigned dp4a is ever relevant? I'm not super familiar with the fields where it's actually used, so not sure if it's just included in SPIR-V for completeness or whether it's actually useful. Either way, that's something that could be added later.

There's currently a validation error from the missing capability which seems to be caused by the JSON spec not properly listing all cases that require capabilities. I'll fix that upstream with an override, it's not something that can be properly fixed here.

saivishwak:
> Looks good in its current form, the one thing I'm wondering is whether unsigned dp4a is ever relevant? I'm not super familiar with the fields where it's actually used, so not sure if it's just included in SPIR-V for completeness or whether it's actually useful. Either way, that's something that could be added later.
> 
> There's currently a validation error from the missing capability which seems to be caused by the JSON spec not properly listing all cases that require capabilities. I'll fix that upstream with an override, it's not something that can be properly fixed here.

Was using it in Gemma Inference with Quant -> "Gemma Q8 decode i8×i8 GEMV."

PR #1477, opened 2026-08-05
fix(cpp): materialize constant bitcasts as lvalues

### LINKED ISSUE #1476: [🤖] `reinterpret` of a compile-time constant emits invalid C++ (CUDA/HIP/Metal)
_(Analysis and reproduction prepared by Claude; not manually reviewed)._

## Summary

`Operator::Reinterpret` whose input is a constant compiles to

```cpp
const float val_7 = reinterpret_cast<float const&>(uint32(4286578688));
```

A `reinterpret_cast` to a reference type requires a glvalue, but `uint32(4286578688)` is a prvalue, so nvrtc rejects the kernel:

```
default_program(110): error: expression must be an lvalue or xvalue
  const float val_60 = reinterpret_cast<float const&>(uint32(4286578688));
                                                      ^
```

Because the error only appears at kernel-compile time, the CUDA server is left in an invalid state and every later op on that client fails.

## Environment

| | |
|---|---|
| cubecl | `b2703c28b3e395793157f3891e8c0827cae4dbb8` |
| cubek | `a0b5f0e0570fdb7a2dcfb713ab17a825b4ce00b4` |
| burn | `a3da20b774207aadbf5fe5789df0368e2a8d4453` |
| GPU / driver | RTX 2060 |
| CUDA | 12.x (nvrtc) |
| rustc | 1.97.1 |

## Why it fires now

`cubek-reduce` builds the max/min identities from IEEE bits (`crates/cubek-reduce/src/components/instructions/extrema.rs:23-41`):

```rust
#[cube]
pub(crate) fn max_identity<E: Numeric>() -> E {
    let elem_type = type_of::<E>();
    if comptime!(elem_type.is_float()) {
        // WGSL has no infinity literal, so construct it from its IEEE-754 bits.
        E::cast_from(f32::reinterpret(0xff80_0000u32))
    } else {
        E::min_value()
    }
}
```

`0xff80_0000u32` stays a constant (`try_const_eval_operator` explicitly does not fold `Operator::Reinterpret`, `cubecl-core/src/post_processing/constant_prop.rs:572`), so **every float `max`/`min` reduce fails to compile on the C++ dialects**.

Downstream this means e.g. `burn::Tensor::max()` panics on the CUDA backend.

## Provenance

The codegen hole is old; only the caller is new.

- **Latent since cubecl `bda1760a` ("Fix bitcast", 2025-01-05)** — the commit that introduced the reference form `reinterpret_cast<T&>({input})` for `Bitcast`. Reshaped later by `35b861d0` (2026-06-12) but never taught about constant operands. `Value::ensure_lvalue` arrived with `cb87b0d2` (2026-05-15) and was wired into the atomic / indexing paths only.
- **Triggered by cubek `09e10a52` (https://github.com/tracel-ai/cubek/pull/457, 2026-08-04)** — it added `extrema.rs`, replacing the previous finite identity `P::EA::min_value()` in `max.rs`/`min.rs` with the ±inf bit-reinterpret above. First cubek pin carrying it is `a0b5f0e0` (https://github.com/tracel-ai/cubek/pull/458), which burn   picked up in `0f556ad1a` (https://github.com/tracel-ai/burn/pull/5299, 2026-08-04).

The cubecl bump in the same burn update is *not* the cause: the `Bitcast` arm is byte-identical at the previous pin `05cac67370653ef6ed6b76730b20a0cb4bede117`.

## Reproducer

`crates/cubecl-cuda/tests/reinterpret_const.rs`:

```rust
use cubecl_core as cubecl;
use cubecl_core::prelude::*;
use cubecl_cuda::CudaRuntime;

#[cube(launch)]
fn neg_inf_from_bits(output: &mut [f32]) {
    if ABSOLUTE_POS < output.len() {
        output[ABSOLUTE_POS] = f32::reinterpret(0xff80_0000u32);
    }
}

#[test]
fn reinterpret_of_a_constant() {
    let client = CudaRuntime::client(&Default::default());
    let out = client.empty(4 * size_of::<f32>());

    unsafe {
        neg_inf_from_bits::launch::<CudaRuntime>(
            &client,
            CubeCount::new_single(),
            CubeDim::new_1d(4),
            BufferArg::from_raw_parts(out.clone(), 4),
        )
    };

    let bytes = client.read_one_unchecked(out);
    assert_eq!(f32::from_bytes(&bytes), &[f32::NEG_INFINITY; 4]);
}
```

## Cause

`Instruction::Bitcast` in `crates/cubecl-cpp/src/shared/instruction.rs:595` formats the input straight into the `reinterpret_cast`, unlike the other reference-taking instructions, which call `Value::ensure_lvalue` — the helper whose doc-comment is literally *"Ensure a variable is a named lvalue, reassigning to a temporary if necessary. This is required for reinterpreting constants."* (`crates/cubecl-cpp/src/shared/value.rs:463-477`).

Shared code, so CUDA, HIP and Metal are all affected. WGSL (`bitcast<f32>(literal)`) and SPIR-V are fine.

## Fix

```diff
--- a/crates/cubecl-cpp/src/shared/instruction.rs
+++ b/crates/cubecl-cpp/src/shared/instruction.rs
@@
                 if out_item.size() != input_item.size() {
                     panic!("Unsupported type for bitcasting {out_item:?} from {input_item:?}");
                 } else {
-                    let out = out.fmt_left();
                     let addr_space = D::address_space_for_value(input);
+                    // A reference can only bind to a glvalue, so a constant operand has to be
+                    // materialized into a named temporary first.
+                    let input = input.ensure_lvalue(f)?;
+                    let out = out.fmt_left();
                     writeln!(
                         f,
                         "{out} = reinterpret_cast<{addr_space}{out_item}{qualifier}&>({input});"
                     )
                 }
```

Const-folding `Operator::Reinterpret` in `constant_prop` would also remove this particular occurrence, but the codegen fix is the one that covers all constant inputs.

Verified: with only this diff applied, all 152 tests of a downstream Burn crate (`burn-mamba`, heavy user of `Tensor::max()`) pass on the CUDA backend; 125 of them failed without it. The reproducer above fails before and passes after.

Related spot, same prvalue hazard: `shared/warp.rs:379` (`fn cast`) formats `reinterpret_cast<T&>({input})` too, and — being a `String`-returning helper with no `Formatter` — cannot materialize a temporary. A plane op with a constant operand of a different item type would hit the same nvrtc error.

## Side note (warning, not an error)

The same reduce kernel emits

```cpp
const bool_4 val_73 = bool_4{isnan(val_74.i_0), isnan(val_74.i_1), ...};
```

which nvrtc reports as `warning #2361-D: invalid narrowing conversion from "int" to "__nv_bool"` — unqualified `isnan` resolves to the C macro/function returning `int`. Harmless, but noisy on every max/min reduce.

jcwal1516:
## Summary

Materialize constant operands before emitting the reference-form `reinterpret_cast` used by C++ dialect bitcasts.

`Value::Constant` formats as a constructor expression, which is a prvalue. A reference-form `reinterpret_cast` requires a glvalue, so the generated CUDA, HIP, and Metal source is invalid when the bitcast input is a constant.

This code-generation limitation predates the current reduction work. tracel-ai/cubek#457 exposed it by constructing floating-point extrema identities through constant IEEE-754 bit reinterpretation.

Fixes #1476.
Related to tracel-ai/cubek#457.

## Changes

- Materialize `Value::Constant` bitcast inputs through the existing `Value::ensure_lvalue` path.
- Preserve the existing lowering for nonconstant and pointer inputs.
- Add a focused Metal source-generation regression test verifying that the cast consumes a named temporary.

This does not change the public API or add dependencies.

## Testing

- `cargo fmt --all -- --check` — passed.
- `cargo test -p cubecl-cpp --features metal --locked --lib shared::instruction::tests::bitcast_constant_uses_lvalue -- --exact` — 1 passed.
- `cargo clippy -p cubecl-cpp --locked --all-targets --features metal -- -D warnings` — passed.
- `cargo check -p cubecl-cpp --locked --no-default-features --features std,metal` — passed.
- `cargo check -p cubecl-cpp --locked --no-default-features --features std,cuda` — passed.
- `cargo check -p cubecl-cpp --locked --no-default-features --features std,hip` — passed.
- Cubek's `nan_extrema` suite using this CubeCL commit on an NVIDIA RTX 4070 SUPER with CUDA 13.2 — 10 passed, 0 failed.
- Cubek's `cargo xtask test --ci` — 8,413 passed and 24 pre-existing TopK-with-indices cases failed; an exact failure was reproduced on untouched Cubek `origin/main` at `de661419` with CubeCL `5d5af999`.

Cubek validation PR: tracel-ai/cubek#461.

Burn validation remains pending while this PR and the Cubek pin are in draft.

### Instructions

- [x] Submit a PR in cubek with your fixes and link it here.

### DIFF
--- a/crates/cubecl-cpp/src/shared/instruction.rs
+++ b/crates/cubecl-cpp/src/shared/instruction.rs
@@ -602,6 +602,10 @@ for (*{i} = {start}; *{i} {cmp} {end}; {increment}) {{
                 } else {
                     let out = out.fmt_left();
                     let addr_space = D::address_space_for_value(input);
+                    let input = match input {
+                        Value::Constant(..) => input.ensure_lvalue(f)?,
+                        _ => *input,
+                    };
                     writeln!(
                         f,
                         "{out} = reinterpret_cast<{addr_space}{out_item}{qualifier}&>({input});"
@@ -1039,5 +1043,33 @@ impl<V: Display, D: Dialect> Display for EnsureBoolArg<'_, V, D> {
         } else {
             write!(f, "{}", self.val)
         }
+    }
+}
+
+#[cfg(all(test, feature = "metal"))]
+mod tests {
+    use cubecl_core::ir::ConstantValue;
+
+    use crate::metal::MslDialect;
+
+    use super::*;
+
+    #[test]
+    fn bitcast_constant_uses_lvalue() {
+        let instruction = Instruction::<MslDialect>::Bitcast(UnaryInstruction {
+            input: Value::Constant(ConstantValue::UInt(0x7f80_0000), Item::Scalar(Elem::U32)),
+            out: Value::Value {
+                id: 0,
+                item: Item::Scalar(Elem::F32),
+            },
+        });
+
+        let source = instruction.to_string();
+
+        assert!(source.contains("uint _tmp_"), "{source}");
+        assert!(
+            source.contains("reinterpret_cast<thread float const&>(_tmp_"),
+            "{source}"
+        );
     }
 }


laggui (approved):
Validated locally (CUDA)

PR #1484, opened 2026-08-07
fix(spirv): name the FPEncoding variants added in tracel-rspirv 0.13.3

ThierryCantin-Demers:
`tracel-rspirv 0.13.3` adds five variants to `FPEncoding`, so the `Display` impl for `Elem` stopped being exhaustive and `cubecl-spirv` no longer builds:

```
error[E0004]: non-exhaustive patterns
  --> crates/cubecl-spirv/src/item.rs:622:15
```

`Cargo.lock` is not tracked, so CI resolves the newest `0.13.x` on every run and hits this. It fails on `main` too, independently of any PR.

The requirement moves to `0.13.3` rather than taking a wildcard arm: `0.13.1` has only the three encodings already matched, where a wildcard would be an unreachable pattern. Keeping the match exhaustive means the next upstream addition fails at compile time instead of silently printing a wrong debug name.

The names follow the existing `FloatKind` vocabulary. `MXInt8EXT` has no cubecl equivalent, so it gets `mxint8`. All five arms are unreachable today, since `compile_elem` panics with "Minifloat not supported in SPIR-V" for those kinds; this only affects debug symbol names.

Verified with `cargo check -p cubecl-wgpu --features spirv`, the exact check that fails ("std with SPIR-V compiler", `xtask/src/commands/check.rs:56`), and `cargo clippy -p cubecl-spirv -p cubecl-wgpu --features spirv --all-targets -- -D warnings`.

### Instructions

Not run: this adds match arms to a debug-name `Display` impl and raises a patch-level floor on an existing dependency. No API surface changes, so there is nothing downstream to fix.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -160,7 +160,7 @@ cudarc = { version = "0.19.6", features = [
 ], default-features = false } # CubeCL-CUDA
 
 # CubeCL-SPIR-V
-rspirv = { package = "tracel-rspirv", version = "0.13.0" }
+rspirv = { package = "tracel-rspirv", version = "0.13.3" }
 
 # CubeCL-WGPU
 ash = "0.38"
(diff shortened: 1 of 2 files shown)


louisfd (approved):

wingertge:
That was my bad, I yanked the breaking patch releases and republished as a major. I'll figure out a way to enforce semver checks on CI going forward.

PR #1487, opened 2026-08-07
fix(opt): keep the switch merge on the block its cases converge to

ThierryCantin-Demers:
A switch whose default or any case diverged took the function return block as its merge, while the remaining cases still branched to the block after the switch. That block is the real convergence point, so the module named one merge and branched to another:

```
%23 = OpLabel
      OpSelectionMerge %39 None          <- declares %39
      OpSwitch %32 %35 1 %36 0 %37 2 %38
%36 =   OpBranch %40                     <- every case goes to %40
%37 =   OpBranch %40
%38 =   OpBranch %40
%35 =   OpUnreachable                    <- default diverges, which triggered the bad merge
%40 = OpLabel                            <- the actual convergence point
```

`spirv-val` rejects this, Mesa then fails `spirv_to_nir` with `VK_ERROR_UNKNOWN`, and wgpu treats the validation error as fatal, so a single bad shader took down **517 of 658 tests** on Intel. NVIDIA accepts the invalid module, which is why CI and NVIDIA developers never saw it.

`is_ret` was computed with `.any()`, so one diverging path was enough to redirect the merge. It now tracks fall-through instead: `next` is the merge whenever anything reaches it, and the return block is used only when nothing does. A case that returns ends in its own terminator and needs no branch to the merge, so mixing the two is fine.

## Verification

Every run below is against a baseline on the same hardware.

| Backend / GPU | main | this PR |
|---|---|---|
| SPIR-V, Arc B390 (validation layers on) | 122 passed / 517 failed | **621 passed / 18 failed** |
| SPIR-V, RTX 4070 Ti SUPER | 639 / 0 | 639 / 0 |
| CUDA, RTX 4070 Ti SUPER | 774 / 0 | 774 / 0 |
| WGSL, Arc B390 | 12/12 on the shuffle subset | 417 / 12 |

Rather than trust one driver, I dumped every shader with `CUBECL_DEBUG_SPIRV` and ran `spirv-val` over all of them: **1 invalid of 476 before, 0 invalid of 685 after**. The offender was `KernelRuntimeVariantsValue`, from `tests_spirv::enums::runtime_enum_value`.

The remaining 18 (SPIR-V) and 12 (WGSL) failures are `plane_shuffle_up`/`plane_shuffle_down` numeric mismatches at the boundary lane. They reproduce with a byte-identical failing set on unmodified main, so they are pre-existing and unrelated.

Two notes for reviewers:

- Under the default parallel test runner the suite now reaches a SIGSEGV in the Intel driver. Single-threaded it completes cleanly. That is a Mesa concurrency issue that was simply unreachable before, since the suite used to die in the first two seconds.
- There is a latent sibling defect this PR does not touch: when a case contains a `break`, `merge` becomes `None` and `compile_switch` emits `OpSwitch` with no preceding `OpSelectionMerge`, which is also invalid. No test covers that path, so I had no way to verify a fix for it.

### Instructions

Not run. This changes no API, only which block a switch names as its merge, and it strictly increases the set of shaders that validate. The CUDA and NVIDIA Vulkan suites above are unchanged from main, which covers the backends a downstream build would exercise.

### DIFF
--- a/crates/cubecl-opt/src/control_flow.rs
+++ b/crates/cubecl-opt/src/control_flow.rs
@@ -194,23 +194,23 @@ impl Function {
                 self.add_edge(current_block, case_id, 0);
                 self.current_block = Some(case_id);
                 let is_break = self.parse_scope(state, case);
-                let is_ret = if let Some(current_block) = self.current_block {
+                let falls_through = if let Some(current_block) = self.current_block {
                     self.add_edge(current_block, next, 0);
-                    false
+                    true
                 } else {
-                    !is_break
+                    false
                 };
                 let val = match val.as_const().expect("Switch value must be constant") {
                     ConstantValue::Int(val) => unsafe { transmute::<i32, u32>(val as i32) },
                     ConstantValue::UInt(val) => val as u32,
                     _ => unreachable!("Switch cases must be integer"),
                 };
-                (val, case_id, is_break, is_ret)
+                (val, case_id, is_break, falls_through)
             })
             .collect::<Vec<_>>();
 
         let is_break_branch = branches.iter().any(|it| it.2);
-        let mut is_ret = branches.iter().any(|it| it.3);
+        let mut falls_through = branches.iter().any(|it| it.3);
         let branches = branches
             .into_iter()
             .map(|it| (it.0, it.1))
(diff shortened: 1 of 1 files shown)


ThierryCantin-Demers:
Superseded by #1402, which fixes this bug as part of the Pliron rewrite.

Both files this PR touched are gone: `crates/cubecl-opt/src/control_flow.rs` and `crates/cubecl-spirv/src/branch.rs` were deleted. Switch lowering now lives in `crates/cubecl-spirv/src/ops/branch.rs`, which **constructs** an explicit merge block and rewrites every `YieldOp` case terminator into a `BranchOp` to that block, while leaving diverging terminators alone (`IsExitTerminator`, kept because they are already valid in SPIR-V).

That is the same rule this PR was computing by hand, except it is now structural: the merge block is built rather than inferred from a flat CFG, so a case can no longer branch somewhere other than the declared merge.

Verified on an Intel Arc B390 (Mesa 26.1.6) with Vulkan validation layers enabled: `Case construct that targets ... has invalid branch to block ...` no longer appears, and there are zero SPIR-V validation errors across the suite. Before, that single invalid shader was fatal and took down 517 of 658 tests.

Unrelated to this PR, but found while confirming the above: the SPIR-V backend has a separate regression on current main. The suite reports 358 failures on an RTX 4070 Ti SUPER and 345 on the Arc, as wrong values rather than compile errors, while WGSL (422 passed, 12 pre-existing plane_shuffle failures) and CUDA (774 passed, 0 failed) are healthy. Five of the six enum tests fail this way, for example `runtime_enum_value` returns 0.0 instead of 5.0. I can open a separate issue with the details.

PR #1496, opened 2026-08-11
Reorder passes to avoid `CondBrOp` panic

wingertge:
`SimplifyCFGPass` panics if the branch cond is constant and the attribute is not an `IntegerAttr`, but the pass ordering meant that constant conditions were still `BoolAttr` attributes. This moves the SimplifyCFG and DCE passes after the conversion to avoid this panic. Fixes the CI failures in cubek.


marcantoinem (approved):
LGTM

PR #1500, opened 2026-08-11
Clean up dead regions before lowering branches to avoid weird crash in DCE

wingertge:
This may be an upstream bug but can be worked around for now. Also fixes dynamic extract for CUDA


marcantoinem (approved):
I don't know why that is causing this, but it doesn't hurt to do a bit more dce

wingertge:
> I don't know why that is causing this, but it doesn't hurt to do a bit more dce

It seems like DCE isn't properly deleting the region's children when it's erased, that's why I'm thinking it's an upstream bug. But this also works, and `branch.dead_region` is mostly a hack to deal with the lack of a `RegionBranchOpInterface` and should eventually be repalced.

PR #1502, opened 2026-08-12
Add SROA pass

wingertge:
Adds a scalar replacement of aggregates (SROA) pass that works on both memory and SSA composites, and supercedes the previous `DisaggregatePass`, `DisaggregateArray` and an old pass that merged successive `insert`s into a single `construct`. It also adds new functionality particularly for C++, where unrolling left behind a lot of pointless vectors that were only used to extract from.

`Composite` and `Vector` ops with constant indices are merged into a single op, with asserts to ensure no non-vector composites remain in the final codegen (since they don't have a backend representation).

Also adds a new internal `NamedRewrite` derive and trait to ensure IR dumps use useful names instead of just being called `MatchRewrite` or `DialectConversion`. I needed it for debugging so it's in here.

Validated in `burn` and using a new test


marcantoinem (approved):
LGTM, really nice that we can name match rewrite pass with that now

PR #1506, opened 2026-08-12
Lower cube.poison to OpUndef in the SPIR-V backend

nathanielsimard:
The SROA pass can scalarize a local array into a plain variable, which mem2reg then promotes with a cube.poison initial value threaded through loops as an iter-arg. cube.poison had no SPIR-V lowering, so emission failed on any kernel where this happened (e.g. the plane vec-mat matmul), and the failure was an .expect() panic on the device thread: the launch silently produced uninitialized output instead of a compile error. Lower poison to spirv.Undef and propagate to_spirv failures as a CompilationError.

### DIFF
--- a/crates/cubecl-runtime/src/memory_management/memory_manage.rs
+++ b/crates/cubecl-runtime/src/memory_management/memory_manage.rs
@@ -633,14 +633,22 @@ fn build_pools(
                 });
             }
 
-            // Add pools from big to small.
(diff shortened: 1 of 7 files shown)


wingertge:
It should be removed by DCE, if poison is left in the binary then something is reading uninitialized memory. I'll look into it

louisfd:
I think it's also a problem on metal

PR #1507, opened 2026-08-12
Fix MSL codegen

louisfd:


wingertge:
Metal doesn't forbid lambdas on my Mac, which version are you on? They compile just fine for me and are in fact used for working around a Metal compiler bug. For the record, it works on this version (`xcrun metal --version`):
```
Apple metal version 32023.864 (metalfe-32023.864)
Target: air64-apple-darwin25.3.0
Thread model: posix
InstalledDir: /private/var/run/com.apple.security.cryptexd/mnt/com.apple.MobileAsset.MetalToolchain-v17.3.7003.10.2cafHX/Metal.xctoolchain/usr/metal/current/bin
```

crates/cubecl-cpp/src/metal/vector.rs
```diff
@@ -30,38 +30,36 @@ metal_op_with_out!(FDotOp, |op, ctx| {
...
     let index = op.index(ctx).0;
+    let out = op.get_result(ctx).name(ctx);
     let vector_ty = op.composite(ctx).get_type(ctx).to_cpp(ctx);
-    scoped_block!(
-        format!("{vector_ty} tmp = {vector};")
-        format!("tmp.i_{index} = {value};")
-        "return tmp;"
-    )
+    format!("{vector_ty} {out} = {vector};\n{out}.i_{index} = {value};\n")
 });
```
wingertge, line 47: This is likely to cause problems, because the codegen assumes all non-pointer locals are `const&` and this creates a non-const local. That's why the lambda was there in the first place, to isolate the mutable part in a subscope and only return the const value.

crates/cubecl-cpp/src/metal/vector.rs
```diff
@@ -30,38 +30,36 @@ metal_op_with_out!(FDotOp, |op, ctx| {
...
+    let out = op.get_result(ctx).name(ctx);
     let vector_ty = op.vector(ctx).get_type(ctx).to_cpp(ctx);
-    scoped_block!(
-        format!("{vector_ty} tmp = {vector};")
-        format!("*(reinterpret_cast<thread {elem_ty}*>(&tmp) + {index}) = {value};")
-        "return tmp;"
+    format!(
+        "{vector_ty} {out} = {vector};\n\
+         *(reinterpret_cast<thread {elem_ty}*>(&{out}) + {index}) = {value};\n"
     )
```
wingertge: Same as above, having non-const non-pointer locals breaks some assumptions I made in other parts of the backend.

wingertge (commented):
I have concerns about the way scoped blocks are removed

wingertge (approved):

PR #1508, opened 2026-08-12
Fix/pliron codegen regressions

nathanielsimard:


crates/cubecl-ir/src/dialect/general.rs
```diff
@@ -117,6 +117,11 @@ const_eval!(BoolNotOp, {
 
 #[cube_op(name = "cube.cast")]
 #[result_ty(argument)]
+// Element-wise, so a wide vector cast splits into narrower casts lane-group by
+// lane-group — without this the unroll pass (used when a kernel's vector width
+// exceeds the target's, e.g. 8-wide q4 lanes against SPIR-V's max of 4) fails
+// on any vector cast.
+#[op_interfaces(TriviallyUnrollable)]
```
wingertge: The comment is only relevant for why it's a fix here, and shouldn't stay there permanently.

crates/cubecl-ir/src/dialect/math.rs
```diff
@@ -680,7 +680,9 @@ pub(super) fn float_attr(ctx: &Context, ty: TypeHandle, val: f64) -> AttrObj {
 
 #[cube_op(name = "math.fma")]
 #[result_ty(same_as = a)]
-#[op_interfaces(SameOperandsType, SameOperandsAndResultType)]
+// Element-wise, so the unroll pass can split a wide vector fma into narrower
+// ones — the same latent failure CastOp had before it was marked.
```
wingertge: Same as before with the comment

crates/cubecl-spirv/src/ops/general.rs
```diff
@@ -52,6 +52,24 @@ impl ToSpirvDialectOp for general::SelectOp {
...
+        let op = self.get_operation();
+        let out_ty = ty_to_spirv_dialect(ctx, self.get_result(ctx).get_type(ctx));
+        let new_op = ops::UndefOp::new(ctx, out_ty);
+        rewriter.append_op(ctx, &new_op);
+        rewriter.replace_operation(ctx, op, new_op.get_operation());
+
+        Ok(())
+    }
+}
+
```
wingertge, line 72: `PoisonOp` indicates an invalid value and should not be compiled into a binary. I'm working on a proper fix for the unitialized arrays.

wingertge (changes requested):

crates/cubecl-cpp/src/shared/operation.rs
```diff
@@ -197,8 +201,25 @@ shared_op_with_out!(FmaOp, |op, ctx| {
...
+    // its scalar and vector float types, including half.
+    let f = if matches!(ctx.target(), Target::Metal) {
+        "fma"
+    } else if res.is_half(ctx) {
+        "__hfma"
+    } else if res.is_half2(ctx) {
+        "__hfma2"
+    } else {
+        "fma"
+    };
```
wingertge, line 218: We should probably also add an FMA test, it's not called anywhere right now so doesn't really get validated properly when testing

crates/cubecl-spirv/src/validate.rs
```diff
@@ -0,0 +1,127 @@
...
+    dr::{Instruction, Module, Operand},
+    spirv::{CooperativeMatrixUse, FPEncoding, Op, Scope},
+};
+
+/// Checks that every subgroup-scope cooperative matrix type declared in the
+/// module matches a fragment shape the device advertises. Drivers only accept
+/// the exact (element type, rows, columns, use) combinations they report;
+/// anything else is undefined behavior at runtime — loads and stores silently
+/// produce garbage on at least RADV — so it must be rejected at compile time.
+pub fn check_cmma_types(module: &Module, props: &DeviceProperties) -> Result<(), String> {
```
wingertge: Should this check not be applied to all targets? C++ will fail to compile, but that's very poorly readable compared to a real validation error. The shapes and types are inherited from the high-level `cube` instructions so the SPIR-V specific behavior seems pointless.

crates/cubecl-spirv/src/ops/general.rs
wingertge: This should be removed, the issue is properly fixed by https://github.com/tracel-ai/cubecl/pull/1511

wingertge: Github was bugged, nevermind

wingertge (approved):

PR #1511, opened 2026-08-13
Zero-init arrays

wingertge:
Rust semantics require all locals to be initialized in all divergent paths, but `Array::new` currently returns an uninitialized local which can cause undefined values to be read from it. This adds a new `ZeroAttr` for easily defining type-agnostic zeroed values, and uses it to initialize the two kinds of locals that are not enforced by Rust. `let x;` is *not* initialized because Rust already ensures it's assigned in every path. We may want to change `Array::new` to `Array::zeroed` at some point and add a new way to construct it with a value, but the fix comes first.

`ZeroAttr` maps to `ZeroOp` for CPU, `{}` for C++, `NullAttr`/`OpConstantNull` for SPIR-V and `{type}()` for WGSL. These all have the same behavior of zeroing the whole array in a single constant instruction.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1244,9 +1244,9 @@ dependencies = [
 
 [[package]]
 name = "cudarc"
-version = "0.19.8"
+version = "0.19.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "42310153e06cf4cd532901f7096beb27504d681736a29ee90728ae4e2d93b2a8"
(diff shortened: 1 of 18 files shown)


marcantoinem (approved):
Lgtm

marcantoinem (approved):
Lgtm

marcantoinem (approved):
Lgtm

marcantoinem (approved):
Lgtm

PR #1514, opened 2026-08-13
Replace QuantLevel with per-tensor and per-block scale levels

ThierryCantin-Demers:
Part of the two-level quantization chain. These merge in order, top to bottom, after this PR:

- [cubek#498](https://github.com/tracel-ai/cubek/pull/498) — the tile path reads both scale levels, folding the per-tensor scale into a register once per kernel.
- [cubek#444](https://github.com/tracel-ai/cubek/pull/444) — the elementwise quantize/dequantize kernels take their scale bindings as a slice, plus the repo-wide migration to the scheme API below. Stacked on cubek#498.
- [burn#5262](https://github.com/tracel-ai/burn/pull/5262) — `QParams` carries the per-tensor scale, the quantize and dequantize call sites pass both bindings, and the backends validate levels.

Each repo pins the one above it, so the pins move together: bumping one alone leaves two cubecl revisions in the dependency graph. This means that the PRs in the other repos will require changing the pinned revs as the PRs get merged.

---

Quantization schemes now describe their scales as up to two optional levels instead of the closed `QuantLevel` enum: a per-block level (`BlockScale`, a `BlockSize` plus the `ScaleDtype` its scales are stored in) and a per-tensor level (a `ScaleDtype`). One level alone is single-level quantization; both together are a two-level scheme whose block scales are normalized by the per-tensor scale. A scheme storing neither resolves to a per-tensor f32 scale in the accessor, not in storage, so the default stays additive: setting one level never drags a phantom second one along.

Before, the shape downstream builds against (the two-level `BlockTensor` variant from #1479 never made it out of cubecl):

```rust
// One scale for the whole tensor.
QuantScheme::default()
    .with_param(QuantParam::F32)
    .with_level(QuantLevel::Tensor);

// One scale per block of 32 values.
QuantScheme::default()
    .with_param(QuantParam::F32)
    .with_level(QuantLevel::Block(BlockSize::new([32])));

// Two levels, as #1479 shipped it: the block scales read their precision from
// the scheme-wide `param`, the per-tensor scale from a second param inside the
// variant.
QuantScheme::default()
    .with_param(QuantParam::UE4M3)
    .with_level(QuantLevel::block_tensor([32], QuantParam::F32));
```

After:

```rust
// One scale for the whole tensor, stored as f32. Also what a scheme with no
// level set resolves to.
QuantScheme::default().per_tensor(ScaleDtype::F32);

// One scale per block of 32 values.
QuantScheme::default().per_block([32], ScaleDtype::F32);

// Two levels: ue4m3 block scales, normalized by a single per-tensor f32 scale
// that absorbs the tensor's dynamic range.
QuantScheme::default()
    .per_block([16], ScaleDtype::UE4M3)
    .per_tensor(ScaleDtype::F32);

// Named formats live in a presets module and expand to the setters above.
// WARNING: nvfp4 is defined as a scheme but probably doesnt actually work as of right now 
// as support for fp8 types is uncertain.
QuantScheme::nvfp4();
```

Each level carries its own scale dtype, and the two setters are additive and order-free. Every combination of the two fields is a valid scheme, so construction cannot fail or panic and there is no validation machinery to maintain.

## What changed

- `cubecl-common`: `QuantLevel` (`Tensor` / `Block` / `BlockTensor`) is replaced by two private optional fields on `QuantScheme`, set with `per_tensor(dtype)` / `per_block(block, dtype)` and read through `tensor_scale()`, `block_scale()`, `scale_dtype()` (innermost), `block_size()`, and `num_levels()`. The fields stay private because they gate interpretation rather than validity: a scheme storing no level resolves to per-tensor f32 in `tensor_scale()`, and raw field reads would let a consumer mistake the default for "no scale". `QuantParam` is renamed to `ScaleDtype`, and the `param` vocabulary follows to `dtype`. Whole-tensor granularity is the per-tensor level rather than a sentinel block size, which leaves `BlockSize` purely rank-relative: it gains `FULL` dimensions covering one axis entirely, a `grid` for scale-tensor shapes, and a canonical form so equal blocks compare and hash equal however they are spelled. An all-`FULL` block is consequently not whole-tensor granularity; on a rank-3 tensor `[FULL, FULL]` is one scale per batch element. Named formats live in `quant/presets.rs`, starting with `nvfp4()`.
- `cubecl-std`: the launch surface takes `ScaleBindings` (`one` / `two`), one scale binding per level, through a single `ViewArg::new_quantized` constructor. The per-tensor scale of a two-level scheme binds as a one-element f32 buffer read once per kernel into a register instead of being re-read per element. The cube-side `QuantizedView` records which part of each read's scale the caller already knew in `KnownScale` (`None` / `Outer` / `Whole`), a comptime discriminant so each case compiles its own kernel, and gains `new_with_outer_scale` (per-position block scales, per-tensor scale already in a register) and `new_with_whole_scale` (the register is the entire scale, the scales view is never read), for callers like a matmul tile that preload scales themselves.

## How this differs from the two-level plan

Two-level quantization was already implemented in cubecl by #1479, as a dedicated enum variant, `QuantLevel::BlockTensor { block, global }`, with the inner param living in the separate `QuantScheme::param` field. Working with that result is what led to this PR: the per-tensor scale entered the code as a "global" exception rather than a rule, one extra field on one variant, a second param living outside the level, and a special case at every consumer that had to remember to ask whether a global factor also applied. Naming both levels as first-class optional fields dissolves the exception: the per-tensor scale is simply the outer level, present or not, with its own dtype.

The `BlockTensor` variant stayed internal to cubecl, cubek and burn still build single-level schemes, so replacing it now means downstream migrates once, straight from the single-level enum to the two-field scheme. Concretely, the old shape had two problems this PR removes:

- The scale precision was split across two places, and a consumer reading `scheme.param` alone silently missed the per-tensor factor (the field doc on main warns about exactly this). Now every level carries its own dtype and there is no second place to forget.
- Each level combination was a new enum variant. The two fields represent exactly the combinations consumers can execute, since only the innermost level's scales are addressed per position; anything deeper has no consumer and would come with its own kernel work anyway.

The launch story also changes: the outer scale was applied through a third view read per element, it is now read once per kernel and folded in a register, which the optimizer could not do on its own.

## Public API changes

Breaking, in `cubecl-common`:

- `QuantScheme` loses its `pub param: QuantParam` and `pub level: QuantLevel` fields. Scales are set additively with `per_tensor(dtype)` and `per_block(block, dtype)` and read through `tensor_scale()`, `block_scale()`, `scale_dtype()`, `block_size()`, and `num_levels()`.
- `QuantLevel` is gone. The old spellings map one to one: `Tensor` becomes `.per_tensor(dtype)`, `Block(b)` becomes `.per_block(b, dtype)`, `BlockTensor { block, global }` becomes `.per_block(block, dtype).per_tensor(global)`.
- `QuantParam` is renamed to `ScaleDtype`.
- `BlockSize` loses `full()` and `is_full()`: whole-tensor granularity is the per-tensor level, not a block size.

Breaking, in `cubecl-std`:

- `ViewArg::new_quantized` / `new_quantized_two_level` collapse into `new_quantized(values, ScaleBindings, scheme)`.

## Serialization is breaking on purpose

Stored schemes from before this PR do not load: the legacy deserialization shim was removed after review so the scheme is free to keep the cleanest shape, and every serde impl in the module is now a plain derive. Users with stored quantized artifacts requantize and reserialize them.

## Mitigations

- **Misuse fails loudly instead of silently.** The scale binding count is checked against the scheme's levels with a message naming both, and a two-level scheme's per-tensor scale must bind as f32. `BlockSize::num_elements` asserts in release rather than multiplying a `FULL` dimension as zero, since a full dimension has no element count without a shape.
- **Every scale shape is GPU-tested.** The quantized view tests cover per-tensor, per-block, and two-level schemes through every read method, plus both cube-side constructors, on wgpu. The CUDA suite passed at an earlier revision of this branch (780 passed, 3 failures reproduced on main); it should be re-run before merge.

### Instructions


laggui (commented):
This is a nice generalization, I like `ScaleLevels`. The representation is extensible for the future, even if not currently a general hierarchy. It might be best described as a uniform representation for one-level and NVFP4-style two-level quantization.

But I think tensor-vs-block should remain explicit through something like `ScaleGranularity`:

```rust
enum ScaleGranularity {
    Tensor,
    Block(BlockSize),
}

struct ScaleLevel {
    granularity: ScaleGranularity,
    param: QuantParam,
}
```

Using `BlockSize::full()` overloads `[FULL]` to mean both the whole tensor and the full trailing dimension. An explicit `Tensor` variant removes this ambiguity.

crates/cubecl-common/src/quant/scheme.rs
```diff
@@ -4,44 +4,36 @@ use core::{default::Default, ops::Deref};
...
-    pub param: QuantParam,
     /// Data type used for storing quantized values.
     pub store: QuantStore,
-    /// Granularity level of quantization (e.g., per-tensor).
-    pub level: QuantLevel,
     /// Quantization mode (e.g., symmetric).
     pub mode: QuantMode,
+    /// The scale levels, innermost first. Private so that ordering and nesting hold by
+    /// construction: go through [`ScaleLevels`] to build them.
+    levels: ScaleLevels,
```
nathanielsimard: ScaleLevels => ScaleSequence? It also has the scales in the order they should be calibrated/applied during dequantization.

crates/cubecl-common/src/quant/scheme.rs
```diff
@@ -107,76 +149,483 @@ impl QuantScheme {
...
+    }
+}
+
+/// One level of scales: values inside each region share one scale, stored as `param`.
+#[derive(Clone, Copy, Debug, Hash, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
+pub struct ScaleLevel {
+    /// The region covered by one scale.
+    pub granularity: ScaleGranularity,
+    /// The precision the level's scales are stored in.
+    pub param: QuantParam,
```
nathanielsimard: QuantParam => ScaleDtype?

crates/cubecl-common/src/quant/scheme.rs
```diff
@@ -107,76 +149,483 @@ impl QuantScheme {
...
+/// A two-level scheme exists so block scales can live in a narrow type: the outer per-tensor scale
+/// absorbs the tensor's dynamic range, and the block param only covers the spread between blocks.
+/// That spread is still bounded: a block whose scale falls further below the largest one than the
+/// block param can express is stored at that param's smallest value, far too coarse for it, and
+/// every value in the block quantizes to zero. [`QuantParam::UE4M3`] spans about 2^18 this way, so
+/// a tensor holding a genuine outlier can lose its ordinary values.
+#[derive(Clone, Copy, Debug, Hash, PartialEq, Eq)]
+pub struct ScaleLevels {
+    /// Slots past `len` hold [`ScaleLevels::FILLER`] so derived comparisons stay truthful.
+    storage: [ScaleLevel; MAX_LEVELS],
```
nathanielsimard: Use SmallVec instead of an array.

crates/cubecl-common/src/quant/scheme.rs
```diff
@@ -107,76 +149,483 @@ impl QuantScheme {
...
+        let levels = Vec::<ScaleLevel>::deserialize(deserializer)?;
+        Self::try_new(&levels).map_err(serde::de::Error::custom)
+    }
+}
+
+/// The `QuantLevel` enum shape legacy stored schemes carry, kept so those artifacts still load.
+/// Deserialization only. Delete together with the legacy branch of [`QuantScheme`]'s
+/// `Deserialize` once stored schemes have migrated.
+#[derive(Deserialize)]
+enum LegacyQuantLevel {
```
nathanielsimard: We don't want to keep legacy

crates/cubecl-std/src/quant/view.rs
```diff
@@ -20,6 +20,21 @@ use cubecl_core::{
 };
 use half::{bf16, f16};
 
+/// How much of each read's scale already sits in a register; whatever the register does not hold
+/// is read per position through the scales view. The discriminant is comptime, so each variant
+/// compiles its own kernel with nothing of the others in it.
+#[derive(Clone, Copy, CubeType, CubeLaunch)]
+pub enum ScaleInRegister {
```
nathanielsimard: Is it necessary?

crates/cubecl-common/src/quant/scheme.rs
ThierryCantin-Demers: It seems the hand rolled version is what was already used for the `BlockSize`, I imagine to make the `BlockSize` and `QuantScheme` `Copy`able:
```rust
/// Copyable block size, specialized version of `SmallVec`.
#[derive(Clone, Copy, Debug, Hash, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
pub struct BlockSize {
    storage: [u8; MAX_DIMS],
    len: u8,
}
```

crates/cubecl-std/src/quant/view.rs
ThierryCantin-Demers: Every read from a quantized view multiplies the raw value by a scale to recover the real number. `UniformScale` answers one question: how much of that multiplier is the same for every value this view will read? Whatever is shared is loaded once and kept in a single variable. Whatever differs per value must be fetched from memory on every read.

Originally this question didn't need asking. With one scale level there is exactly one scale per position, so every read just looked its scale up. Per-tensor vs per-block wasn't this code's problem, it was the scale layout's: per-block maps coordinates to that block's scale, per-tensor maps every coordinate to the single scale. One code path.

Two-level schemes (NVFP4) broke that. The multiplier is now `block_scale * tensor_scale`: two scales, at different granularities, in different buffers. One lookup can't produce it, and looking both up would re-load the tensor scale, a constant, on every read. So the constant part rides in a variable instead, and there are only three ways the split can go:

* `None`: nothing is shared, each read fetches its own scale. This is the old one-level world.
* `Outer(f32)`: the per-tensor scale is shared and held in the variable; each read fetches its block scale and multiplies the two. This is the two-level case.
* `Whole(f32)`: the entire multiplier is shared, because the caller knows all its reads land in one block (a tile reader does). No scale is fetched at all, and the lookup machinery drops out of the kernel.

The choice is made at compile time, so each case produces its own lean kernel containing only the code it needs.

In one sentence: it records which part of the dequantization multiplier is identical across all reads, so that part is loaded once instead of per-value. One case became three because two-level schemes turned "the scale" into a product of two.

ThierryCantin-Demers:
The scale structure is now the entry point: `per_*` constructs a scheme, `with_*` sets the other fields, `and_*` extends. `QuantParam` is renamed to `ScaleDtype` and `ScaleLevels` to `ScaleSequence`, which stays the private storage behind these constructors.

```rust
// One scale for the whole tensor, stored as f32.
QuantScheme::per_tensor(ScaleDtype::F32);

// One scale per block of 32 values, other fields from Default, adjusted with the setters.
QuantScheme::per_block([32], ScaleDtype::F32).with_value(QuantValue::Q4F);

// Two levels: ue4m3 block scales, normalized by a single per-tensor f32 scale.
QuantScheme::per_block([16], ScaleDtype::UE4M3).and_per_tensor(ScaleDtype::F32);

// Named formats live in quant/presets.rs and expand to the constructors above.
QuantScheme::nvfp4();

// Dynamic level lists (e.g. checkpoint headers) go through the validating path.
QuantScheme::default().with_scales(ScaleSequence::try_new(&levels)?);
```

Also from the review: legacy serde compatibility is gone (stored schemes must be requantized), every hand-written serde impl is replaced by plain derives, and the validity errors derive `Display`/`Error` through `derive_more`.

crates/cubecl-common/src/quant/scheme.rs
```diff
@@ -531,13 +574,53 @@ impl BlockSize {
...
+                    dim
+                } else {
+                    block as usize
+                }
+            })
+            .collect()
+    }
+
+    /// The shape of the grid of blocks laid over a tensor of `shape`: one scale per block.
+    pub fn grid(&self, shape: &[usize]) -> Vec<usize> {
```
nathanielsimard: Unsure what resolved_dims and grid are

ThierryCantin-Demers: resolved_dims is not needed anymore so I have removed it and I have renamed grid to num_blocks so it's clearer (it's basically the number of blocks per dimension, moved up from burn). It had some machinery for the first version of this PR which is not needed anymore.

PR #1516, opened 2026-08-13
Fix/cuda pliron regressions

nathanielsimard:


crates/cubecl-cpp/src/shared/binary.rs
```diff
@@ -310,8 +310,13 @@ no_half!(PowfOp);
...
     let lhs = op.lhs(ctx);
-    let rhs = op.rhs(ctx).name(ctx);
-    format!("pow({}, {rhs})", lhs.name(ctx))
+    let rhs = op.rhs(ctx);
+    // The exponent arrives as an integer, and `pow(float, int)` resolves to CUDA's `powif`,
+    // which the headers declare but nothing defines — ptxas then fails the whole module with
+    // "Unresolved extern function '_Z5powiffi'". Cast to land on the float overload instead,
+    // which is exact for integral exponents, negative bases included.
+    let ty = lhs.get_type(ctx).to_cpp(ctx);
+    format!("pow({}, ({ty}){})", lhs.name(ctx), rhs.name(ctx))
```
wingertge, line 319: Pretty sure something is broken on your driver install because I can't reproduce it on either the 5090 or the 3080, and it's not the kind of error you get from an unsupported feature, only a linking issue.

crates/cubecl-cpp/src/shared/plane.rs
```diff
@@ -87,6 +103,24 @@ pub fn elect() -> bool {
...
+/// Note this elects lane 0 rather than the lowest *active* lane, which is what `elect.sync`
+/// gives you. The two agree whenever the plane is converged, which is the case for every
+/// current caller; they diverge only under non-uniform control flow.
+#[op_interface_impl]
+impl LowerOp for plane::ElectOp {
+    fn lower(&self, scope: &Scope) -> Vec<Value> {
+        vec![elect::expand(scope).read_value(scope)]
+    }
+}
+
```
wingertge, line 123: Better implemented in https://github.com/tracel-ai/cubecl/pull/1517

crates/cubecl-cpp/src/cuda/atomic.rs
```diff
@@ -21,7 +21,20 @@ use crate::{
...
+/// A 2-wide half is instead one 32-bit operand typed `f16x2`, so it carries no qualifier and
+/// packs into a single register.
+fn is_packed_half_pair(ctx: &Context, val: impl Typed + Copy) -> bool {
+    val.is_half(ctx) && val.vector_size(ctx) == 2
+}
+
+fn atom_vec(ctx: &Context, val: impl Typed + Copy) -> &'static str {
+    if is_packed_half_pair(ctx, val) {
+        return "";
+    }
```
wingertge, line 37: This entire file is fixed properly in https://github.com/tracel-ai/cubecl/pull/1517

wingertge (changes requested):

crates/cubecl-cpp/src/shared/binary.rs
wingertge: I found 4 issues about this from various ancient versions and they were each solved by "updating" but from and to different versions. So this is almost certainly a local corruption issue that's fixed by just reinstalling (or updating to a newer version whatever that version is), something in your local CUDA install got messed up somewhere.

PR #1517, opened 2026-08-13
Fix atomics and elect on CUDA and WGSL

wingertge:
Cleaner fixes for some of the issues from https://github.com/tracel-ai/cubecl/pull/1516

Also fixes WGSL, and the elect test


crates/cubecl-cpp/src/cuda/plane.rs
wingertge: This needs to be removed but I'm on my phone

nathanielsimard (approved):

PR #1519, opened 2026-08-14
Fix Pliron timer panic on WebAssembly

syl20bnr:
## Summary

Keep Pliron in `no_std` mode on WebAssembly while preserving its existing
`std` configuration on native targets.

## Problem

CubeCL's `std` features currently forward `pliron/std` from several compiler
crates.

On `wasm32-unknown-unknown`, Pliron's `std` timer uses
`std::time::Instant`. Pliron's pass manager starts this timer when running
compiler passes, causing browser inference to panic with:

```text
time not implemented on this platform
```

This happens even when pass timing output is not enabled because the timer is
constructed before the timing result is conditionally recorded.

## Solution

Stop forwarding `pliron/std` through the target-independent `std` features of:

- `cubecl-core`
- `cubecl-cpp`
- `cubecl-spirv`
- `cubecl-wgpu`

Re-enable `pliron/std` through target-specific native dependencies:

```toml
[target.'cfg(not(target_family = "wasm"))'.dependencies]
pliron = { workspace = true, features = ["std"] }
```

Pliron's existing `no_std` implementation provides the programmatic IR and
pass infrastructure used by CubeCL without calling the unsupported browser
clock. Native builds retain their previous Pliron functionality.

## Validation

- [x] `cargo fmt --all --check`
- [x] WASM feature resolution: `pliron v0.17.0 []`
- [x] Native feature resolution: `pliron v0.17.0 [std]`
- [x] `cargo check -p cubecl-wgpu --features std`
- [x] Built the downstream Metabolic local-compute WASM worker with this
      change and the runtime dependency fix applied

The full downstream WASM validation uses the separate
`wasm-bindgen-futures` dependency fix as well.

## Downstream validation

- [ ] Cubek validation PR: not done :-)
- [ ] Burn validation PR: not done :-)

### DIFF
--- a/crates/cubecl-core/Cargo.toml
+++ b/crates/cubecl-core/Cargo.toml
@@ -22,7 +22,7 @@ workspace = true
 browser-cache = ["cubecl-runtime/browser-cache"]
 default = ["cubecl-runtime/default", "cubecl-ir/default"]
 export_tests = ["tempfile", "test-log/trace", "test-log/log", "std"]
-std = ["cubecl-runtime/std", "pliron/std", "cubecl-environment/std"]
+std = ["cubecl-runtime/std", "cubecl-environment/std"]
 template = []
 tokio = ["cubecl-runtime/tokio"]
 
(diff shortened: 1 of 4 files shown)


laggui (approved):
LGTM as a cubecl-only workaround.

`std` feature forwarding should likely be reinstated once pliron has a wasm-compatible timer or avoids constructing the timer when timing isn't requested.

vaivaswatha:
> LGTM as a cubecl-only workaround.
> 
> `std` feature forwarding should likely be reinstated once pliron has a wasm-compatible timer or avoids constructing the timer when timing isn't requested.

https://github.com/pliron-org/pliron/pull/233

PR #1529, opened 2026-08-18
feat(quant): Software fp8 conversion on every backend, native where the hardware has it

ThierryCantin-Demers:
Software fp8 (e4m3, e5m2) conversion for every backend, native where the hardware has it, one set of semantics everywhere, checked exhaustively against the host codecs.

## Why

`e4m3`/`e5m2` only converted where the hardware did: CUDA through `cuda_fp8.h`, SPIR-V through `OpTypeFloat 8` + `OpFConvert` behind `VK_EXT_shader_float8`. HIP, Metal, CPU and any Vulkan driver without the extension (Intel Arc under Mesa, for one) failed at codegen on the first fp8 cast. That blocks two-level quantization's point: `UE4M3` block scales are what make the second level worth having, and they have to read on the machines with the least memory, not only on Ada and up.

The polyfill is integer arithmetic on `u32` bit patterns and touches no 8- or 16-bit type, so a backend with no byte type at all can still call it on the bits inside a word (that is what a packed-scale binding will use for WGSL later).

## What changed

- `cubecl-core/src/post_processing/minifloat.rs`: `Fp8Format` (constants read off `cubecl_common::{e4m3, e5m2}`, nothing restated), the `#[cube]` polyfills `fp8_bits_to_f32` / `f32_to_fp8_bits`, and `LowerMinifloatCast`, a `MatchRewrite` pass modelled on `saturating.rs` that rewrites any `CastOp` touching an fp8 type the target does not convert natively into reinterpret-to-byte, widen, polyfill, cast. Each compiler wires it with its own `EnumSet<Fp8Format>` of the formats its target converts natively.
- SPIR-V: new `VulkanCompilationOptions::supports_float8`. Without it, `Float8E4M3Type`/`Float8E5M2Type` map to `OpTypeInt 8` and casts go through the pass; a same-type reinterpret forwards its operand instead of emitting an invalid `OpBitcast`; `cast()` widens an emulated fp8 as an unsigned byte. With it, the native float type stays and the fp8-producing `OpFConvert` gets `SaturatedToLargestFloat8NormalConversionEXT`.
- Vulkan registration: fp8 `Conversion` wherever `shaderInt8` or `shaderFloat8` exists, `Buffer` wherever 8-bit storage does. `shaderInt8` is what the emulated path needs, since fp8 is an 8-bit int there, but `VK_EXT_shader_float8` does not depend on it, so gating on int8 alone would have let a device with native fp8 and no int8 compile fp8 while reporting it unsupported. The `storage8` gate read the 16-bit storage feature; it now reads `buf_8`.
- CPU: fp8 maps to LLVM `i8`, pass in the pipeline, `Conversion | Buffer` registered.
- cpp: HIP and Metal map fp8 to `uint8_t` and take the pass; CUDA keeps its intrinsics (both formats marked native), and since `cuda_fp8.hpp` carries its own software path below sm_89 the CUDA runtime registers fp8 on every arch rather than from Ada up.
- CUDA encode: `__NV_SATFINITE` instead of `__NV_NOSAT` (overflow to NaN, and the header's software path even on sm_89; `SATFINITE` is the `cvt.rn.satfinite` instruction), and one `__nv_cvt_{halfraw,bfloat16raw,float,double}_to_fp8` per source width instead of a detour through f16, which rounded twice. Packed pairs keep `halfraw2` / `bfloat16raw2`. `ue8m0` is deliberately outside this: `__nv_cvt_bfloat16raw_to_e8m0` is the only intrinsic for it, so it keeps the bf16 intermediate and the direct-encode check keys on e4m3/e5m2 rather than on the broader "is an 8-bit float" predicate.

## Semantics

Round to nearest even; overflow and infinities saturate to the largest finite value (448 / 57344); NaN stays NaN; below half the smallest subnormal flushes to zero. This is what `e4m3::from_f32` / `e5m2::from_f32` already do, and now what every device path does too. Quantization never overflows (scales are rounded up and saturated, values clamped), so no quantized number changes; the edge cases just agree across backends.

## Tests

- `runtime_tests/minifloat.rs`: all 256 codes of both formats decode bit-exact against the codec; every f16 value plus rounding ties, underflow and saturation edges and the specials encode to the codec's byte; vector sizes 1 and 4. The existing `test_fp8` now runs everywhere. A mutation check (round-half-up instead of RNE) fails on the first tie.
- `cubecl-std`: two-level quantized view with e4m3 block scales, f32 and f16 compute.

Verified on hardware:

| runtime | path | result |
| --- | --- | --- |
| CPU | emulated | `cargo test -p cubecl-cpu` 678 passed |
| wgpu Vulkan, Intel Arc B390 (no `VK_EXT_shader_float8`) | emulated, u8 storage | fp8 and ue4m3 tests pass; suite 650 passed, the 18 `plane_shuffle` failures are pre-existing on main |
| CUDA, RTX 4070 Ti SUPER (sm_89) | native | 12/12 fp8 tests bit-exact; suite 805 passed, `test_cmma_manual` fails on main too |
| HIP, RX 7600 (gfx1102) | emulated | 6/6 fp8 tests |
| Metal | emulated | not run, no hardware reachable |
| SPIR-V with the extension | native | not run, no hardware reachable |

The CUDA row predates the rebase onto main and the e8m0 fix above. sm_89 does not register `UE8M0` at all, so no CUDA machine here can reach the e8m0 cast path either way; it is unverified on hardware and reasoned from the generated source.

## Not in this PR

- WGSL: no 8-bit types, so fp8 stays unregistered there; the packed-scale binding that reaches it calls these polyfills on `u32` words and comes separately.
- fp4 (`e2m1`), fp6, `ue8m0`: same polyfill shape, but fp4 needs lane doubling in the pass and `ue8m0` needs a round-up rule downstream before its cast is useful; each its own PR.
- HIP native fp8 (gfx950 only, gfx942 is FNUZ): parked, no hardware to verify on.
- A compile check for the cpp backends. Nothing in CI builds or tests `cubecl-cuda`, `cubecl-hip` or `cubecl-metal`, so their codegen has no automated cover. Asserting on the generated source text pins how it is spelled rather than whether it is correct, and fails on any refactor that emits equally valid code, so that is not the answer. Feeding the generated source through NVRTC tests the real property and needs the CUDA toolkit but no GPU; separate PR.

## Downstream

No scheme or launch API changes. burn's ue4m3 tests currently gated `#[cfg(not(feature = "wgpu"))]` can become runtime `type_usage` checks; cubek's `e4m3::supported_uses(...)` gates start running on Vulkan/CPU/HIP with no change.

### Instructions

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -890,6 +890,7 @@ dependencies = [
  "bytemuck",
  "cubecl-common",
  "cubecl-core",
+ "cubecl-cpu",
  "cubecl-environment",
  "cubecl-opt",
  "cubecl-runtime",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1178,6 +1178,7 @@ dependencies = [
  "bitflags 2.13.1",
  "cubecl-common",
  "cubecl-core",
+ "cubecl-cpu",
  "cubecl-environment",
  "cubecl-ir",
  "cubecl-opt",
--- a/crates/cubecl-core/src/codegen/compiler.rs
+++ b/crates/cubecl-core/src/codegen/compiler.rs
@@ -17,6 +17,8 @@ pub struct VulkanCompilationOptions {
     pub supports_arbitrary_bitwise: bool,
     pub supports_uniform_standard_layout: bool,
     pub supports_uniform_unsized_array: bool,
+    /// Whether the driver has `VK_EXT_shader_float8`; without it fp8 casts are lowered in software.
+    pub supports_float8: bool,
 
     pub max_spirv_version: (u8, u8),
     pub max_vector_size: usize,
--- a/crates/cubecl-core/src/codegen/compiler.rs
+++ b/crates/cubecl-core/src/codegen/compiler.rs
@@ -17,7 +17,6 @@ pub struct VulkanCompilationOptions {
     pub supports_arbitrary_bitwise: bool,
     pub supports_uniform_standard_layout: bool,
     pub supports_uniform_unsized_array: bool,
-    /// Whether the driver has `VK_EXT_shader_float8`; without it fp8 casts are lowered in software.
     pub supports_float8: bool,
 
     pub max_spirv_version: (u8, u8),
--- a/crates/cubecl-core/src/post_processing/minifloat.rs
+++ b/crates/cubecl-core/src/post_processing/minifloat.rs
@@ -1,0 +1,321 @@
+//! Software fp8 conversion for targets without hardware fp8.
+//!
+//! The polyfills work on `u32` bit patterns and use no 8- or 16-bit types, so they also serve
+//! backends that cannot store a byte, given the bits in a word. [`LowerMinifloatCast`] rewrites
+//! the casts a target cannot convert natively into them.
+
+use cubecl_ir::{
+    NamedRewrite, Scope,
+    dialect::{base::OperationPtrExt, general::CastOp},
+    interfaces::TypedExt,
+    prelude::*,
+    types::scalar::{Float8E4M3Type, Float8E5M2Type},
+};
+use pliron::r#type::TypeHandle;
+
+use crate::{self as cubecl, prelude::*};
+
+define_size!(N);
+
+/// An 8-bit float encoding, by its field widths.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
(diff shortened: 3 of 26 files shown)


ThierryCantin-Demers:
Will be reopened as another PR to cleanup history

PR #1532, opened 2026-08-18
Feat/quant lookup

louisfd:


crates/cubecl-std/src/quant/view.rs
```diff
@@ -271,10 +299,39 @@ fn expand_global(
...
 
+/// Register the lookup table's binding, checking it against the scheme so the two cannot register
+/// apart. Registered as f32 to match the element type [`expand_table`] reads it back with.
+fn register_table<R: Runtime>(
+    table: Option<BufferArg<R>>,
+    scheme: &QuantScheme,
+    launcher: &mut KernelLauncher<R>,
+) -> Option<BufferCompilationArg> {
+    check_table_bindings(scheme, table.is_some());
+    table.map(|table| <[f32] as LaunchArg>::register(table, launcher))
```
ThierryCantin-Demers: `register_table` has the `BufferArg` here and runs on every launch, so `assert_eq!(table.size(), 1 << scheme.size_bits_value())` would catch a codebook built for the wrong bit width. Nothing ties the two right now, and it fails as garbage entries rather than an error. Was leaving it out deliberate, given `check_table_bindings` only gets presence?

crates/cubecl-std/src/tensor/view/launch.rs
```diff
@@ -497,20 +498,40 @@ mod dynamic {
...
+
+        /// Create a new view arg that dequantizes on read through a lookup table
+        /// ([`QuantMode::Lookup`](cubecl_common::quant::scheme::QuantMode::Lookup)): each stored
+        /// field indexes `table` and a read reconstructs `table[field] * scale`.
+        ///
+        /// `table` must hold `2^bits` f32 entries — the mask bounds every index to that range, so
+        /// a shorter buffer is read out of bounds. Panics for a scheme whose mode is not lookup,
+        /// whose store is not packed-u32, or whose value is a minifloat (an index has no float
+        /// semantics; use the integer value of the same width).
+        pub fn new_quantized_lookup(
```
ThierryCantin-Demers: There's no public path for `(Some(global), Some(table))` since `quantized` is private. That's NF4 with double quant, so probably wanted eventually. Not asking for it here, but right now going through this constructor with a two-level scheme panics in `check_global_bindings` instead of just being unsupported.

Wondering if `ScaleBindings` could grow into something like `QuantBindings { scales, table }` during the 1514 rebase? Global's already in there for the same reason, and then NF4 is just a new field later instead of a signature change. `len()` would need to stay scales-only though, since it's checked against `num_levels()`.

crates/cubecl-std/src/quant/dequantize.rs
```diff
@@ -77,6 +87,37 @@ pub fn unpack_cast_u32<F: Numeric, NQ: Size, NF: Size>(
...
+        "unpack_fields: e2m1 decodes in native pairs, which a sub-word line would split"
+    ));
+    let size_bits = scheme.size_bits_value();
+    let mask = comptime![packing_mask(scheme)];
+    let size!(N1) = 1usize;
+
+    let mut out = Vector::<F, NF>::empty();
+    #[unroll]
+    for j in 0..NF::value() {
+        let shift = (first + j as u32) * size_bits as u32;
```
ThierryCantin-Demers, line 133: `first` is runtime and unbounded here. If `first + NF > num_quants` the shift goes past 32 and most ISAs mask it to 5 bits, so you silently get the wrong fields. `scan_words` gets it right, but this is `pub` and the doc doesn't mention the bound.

ThierryCantin-Demers (commented):
Shape looks right, and the test is good, the non-affine table plus permuted nibble walk catches both an integer-cast fallback and a shift error.

Heads up on sequencing: #1514 rewrites most of this, 5 of 6 files conflict on merge. It lands tomorrow so probably moot, just flagging in case someone picks it up first. Better to rebase onto it than the other way. One thing to know going in, #1514 drops `global` from `dequantize_aligned`, so `table` has to stay a real parameter through `unpack_cast_u32` and `cast_masked`.

PR #1540, opened 2026-08-20
fix(metal): correct MSL capability detection

laggui:
Changes in #1512 broke the MacOS CI in burn. `!raw.supportsFamily(MTLGPUFamily::Apple7)` immediately returned, and the WGSL fallback was applied.

I updated the family support check.

MacOS CI now runs the tests: https://github.com/tracel-ai/burn/pull/5396 (looks like some tests fail, but at least now we can detect those lol)

## Follow-up

This fixes the immediate detection issue, but there might be a broader question wr.t. the fallback contract.

Right now, WGSL fallback applies even when `MslCompiler` is requested explicitly. Shouldn't it only apply to `AutoCompiler`, where selecting an available compiler is part of its contract? And `MslCompiler` should produce a clear initialization error instead of silently selecting WGSL.

### DIFF
--- a/crates/cubecl-wgpu/src/backend/metal.rs
+++ b/crates/cubecl-wgpu/src/backend/metal.rs
@@ -106,6 +106,13 @@ pub fn register_metal_features(
             return false;
         };
         let raw = adapter.raw_device();
+        eprintln!(
+            "Metal device={:?}, metal3={}, apple7={}, mac2={}",
+            raw.name(),
+            raw.supportsFamily(MTLGPUFamily::Metal3),
+            raw.supportsFamily(MTLGPUFamily::Apple7),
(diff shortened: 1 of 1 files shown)


wingertge (approved):
Mac2 is listed as deprecated so I assumed they were unified under the Apple category now but apparently they use deprecation to just indicate a target isn't actively being developed anymore. Not really what deprecation is meant for but whatever.

laggui:
Yeah it's a bit unclear, Mac2 is marked as deprecated but it's still in their feature-set table, and the paravirtualized device on Burn's MacOS CI reports it. So it's clearly still used. I think it's more like "deprecated but still reported and required for compatibility" in this case.

PR #1545, opened 2026-08-21
fix(cpu): yield between idle polls instead of spinning

ThierryCantin-Demers:
Dispatcher workers are pinned across every logical CPU, so the 200us idle poll window spinning in earnest starves the client thread and any units still running. On an 8-core machine a 16-unit launch writing 600KB regressed from ~33us to 150-250us medians with millisecond tails; yielding on each missed poll restores 33-42us at 42-54 percent of the measured write peak, while the window still spares barrier kernels a futex wake per worker.

Isolated by toggling each mechanism of #1533 separately on an idle machine (launch geometry pinned, 300 samples): the idle spin alone reproduces the regression; the flush yield and core reordering do not contribute. Capping the number of spinners benched slightly better medians on two of three shapes but reproduced poorly and parks most workers, which defeats the window for barrier kernels.

### DIFF
--- a/crates/cubecl-cpu/src/compute/threadpool/scheduler/dispatcher.rs
+++ b/crates/cubecl-cpu/src/compute/threadpool/scheduler/dispatcher.rs
@@ -122,7 +122,9 @@ impl Worker for DispatcherWorker {
                             received = Some(task);
                             break;
                         }
-                        Err(_) => std::hint::spin_loop(),
+                        // The workers cover every logical CPU, so polling in earnest
+                        // starves the client and any still-running units.
+                        Err(_) => std::thread::yield_now(),
                     }
                 }
                 let task = match received {
--- a/crates/cubecl-cpu/src/compute/threadpool/scheduler/dispatcher.rs
+++ b/crates/cubecl-cpu/src/compute/threadpool/scheduler/dispatcher.rs
@@ -104,19 +104,19 @@ impl DispatcherWorker {
     }
 }
 
-/// How long an idle worker spins on `try_recv` before parking in the blocking
-/// `recv`. A parked worker costs a futex wake per launch — and a barrier
-/// kernel runs at the latency of its last-woken unit — while the budget caps
-/// what an actually-idle pool burns.
-const IDLE_SPIN: std::time::Duration = std::time::Duration::from_micros(200);
+/// How long an idle worker polls `try_recv`, yielding the CPU between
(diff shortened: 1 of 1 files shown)


ThierryCantin-Demers:
The real fix is actually not just switching to yield from spin, it could regress the cases where the spin is the right one. Investigating more to do a better and more general fix.

marcantoinem (approved):

ThierryCantin-Demers:
Validation summary, all on a Ryzen 5700X (8c/16t), measured against stock at the same rev with geometry pinned.

Alternatives benched before settling on the one-liner: counter-bounded poll windows (no effect, the Instant read is not a cost), a tiered pause-then-yield window (mid-pack, no win anywhere), capping how many workers spin at once (best medians on two shapes but irreproducible run to run, and it parks most workers, defeating the window for barrier kernels), and gating spin-vs-yield on a busy-worker count (reproduces stock's slow behavior: workers finish in lockstep, so a pool-visible signal reads idle exactly during the harmful window; the starved parties are the unpinned client and the bus, which only the OS scheduler sees, and yielding is what consults it).

Four-regime validation of the one-liner:

1. Kernel sweep (cubek random/reduce/unary/contiguous/gemm_cpu on the cpu runtime, 50 samples): no regressions; 16-unit memory-bound launches improve 2.4-7.3x.
2. Barrier kernels, hot (1000 back-to-back sync_cube launches): 8.8x better median at cube_dim=16, and yield eliminates a stall stock hits on 12 percent of hot launches (~3.05ms tail, gone). Oversubscribed cube_dim=32 is better at every percentile.
3. Cold launches (1ms spacing, so every launch takes the parked path): cube_dim=32 neutral. At cube_dim=16 the median improves (230us vs 354us) but 28-46 percent of samples land in a tight ~3.05ms cluster stock's cold path does not show. The magnitude matches the stall stock exhibits hot, so this looks like one pre-existing ~3ms mechanism that stock trips on hot traffic and yield trips on cold, at that one geometry. Worth its own investigation; the harness also exercises the shared-memory flush path.
4. taskset -c 0-3 (16 workers over 4 cores): neutral to better, no new contention effect.

Net: yield wins the common cases outright, the poll window keeps its wake-free pickup for barrier kernels, and the one exposed tail needs cold plus barrier plus exact-core-count simultaneously and still wins the median there.

nathanielsimard (approved):

PR #1546, opened 2026-08-21
feat(bench): write-only memory probe and roofline scoring for benchmarks

ThierryCantin-Demers:
Benchmarks get honest peaks to compare against, and the machinery to do the comparing: the write probe, the roofline scoring, and the probe-honesty fixes the first real measurements exposed.

**Write-only probe.** `MemoryAccess::Write` and `ThroughputMode::MemoryWrite`: a store-only runner symmetric with the read probe (rotating cold windows, stores seeded from a runtime scalar so they cannot fold into a memset). Store-bound kernels (RNG fills, memsets, broadcasts) are judged against the copy peak today, which spends half its traffic on reads they never issue. With ordinary stores the probe pays read-for-ownership where the hardware does, which keeps it the honest ceiling: on Zen 3 the write peak reads roughly half the bus (RFO paid), on recent Intel and on GPUs write lands near read (full-line store streams get the RFO elided).

**Work declaration and roofline scoring.** `Work` moves to cubecl-common so autotune bounds and benchmarks share one definition, `Benchmark::work()` optionally declares a run's work, and a neutral `ResourceBound` (amount against a peak rate) in the throughput module carries the roofline arithmetic: `time_at_peak`, per-resource scoring against a measured duration, and the binding-resource convention (largest time at peak, equivalently the highest achieved fraction). `AutotuneBound` becomes that record plus its threshold. The persisted bound shape changes; old cache entries are dropped with the store's existing corrupted-entry warning and retune.

**Probe fix: a real backing buffer for read and copy.** The read and copy probes measured a buffer nothing had written. A fresh `alloc_zeroed` allocation is backed by the kernel's shared zero page until first write, so the "DRAM reads" were largely one cached page: a 5700X reported 95.8 GB/s of reads against a ~51 GB/s physical ceiling, flat far past L3, with the loads verifiably surviving O3. Priming the buffer once outside the timed path brings it to 44 GB/s, consistent with a hand-written reference scan (43.5). CPU read and copy peaks measured before this fix were inflated.

**Probe fix: CPU-shaped geometry and layout.** The probes launched an SM-derived grid (the SM count defaults to 64 with no SMs reported, so the CPU got 2048 cubes of tiny strided slivers) and addressed lines coalesced, adjacent threads adjacent lines, which on a CPU makes every worker's consecutive accesses stride apart. With `num_cpu_cores` reported the launch is now `cube_dim` = one unit per core, and the memory probes give each unit a contiguous run. GPU geometry and addressing are unchanged.

**Probe fix: cube positions for compute, one cube for memory.** On this backend a cube's units are the real dispatched workers and cube positions loop sequentially inside them, wrapped around the whole kernel body. Compute probes need several positions per worker: each position is an independent chain, and the pipelining across them is where their ILP comes from (a single position left a heterogeneous laptop latency-bound at 15 GOPS instead of 201). Memory probes need exactly one: with blocked addressing, more positions replay a narrow cache-resident slice and the sweep stops measuring DRAM. So the CPU launch carries `cube_count = 64` and `MemoryProbe` pins its own count to 1 when blocked.

Corrected numbers, 5700X (DDR4, ~51 GB/s physical): read 44.1, copy 26.1, write 19.9 GB/s, compute 92 GOPS, all flat across the working-set sweep. CUDA and Vulkan on the same machine agree within a few percent of each other and are unchanged by the CPU-side fixes. Two known limits, documented in code: the write probe trails a hand-tuned contiguous fill by ~15 percent (windowed bookkeeping), so a well-shaped store kernel can legitimately score above 100 percent of this peak, and the summary type documents a >1 fraction as a finding about the probe; on heterogeneous P/E-core CPUs the equal per-core split under-reports write (~58 vs ~80 GB/s on a Lunar Lake laptop), since dispatch is static and a slow core keeps its fixed share, which needs topology-aware partitioning to fix properly.

First consumer will be cubek's bench registry, which currently judges its RNG fill against the copy peak.


SamuelBelanger (approved):
LGTM

louisfd (approved):

PR #1549, opened 2026-08-23
fix(cpp): emit casts using destination type

### LINKED ISSUE #1522: Metal (native runtime): generated MSL for f32->bf16 conversion does not compile — 'cannot initialize const bfloat with an rvalue of type float'
### Summary

At rev `f0721ea` the Metal (native `cubecl-metal`) runtime **registers** BF16 (`crates/cubecl-metal/src/runtime.rs:218`) but the MSL it generates for f32→bf16 conversion does not compile: kernels contain direct initialization of `bfloat` from a `float` rvalue, which MSL rejects.

```
Failed to compile MSL: "program_source:181:14: error: cannot initialize a variable of type 'const bfloat' with an rvalue of type 'float'
bfloat const v170 = float(v166);
             ^      ~~~~~~~~~~~
program_source:182:27: error: cannot initialize a variable of type 'const bfloat' with an rvalue of type 'float'
*v213 = v170;bfloat const v172 = float(v169);
```

MSL 3.1 `bfloat` needs an explicit conversion (`bfloat(x)` / `static_cast<bfloat>(x)`); the emitted `float(...)` wrapper produces a `float` on the RHS. The emission appears to come from the cubecl-cpp conversion/rewrite path rather than the Metal type table (`crates/cubecl-cpp/src/metal/ty.rs:28` maps BF16→`bfloat` correctly).

### Reproducer (~20 s, no model)

Any burn tensor graph on `MetalRuntime` that casts f32→bf16 and runs backward — e.g. create two f32 tensors on the metal device, `.cast(BF16)`, matmul, `sum().backward()`. First launched cast kernel fails compilation deterministically. (We hit it via an 8B LoRA step and via a small raw-op differential harness; both fail identically at the first cast kernel.)

### Environment

- Apple M1 Max, macOS 14.x (Metal 3.1), Apple7 GPU family
- cubecl `f0721ea4b11bd56d30c2a24e6a0548246df9724f` via burn `5b3bd6c`
- `cubecl-metal` with `metal-native`; MSL 3.1, `MTLMathMode::Safe`

### Note

The wgpu-Metal path has the mirrored gap: it **compiles** bf16 fine (same cubecl-cpp MSL) but does not **register** it — that's #1509. Each Metal path currently has the other's missing half.

y7nieSEl5:
## Summary

- emit shared C++ casts using the `CastOp` destination type instead of reconstructing the source type
- add a native Metal regression test for f32-to-bf16 conversion, including rounding boundaries, signed zero, subnormals, range extremes, infinities, and NaN

## Background

The shared emitter declared the cast result with its destination type but formatted the right-hand side with the input type. For f32-to-bf16 this generated invalid MSL like:

```metal
bfloat const v18 = float(v17);
```

Using the result type restores the cast semantics and emits the explicit conversion Metal requires:

```metal
bfloat const v18 = bfloat(v17);
```

The destination-typed behavior existed before the Pliron migration in #1402; the source-typed constructor was introduced during that refactor.

## Relationship to #1509

#1509 and this PR address separate layers. #1509 registers BF16 as a WGPU/MSL runtime capability, while this PR fixes the shared C++ `CastOp` emitter used by native Metal and WGPU/MSL.

On an Apple M5, locally enabling the same full BF16 registration produced by #1509 after its probe succeeds made WGPU report `Conversion | Arithmetic | DotProduct | Buffer`. An actual f32-to-bf16 kernel still emitted `bfloat = float(...)` and failed Metal compilation without this change. With this change it emitted `bfloat = bfloat(...)`, compiled, and returned the expected values.

## Testing

- `cargo check -p cubecl-cpp --all-features`
- `cargo test -p cubecl-metal f32_to_bf16_cast_compiles_and_runs -- --nocapture`
  - passed on an Apple M5 MacBook Air with a 10-core GPU, macOS 26.5.2
  - non-NaN results matched `half::bf16::from_f32` bit-for-bit; NaN was verified by classification
- restoring the old emitter reproduced the Metal compilation error deterministically
- enabling BF16 registration with the old emitter reproduced the same failure through WGPU/MSL

## Performance

No measurable performance regression was observed on the Apple M5. Since f32-to-bf16 does not compile before the fix, the shared `CastOp` emitter was compared using f32-to-f16, which compiles on both revisions and exercises the same changed path.

The device-timed benchmark used warmed compilation and allocations, 11 alternating samples of 200 large-buffer launches, and an unchanged f32 copy as a drift control. The median normalized cast/copy ratio was `0.760752` before the fix and `0.759177` after it, a `-0.21%` difference within normal run-to-run dispersion.

Fixes #1522.

### DIFF
--- a/crates/cubecl-cpp/src/shared/unary.rs
+++ b/crates/cubecl-cpp/src/shared/unary.rs
@@ -185,7 +185,7 @@ unrolling!(FindFirstSetOp);
 
 shared_op_with_out!(CastOp, |op, ctx| {
     let input = op.input(ctx);
-    let ty = input.get_type(ctx);
+    let ty = op.get_result(ctx).get_type(ctx);
     format!("{}({})", ty.to_cpp(ctx), input.name(ctx))
 });
 unrolling!(CastOp);
--- a/crates/cubecl-metal/src/lib.rs
+++ b/crates/cubecl-metal/src/lib.rs
@@ -14,6 +14,8 @@ pub use runtime::MetalRuntime;
 pub(crate) type MetalCompiler = cubecl_cpp::shared::CppCompiler<cubecl_cpp::target::Metal>;
 
 #[cfg(test)]
+mod tests_bf16_cast;
+#[cfg(test)]
 mod tests_expm1;
 #[cfg(test)]
 mod tests_launch_errors;
--- a/crates/cubecl-metal/src/tests_bf16_cast.rs
+++ b/crates/cubecl-metal/src/tests_bf16_cast.rs
@@ -1,0 +1,38 @@
+use cubecl_core::{self as cubecl, prelude::*};
+use half::bf16;
+
+type R = crate::MetalRuntime;
+
+#[cube(launch_unchecked)]
+fn f32_to_bf16_cast(input: &[f32], output: &mut [bf16]) {
+    if ABSOLUTE_POS < output.len() {
+        output[ABSOLUTE_POS] = bf16::cast_from(input[ABSOLUTE_POS]);
+    }
+}
+
+#[test]
+fn f32_to_bf16_cast_compiles_and_runs() {
+    let client = R::client(&Default::default());
+    let input = [1.0f32, -2.5, 3.25, 0.0];
+    let len = input.len();
+
+    let input_handle = client.create_from_slice(f32::as_bytes(&input));
+    let output_handle = client.empty(len * core::mem::size_of::<bf16>());
+
+    unsafe {
+        f32_to_bf16_cast::launch_unchecked::<R>(
+            &client,
+            CubeCount::Static(1, 1, 1),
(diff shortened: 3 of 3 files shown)


y7nieSEl5:
Reopening after testing the interaction with #1509 more closely. The 2 PRs actually address separate layers. This PR fixes the shared C++ `CastOp` emitter used by both native Metal and WGPU/MSL.
On an M5, enabling #1509-equivalent full BF16 registration made WGPU report `Conversion | Arithmetic | DotProduct | Buffer`, but without #1549 an f32-to-bf16 kernel still emitted `bfloat = float(...)` and failed Metal compilation. With #1549 it emits `bfloat = bfloat(...)` and passes.
I originally closed this as a duplicate, but now I guess it's somehow worth reviewing.

wingertge (approved):

PR #1553, opened 2026-08-24
feat(cpu): Vectorize exp,log,sin,cos,tanh with polynomial approximation

ThierryCantin-Demers:
The CPU JIT builds its LLJIT with `new_with_default_builder()`, so there is no vector math
library behind a vector transcendental: `llvm.exp(<8 x float>)` reaches codegen as a bare
`declare` and every lane becomes a libm call.

`exp`, `log`, `sin`, `cos` and `tanh` now lower to short polynomials instead, as a `LowerOp`
under `crates/cubecl-llvm/src/shared/polyfill/transcendental/`. CPU only. Nothing was
removed: `LowerComplexOpPass` already runs ahead of `CubeToLLVMPass`, and a `should_lower`
that declines leaves the existing intrinsic lowering as the fallback.

## Speed

Ryzen 7 5700X, twenty samples, chained at a line of eight, net of a control kernel carrying
the same traffic without a transcendental. 67.1M evaluations:

| | libm | polynomial | |
|---|---|---|---|
| `exp` | 17.92ms | 4.50ms | 4.0x |
| `log` | 20.14ms | 6.11ms | 3.3x |
| `sin` | 23.24ms | 5.30ms | 4.4x |
| `cos` | 23.33ms | 5.37ms | 4.3x |
| `tanh` | 37.59ms | 7.95ms | 4.7x |

End to end with no kernel edited, a vectorised normal draw falls from 26.6ms to 7.4ms at
33.5M elements, 3.21ms to 0.64ms at 4.2M, and 128us to 41us vocabulary-sized. The uniform
and Bernoulli draws touch no transcendental and do not move, which is the control saying
this is the polynomial and not the build. In a streaming pass all five land on the memory
floor, where `tanh` had been at twice it.

## Accuracy

1.09 to 2.03 units in the last place, against 0.50 to 0.69 for the libm calls being
replaced. A real trade, and still better than this project's GPU backends deliver for the
same functions.

One capability is given up: `sin` and `cos` past 2^20 radians return NaN. Everything else
matches the library, including `log(0) = -inf`, `log(-1) = NaN`, the subnormals, and the
sign of a zero through `sin`.

<details><summary>Why 2^20, and how the constants are pinned</summary>

Measured by decade, the worst absolute error holds near 7e-8 through 1e5, reaches 2.8e-7 at
1e6 and 1.5e-4 at 1e7, and by 1e8 the answer is arbitrary. glibc does full Payne-Hanek
reduction and stays correct throughout, so this is a capability given up, not only
precision. A million radians is a wider promise than most vector libraries make and covers
a rotary embedding at a million positions.

Every constant is either derived in the source or pinned by a test. The Cody-Waite splits
come from a const fn that clears the low mantissa bits, with tests for the exactness that
buys. Each coefficient block names its function, interval, objective and degree, and a test
evaluates the checked-in values over that interval against the accuracy claimed, which is
what stands in for checking a minimax digit against a formula it does not have.
</details>

## The one decision I would like ruled on

The polyfill declines a scalar:

```rust
input.vector_size(ctx) > 1 && !input.scalar_ty(ctx).is_float64(ctx)
```

Ratio of library call to polynomial, net of control. Above 1.00 the polynomial wins:

| | vec1 | vec2 | vec4 | vec8 |
|---|---|---|---|---|
| `exp` | 0.87 | 1.23 | 2.10 | 3.97 |
| `log` | 0.75 | 1.12 | 1.78 | 3.27 |
| `sin` | 1.16 | 1.48 | 2.37 | 3.36 |
| `cos` | 1.08 | 1.45 | 2.28 | 4.27 |
| `tanh` | 0.86 | 1.26 | 2.54 | 4.60 |

The crossover for `exp`, `log` and `tanh` falls between one and two, which is where the gate
sits. The library row is flat across widths, `tanh` reading 38.65, 39.70, 38.41 and
37.22ms, because eight lanes still cost eight calls, while the polynomial row halves at
every doubling. The thin cell is `log` at vec2, 1.12 here and 1.04 on the independent
variant, which is break-even rather than a win.

**What the gate costs:** a scalar `exp` and an eight-wide `exp` no longer return the same
bits, so a kernel whose vectorisation an autotuner picks has a result that moves with the
choice. If that is the wrong trade for this project, the alternative is polyfilling every
width and accepting the scalar regression, and I have the numbers for either.

Double precision is declined because fitting a second set of coefficients would double the
surface to verify for a format no kernel here reaches for.

<details><summary>Why sin and cos are declined at a scalar even though they win there</summary>

Letting them through hands the FFT back a 1.5x regression, which was never about arithmetic
speed. In a loop of its own the polyfill runs a scalar cosine and sine pair in 43.7ms
against libm's 70.4ms, and the kernel still lost half. Comparing the compiled `rfft_kernel`
both ways, the polyfill version spills less, 72 stack references against 102, and is 7
instructions longer. What changed is division: three at IR level became seven, including a
`vdivss` the baseline does not have. In that loop `m` is a power of two, so `theta = ... / m`
is an exponent adjustment, and LLVM proved it while the body held two opaque libm calls and
stopped proving it once a polynomial was inlined there. Ungated, a scalar cosine kernel lost
a quarter and the FFT lost half; gated, both are byte-identical to before.
</details>

## Tests

Nothing in cubecl evaluated `exp` or `log` through a kernel at all, and `sin`, `cos` and
`tanh` only at `0`, `pi/2`, `pi` and `-pi/2`, which a polynomial reaches without its range
reduction ever running a quadrant.

- `runtime_tests/unary.rs` gains `exp` and `log` at line sizes one, two and four, and the
  three already there gain a block spanning several turns either side of zero and both
  sides of the half where `tanh` changes form. Every input is exactly representable in `f16`.
- `cubecl-cpu` gains the edge contract, against the library on the same host, under a kernel
  carrying no `FastMath`. It cannot live in `cubecl-core`, where it would be asserted for
  wgpu and CUDA too.

Green on the CPU backend (724), on CUDA on an RTX 4070 Ti SUPER (136 unary across bf16, f16,
f32 and f64), and on wgpu over Intel Vulkan. `xtask check format` and `xtask check --ci lint`
both pass.

## Not in this PR, on purpose

- **`pow`.** `exp2(b * log2(a))` measures 22 ULP on a typical domain and 279 on a wide one,
  because the exponent multiplies the logarithm's absolute error, so it needs a two-part
  `log2`.
- **Dropping `log`'s NaN and infinity selects** under `NotNaN`/`NotInf`. Worth about a fifth
  of the evaluation.
- **Folding a constant multiplier into the range reduction**, so `cos(c * x)` skips
  Cody-Waite entirely. Worth about a tenth on a `cos`/`sin` pair, and every FFT twiddle has
  that shape. Needs reassociation permission.
- **Writing the polynomials as `a * b + c`** and letting #1550's `InstCombinePass` fuse them.

<details><summary>Why those last three cannot be done yet</summary>

**There is no fast-math gate to reach for.** `#[cube(fast_math = ...)]` calls
`fast_math_expand`, which sets `GlobalState.modes.fp_math_mode`, runs the body and restores
the previous value. Nothing in the repository reads that field, and cubecl-llvm hands
`FastmathFlagsAttr::default()` to every instruction it builds. Reading it from a `LowerOp`
would not work either: the pass runs after expansion, so the mode has already been restored,
and no per-operation attribute records what a given operation was expanded under. Either
change means wiring `FastMath` up first, starting by carrying the mode on the operations
themselves. I can open an issue for that if it is wanted.

**And `InstCombinePass` never sees polyfill arithmetic.** It runs three passes ahead of
`LowerComplexOpPass`, and nothing fuses afterwards either, since without `contract` LLVM may
not pair an `fmul` with an `fadd`. One width-eight kernel's forty `llvm.fmuladd` would
become forty unfused pairs. Moving the pass after `LowerComplexOpPass` does fuse them and
pays elsewhere too, a vector `erf` going from 8 `fmul` and 5 `fadd` to 3 `fmul` and 5
`llvm.fmuladd`, 57 instructions to 52. But it also trips something underneath: in `arc_sinh`
the only IR it changes is `x * x + 1` becoming one `fmuladd`, and LLVM then segfaults
optimizing the 2-wide f16 kernel. The IR looks valid, since `<2 x half> @llvm.fmuladd` and
`sqrt(fma(x, x, one))` both compile fine on their own, so this reads as a bug in the LLVM
build rather than in the pass. And `as_fma` matches `FAddOp` only, so the five Cody-Waite
subtractions here would not fuse regardless.
</details>

### DIFF
--- a/crates/cubecl-core/src/runtime_tests/unary.rs
+++ b/crates/cubecl-core/src/runtime_tests/unary.rs
@@ -286,6 +286,24 @@ test_unary_impl!(test_sin, F, Vector::sin, [
         out_vectorization: 4,
         input: as_type![F: 0., 1.570_796_4, 3.141_592_7, -1.570_796_4],
         expected: as_type![F: 0., 1., 0., -1.]
+    },
+    {
+        input_vectorization: 1,
+        out_vectorization: 1,
+        input: as_type![F: 4.5, 6., -7.5, 8., -12.5, 12., 20., -25.],
+        expected: as_type![F: -0.977_530_1, -0.279_415_5, -0.937_999_96, 0.989_358_25, 0.066_321_895, -0.536_572_93, 0.912_945_3, 0.132_351_76]
+    },
+    {
+        input_vectorization: 2,
+        out_vectorization: 2,
+        input: as_type![F: 4.5, 6., -7.5, 8., -12.5, 12., 20., -25.],
+        expected: as_type![F: -0.977_530_1, -0.279_415_5, -0.937_999_96, 0.989_358_25, 0.066_321_895, -0.536_572_93, 0.912_945_3, 0.132_351_76]
+    },
+    {
+        input_vectorization: 4,
+        out_vectorization: 4,
+        input: as_type![F: 4.5, 6., -7.5, 8., -12.5, 12., 20., -25.],
+        expected: as_type![F: -0.977_530_1, -0.279_415_5, -0.937_999_96, 0.989_358_25, 0.066_321_895, -0.536_572_93, 0.912_945_3, 0.132_351_76]
     }
 ]);
 
@@ -328,6 +346,24 @@ test_unary_impl!(test_cos, F, Vector::cos, [
         out_vectorization: 4,
         input: as_type![F: 0., 1.570_796_4, 3.141_592_7, -1.570_796_4],
         expected: as_type![F: 1., 0., -1., 0.]
+    },
+    {
+        input_vectorization: 1,
+        out_vectorization: 1,
+        input: as_type![F: 4.5, 6., -7.5, 8., -12.5, 12., 20., -25.],
+        expected: as_type![F: -0.210_795_8, 0.960_170_27, 0.346_635_3, -0.145_500_03, 0.997_798_26, 0.843_853_95, 0.408_082_07, 0.991_202_83]
+    },
+    {
+        input_vectorization: 2,
+        out_vectorization: 2,
+        input: as_type![F: 4.5, 6., -7.5, 8., -12.5, 12., 20., -25.],
+        expected: as_type![F: -0.210_795_8, 0.960_170_27, 0.346_635_3, -0.145_500_03, 0.997_798_26, 0.843_853_95, 0.408_082_07, 0.991_202_83]
+    },
+    {
+        input_vectorization: 4,
+        out_vectorization: 4,
+        input: as_type![F: 4.5, 6., -7.5, 8., -12.5, 12., 20., -25.],
+        expected: as_type![F: -0.210_795_8, 0.960_170_27, 0.346_635_3, -0.145_500_03, 0.997_798_26, 0.843_853_95, 0.408_082_07, 0.991_202_83]
     }
 ]);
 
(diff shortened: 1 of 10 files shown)


marcantoinem (approved):
LGTM, will be used with the AMDGPU target

PR #1572, opened 2026-08-28
Add data flow solver

wingertge:
Adds a new solver for DFAs and re-implements the upstream SCCP implementation using the solver as a test case.
The upstream version already handled 99% of all cases so performance impact for the SCCP replacement is negligible (possibly 0.3% but it might be margin of error), but the solver drastically simplifies more impactful future optimizations like partial redundancy elimination, common subexpression elimination, load after store, etc. SCCP just serves as a good way to test it, and it's technically slightly better because it now supports region args and results.

Adds the upstream SCCP tests plus some extra region-specific ones to `cubecl-opt`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1052,7 +1052,6 @@ dependencies = [
  "stable-vec",
  "test-log",
  "thiserror",
- "tracel-llvm-bundler",
  "tynm",
  "type-map",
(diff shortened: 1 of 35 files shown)


nathanielsimard (approved):
LGTM

PR #1582, opened 2026-08-31
Refactor/amdgpu llvm review

nathanielsimard:
A review pass over the AMDGPU LLVM backend landed in #1571 and #1577.
No new features. Mostly the same code paths written once instead of twice, plus the failures they were hiding.


crates/cubecl-cpp/src/hip/arch.rs
```diff
@@ -1,80 +1,12 @@
...
-        } else if norm == "gfx90a" {
-            Ok(AMDArchitecture::GFX90A)
-        } else if norm.starts_with("gfx94") {
-            Ok(AMDArchitecture::GFX94)
-        } else {
-            Ok(AMDArchitecture::Other)
-        }
-    }
-}
+pub use cubecl_core::ir::amd::{AMDArchitecture, AmdWmma};
```
marcantoinem, line 3: It's quite strange to have in cubecl-ir something backend dependant, but I don't think there's a cleaner option

crates/cubecl-ir/src/lib.rs
```diff
@@ -16,6 +16,7 @@ mod value;
 #[cfg(feature = "spirv")]
 mod spirv;
 
+pub mod amd;
```
marcantoinem, line 20: Maybe gating it like spir-v does

crates/cubecl-llvm/src/amdgpu/intrinsic.rs
```diff
@@ -0,0 +1,83 @@
+//! Calling the `llvm.amdgcn` intrinsics, and the lane index the cross-lane lowerings
+//! start from.
+//!
+//! Two stages reach for these: [`builtins`](super::builtins) substitutes
```
marcantoinem: Slop comment

marcantoinem (commented):
In general LGTM

PR #1588, opened 2026-09-01
fix: stop advertising tensor cores on dies that have none

ThierryCantin-Demers:
Two backends advertise tensor-core support on dies that have none. The capability lands in `properties.features.matmul.cmma`, the gate everything downstream reads, so the result is not a refusal but a number that looks like any other.

## CUDA

`supported_cmma_combinations_*` gate on compute capability alone, and TU116/TU117 report 7.5 exactly like every Turing that has tensor cores. The kernel still compiles and runs: `mma.sync` is part of sm_75's ISA and falls back to the ordinary FP16 pipeline. Measured through `examples/throughput` on one machine holding both cards:

| card | compute-cmma |
|---|---|
| RTX 2060 | 60.67 TOPS/s |
| GTX 1660 SUPER | 1.17 TOPS/s |

`CudaArchitecture` now carries `tensor_cores`, decided by the device name because CUDA exposes no attribute for it. No GTX-branded Turing has tensor cores, and every Turing that does shipped under another brand, so the two coincide exactly at 7.5. Two limits: the professional TU117 parts (T400 through T1000) also lack them and are not caught, and a driver that will not name the device keeps the old behaviour rather than silently disabling CMMA on hardware that has it.

## AMD

`AMDArchitecture::parse` matched the `gfx10` prefix, folding gfx101x (RDNA1) into the RDNA2 variant. Both are wave32, which is why nothing keying on `plane_dim` noticed, but RDNA1 has no WMMA instructions at all: they arrive with RDNA3, and rocWMMA does not target gfx101x. The new `GFX101` variant reports wave32 and no WMMA. Folding into `Other` instead would have made the cards unusable, because the runtime refuses a device whose wave width it does not know.

The CUDA half is measured on hardware, the numbers above. The AMD half is unit-tested only: gfx1010 is outside ROCm's support matrix and I have no RDNA1 machine, so nothing here has executed on one.

Draft while the cross-repo validation below is outstanding.

---

### Instructions

### DIFF
--- a/crates/cubecl-cpp/src/cuda/arch.rs
+++ b/crates/cubecl-cpp/src/cuda/arch.rs
@@ -15,11 +15,13 @@ impl CudaArchitecture {
     /// Turing has tensor cores and every Turing that does shipped under another brand, so the
     /// two coincide exactly at 7.5. The professional TU117 parts (T400 to T1000) are not caught.
     pub fn has_tensor_cores(version: u32, name: &str) -> bool {
-        // Tensor cores arrive with Volta.
-        if version < 70 {
-            return false;
+        match version {
+            // Tensor cores arrive with Volta.
+            ..70 => false,
+            // The GTX-branded Turings are the only later dies without them.
+            75 => !name.to_uppercase().contains("GTX"),
+            _ => true,
         }
-        !(version == 75 && name.to_uppercase().contains("GTX"))
     }
 }
 
@@ -59,6 +61,8 @@ mod tests {
             52,
             "NVIDIA GeForce GTX 980"
         ));
+        // Volta itself has them: the boundary is inclusive.
+        assert!(CudaArchitecture::has_tensor_cores(70, "Tesla V100-SXM2"));
     }
 
     #[test]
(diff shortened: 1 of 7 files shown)


crates/cubecl-cpp/src/cuda/arch.rs
```diff
@@ -5,6 +5,69 @@ use crate::shared::Architecture;
...
+    /// Decided by the marketing name because CUDA exposes no attribute for it. No GTX-branded
+    /// Turing has tensor cores and every Turing that does shipped under another brand, so the
+    /// two coincide exactly at 7.5. The professional TU117 parts (T400 to T1000) are not caught.
+    pub fn has_tensor_cores(version: u32, name: &str) -> bool {
+        // Tensor cores arrive with Volta.
+        if version < 70 {
+            return false;
+        }
+        !(version == 75 && name.to_uppercase().contains("GTX"))
+    }
```
wingertge, line 25: That's a very confusing way to implement that condition. I'd implement it as
```rust
    pub fn has_tensor_cores(version: u32, name: &str) -> bool {
        // Tensor cores arrive with Volta.
        if version > 75 {
            return true;
        }
        version == 75 && !name.to_uppercase().contains("GTX")
    }
```

wingertge: That's even better

ThierryCantin-Demers: I changed it to a match, i think it reads clearer this way in the end

wingertge (approved):

PR #1594, opened 2026-09-02
fix(throughput): key the cache on the device part, and version the probes

ThierryCantin-Demers:
Now stacked on the probes PR (`dev/throughput-probes`), which is the API-breaking one and goes first; this PR's own five commits are unchanged, rebased with one conflict in `client.rs` where both branches changed the cache lookup.

The throughput cache namespace was `throughput/<crate version>/<runtime>_dev<index>`, and the index is the one thing about a device that a user changes. Pinning a card with `CUDA_VISIBLE_DEVICES`, which is how anyone selects one, makes whichever card was pinned index 0 — so two different cards measured one after the other on one machine key to one entry.

On a machine holding an RTX 2060 and a GTX 1660 SUPER, cache on, same build, one run after the other:

| | 2060, main | 1660S, main | 2060, branch | 1660S, branch |
| --- | --- | --- | --- | --- |
| compute-direct f16 | 11.1868 | **11.1868** | 11.0921 | **7.9767** |
| memory 1 GiB | 252.6 | **252.6** | 252.5 | **273.5** |
| memory-read | 304.2 | **304.2** | 305.6 | **316.1** |
| memory-write | 257.3 | **257.3** | 254.9 | **282.2** |
| launch | 8.156 µs | **8.156 µs** | 9.085 µs | **8.138 µs** |

The 1660 SUPER's whole row is the 2060's, digit for digit. Its own f16 peak is 7.98, so a denominator every percent-of-peak figure divides by was 40 % wrong with nothing in the output saying so.

## What changed

**Key on the part.** `DeviceProperties::identity` already carries what names it: the device's own name and the fingerprint the runtime compiles for. The fingerprint alone does not separate these two cards, which both report `ptx_sm75`; the name does. Two identical cards now share an entry, taken deliberately — identical parts have the same peak, whereas the index was not buying separation, it was assigning the wrong number.

Keying a measurement cache on `identity.name` is not the gating its doc forbade. That rule protects correctness: a name collision must never decide a capability or address a compiled artifact, because two parts sharing a name would then run each other's code. The doc is reworded to say what it protects rather than to forbid every use.

**Version the probes.** The crate version answers "which release wrote this", not "what did the probe measure", and those come apart inside a release: within `0.11.0-pre.3` one commit moved CUDA memory write +39 % and copy +14 %, and another moves compute-direct f32 +45 %. `PROBE_VERSION` sits beside the throughput keys so it lands in the diff of anyone changing what a probe reports, and is its own namespace segment. It is a bare counter, disposable once the probes stop moving.

**Reclaim what it strands.** A version bump alone leaves the old namespace in the store for good, since nothing reads it again and nothing deletes it. Opening the throughput cache now purges the namespaces this build wrote under an earlier probe version, once per process — scoped to this crate version, because another may belong to a cubecl still in use, and a *later* probe version to a build running right now. Planting namespaces around a run:

```
throughput/<this version>/probe-v0/...   dropped
throughput/<this version>/cuda_dev0      dropped   (predates the version)
throughput/<this version>/probe-v2/...   kept      (a build running right now)
throughput/<other version>/probe-v1/...  kept      (a cubecl still in use)
autotune/<this version>/...              kept
```

`persistence::namespaces` is the enumeration this needs, beside `open` and abstracting the backend the same way; it answers empty where the backend cannot enumerate itself.

Namespaces now read `throughput/0.11.0-pre.3/probe-v1/cuda_ptx_sm75_NVIDIA-GeForce-RTX-2060`. Every component is reduced to characters that can sit in one path segment, since a runtime names itself `wgpu<spirv>`.

**This invalidates every value cached under the old namespace, which is intended:** those entries cannot be told apart from the ones this fixes.

## Not fixed here

A bad measurement is still frozen: the cache returns the first value written and never re-measures, so an outlier from an unstable host persists for the life of the namespace. Every fix for that is a tunable heuristic, and on stable hardware the estimator holds to 0.15 % across runs, so it is left alone deliberately rather than overlooked.

### Instructions

### DIFF
--- a/crates/cubecl-environment/src/persistence/storage.rs
+++ b/crates/cubecl-environment/src/persistence/storage.rs
@@ -344,9 +344,8 @@ pub struct NamespaceSummary {
 
 /// Every namespace the active environment holds.
 ///
-/// Empty where the backend cannot enumerate itself, which is every backend
-/// but the database: a caller sweeping what it no longer reads has nothing to
-/// sweep when the store did not outlive the process that wrote it.
+/// Empty on every backend but the database, which are the ones that do not
+/// outlive the process that wrote them.
 pub fn namespaces() -> Vec<String> {
     cfg_if::cfg_if! {
         if #[cfg(native_cache)] {
--- a/crates/cubecl-environment/src/persistence/storage.rs
+++ b/crates/cubecl-environment/src/persistence/storage.rs
@@ -342,6 +342,23 @@ pub struct NamespaceSummary {
     pub bytes: u64,
 }
 
+/// Every namespace the active environment holds.
+///
+/// Empty where the backend cannot enumerate itself, which is every backend
+/// but the database: a caller sweeping what it no longer reads has nothing to
+/// sweep when the store did not outlive the process that wrote it.
+pub fn namespaces() -> Vec<String> {
+    cfg_if::cfg_if! {
+        if #[cfg(native_cache)] {
+            super::Database::open_active()
+                .map(|database| database.namespaces())
+                .unwrap_or_default()
+        } else {
+            Vec::new()
+        }
+    }
+}
+
 /// The storage serving `namespace` in the active environment.
 ///
 /// The location is not a parameter: an environment is the store, so a cache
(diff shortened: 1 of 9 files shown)


ThierryCantin-Demers:

PR #1596, opened 2026-09-03
fix(cuda): brace the ldmatrix destination when the vector is 1 wide

michaelvsinko:
## Issue

PTX writes `ldmatrix`'s destination as a **vector operand at every width, `.x1` included**. The operand is `Vector<u32, NCD>`, and a width-1 `Vector` lowers to its scalar type (`frontend/container/vector/base.rs`, `__expand_as_type`: `if vectorization > 1`), so the placeholder substitution takes the scalar branch and emits a bare `%0` where ptxas demands `{%0}`.

```
ptxas application ptx input, line 300; error   : Vector of size 1 is expected for argument 0 of instruction 'ldmatrix'
ptxas fatal   : Ptx assembly aborted due to errors
```

Both forms in one generated kernel — `.x2` is correct, `.x1` is not:

```cuda
ldmatrix.sync.aligned.m8n8.x2.shared::cta.b16 {%0, %1}, [%2];   // fine
ldmatrix.sync.aligned.m8n8.x1.trans.shared::cta.b16 %0, [%1];   // needs {%0}
```

## Reproducer

Any f16 `mma` matmul on `sm_75` with the `m16n8k8` tile — the B fragment is 8x8, one register, so it takes `.x1`.

Surface the compile error rather than letting it be swallowed (`CUBECL_DEBUG_OPTION=debug`).

### DIFF
--- a/crates/cubecl-cpp/src/cuda/ptx/mma.rs
+++ b/crates/cubecl-cpp/src/cuda/ptx/mma.rs
@@ -97,12 +97,14 @@ fn ldmatrix(
     row: *const u32,
     #[comptime] num: usize,
     #[comptime] transpose: &str,
+    #[comptime] open: &str,
+    #[comptime] close: &str,
 ) -> Vector<u32, NCD> {
     let row_addr = generic_to_shared::<u32>(row);
 
     let out: Vector<u32, NCD>;
     gpu_asm!(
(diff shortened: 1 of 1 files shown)


wingertge (approved):
I'll approve this as a quick and dirty fix but I think the long term solution is to add format specifiers like the ones the regular `asm` uses for things like neon's register views. I'll open an issue for that later.

PR #1599, opened 2026-09-03
refactor(throughput): resolve the probe cube once, and ask for 256 units outright

ThierryCantin-Demers:
Stacked on #1594.

`LaunchConfig::cube_dim` was a `usize` that every runner pushed back through `CubeDim::new` inside its sample closure while `ops_count` counted the raw value: two derivations of one quantity, agreeing only because the constructor is idempotent on its own output. `LaunchConfig` now carries the resolved `CubeDim`, and the runners launch with it and count from it.

`launch_config` also computed the largest legal cube, 1024 units on NVIDIA, and handed it to `CubeDim::new`, which caps a cube at eight planes and answered 256. The computation read as though it chose the geometry and chose nothing. It now asks for 256 outright, which `CubeDim::new` already clamps on every device.

Wider cubes were measured before choosing, built directly to bypass the cap. RTX 4070 Ti SUPER, best of two rounds:

| units | compute-direct f32 | f16 | memory read |
|---|---|---|---|
| 128 | 46.01 TOPS/s | 48.06 | 648 GB/s |
| 256 | 46.01 | 48.06 | 663 |
| 512 | 45.77 | 48.33 | 2963 |
| 1024 | 45.41 | 48.46 | 3421 |

Arithmetic is flat across the range, so the maximal cube buys nothing. The memory rows above 256 are not a result: 663 GB/s is the bus, and a wider cube claiming five times it is a launch whose traffic is not happening. That defect sits at a geometry nothing reaches and is left alone here.

### Instructions

### DIFF
--- a/crates/cubecl-std/src/throughput/base.rs
+++ b/crates/cubecl-std/src/throughput/base.rs
@@ -24,9 +24,8 @@ use crate::throughput::{
 /// [`MemoryProbe::new`](crate::throughput::memory_probe::MemoryProbe::new)).
 const CPU_CHAIN_DEPTH: usize = 64;
 
-/// Units a GPU probe asks for. Every larger ask resolves here anyway, since
-/// [`CubeDim::new`] caps a cube at eight planes; cubes built wider than that
-/// measure no faster, and make the memory probes report impossible rates.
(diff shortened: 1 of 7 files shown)


ThierryCantin-Demers:

PR #1600, opened 2026-09-03
fix(throughput): size the memory probe's pool past the cache, not equal to its window

ThierryCantin-Demers:
Stacked on #1599.

The pool was a fixed 512 MiB and the largest window was the same 512 MiB, so the pool the sweep walked was never larger than a working set. On a card with 48 MiB of L2 enough of it stayed resident that stores never reached memory, and the write probe reported traffic the bus never carried: 747 to 751 GB/s across a 672 GB/s bus on an RTX 4070 Ti SUPER, at every working set from 32 MiB up.

The pool is now twice the largest window. Two windows is a floor a probe that cannot read the device's last level cache size can justify, not a ratio known to be enough; sizing it from that cache is still open.

Single-size probes on that card, three runs each, cache off:

| | before | after | |
|---|---|---|---|
| write | 747.8 to 751.0 | 581.0 to 583.8 GB/s | 111% of the bus to 87% |
| read | 658.7 to 661.6 | 638.3 to 638.4 | |
| copy | 600.4 to 602.1 | 591.5 to 591.7 | |

A plain CUDA fill of a 512 MiB buffer with the same grid, compiled by nvcc and timed with events, moves 606 to 619 GB/s, which is what the sweep now reports. The write curve loses the step it had at 32 MiB where it jumped from 593 to 754 and held past the bus to the end of the sweep. On an RX 5600 XT under Vulkan, which never reported above its bus, nothing moves.

Costs: each buffer doubles, so a copy probe reserves 2 GiB where it reserved 1, and a device that will not hand out a gigabyte in one allocation now measures its headline peak at half that allocation.

### Instructions

### DIFF
--- a/crates/cubecl-runtime/src/throughput/base.rs
+++ b/crates/cubecl-runtime/src/throughput/base.rs
@@ -8,9 +8,10 @@ use thiserror::Error;
 /// a probe changes what it reports.
 pub const PROBE_VERSION: u32 = 1;
 
-/// Bytes per buffer of a [`ThroughputMode::Memory`] probe left at its default
-/// working set. Clamped to the device's maximum allocation when the probe runs.
-pub const DEFAULT_BUFFER_BYTES: u64 = 512 * 1024 * 1024;
+/// Bytes one buffer of a [`ThroughputMode::Memory`] probe moves per pass at its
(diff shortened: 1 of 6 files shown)


ThierryCantin-Demers:

PR #1601, opened 2026-09-03
fix(throughput): outvote a measurement taken at a low clock

ThierryCantin-Demers:
Stacked on #1600.

**Cost first: the full `all` table takes 17.5 s where it took 9.4 s, 1.86x**, paid once per key per crate version. The gain is repeatability: on a quiet RTX 4070 Ti SUPER ten interleaved runs of compute-direct f32 read 46.015 every time where they spread 45.53 to 46.51 before.

A probe warms one kernel shape to a plateau and keeps the fastest of the draws it takes at that iteration count, all inside one window of about four hundred milliseconds. A device sitting at a low clock for that whole window is invisible from inside it: the warmup plateaus at the low rate and every draw confirms it. The minimum rejects one slow draw; it cannot reject a slow window. On an Intel Arc B390 a poisoned arithmetic probe reads 7.0 us an iteration through its warmup and all ten draws where a clean one reads 2.4 us, at the same iteration count.

`sample` now takes the whole measurement again from a cold warmup until two agree within 5%, to a cap of four, and keeps the fastest. Two agreeing measurements is the common case and the only one a quiet device pays for. Three deterministic unit tests cover it.

Not fixed here: sustained contention, which answers the same however often it is asked; and the minimum is still taken over however many draws fit the 200 ms budget, three to thirteen in practice.

### Instructions

### DIFF
--- a/crates/cubecl-runtime/src/throughput/benchmarker.rs
+++ b/crates/cubecl-runtime/src/throughput/benchmarker.rs
@@ -67,17 +67,52 @@ impl ThroughputBenchmarker {
         Ok(value)
     }
 
-    /// Warm one shape of a kernel up to its plateau, then keep its fastest sample.
+    /// The peak of one kernel shape, over measurements taken until two agree.
     pub fn sample(kernel_config: KernelConfig) -> ThroughputValue {
-        let sample = kernel_config.sample;
-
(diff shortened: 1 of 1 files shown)


ThierryCantin-Demers:

PR #1617, opened 2026-09-06
Fix allocator page-index overflow after 65,536 pages

jbg:
The allocator stores page positions in 16 bits, so a pool's 65,537th page wraps back to page zero. A new allocation can then point to memory that is still in use; in the reproducer, a 32-byte request returns the first live 8-byte buffer. This can corrupt data even though the application has kept its buffers alive and has not run out of memory.

This change uses `usize` for page and slice positions throughout allocation, splitting and compaction, and lets sliced pools use their configured capacity without truncating the page count. The regression tests keep 65,538 allocations alive, check that each has its own correctly sized storage, and verify that surviving handles still point to the same memory after cleanup and reuse. All 409 common/runtime tests pass, as do Clippy, formatting, and the build without default features. The full workspace check currently stops on an existing unused `StreamId` import in `cubecl-metal`.

### DIFF
--- a/crates/cubecl-runtime/src/memory_management/memory_pool/direct_pool.rs
+++ b/crates/cubecl-runtime/src/memory_management/memory_pool/direct_pool.rs
@@ -214,7 +214,7 @@ impl MemoryPool for DirectPool {
             }
         };
         let mut location = self.location_base;
-        location.slice = index as u32;
+        location.slice = index;
         slice.descriptor().update_location(location);
 
         let handle = slice.handle.clone();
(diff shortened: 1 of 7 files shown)


nathanielsimard:
I don't think you're supposed to have more than 65,537 pages. I would much prefer a check with a panic or error handling, since it's a pure waste to carry 64-bit indexes for stuff that should be 8–256 in size most of the time. Pages are pretty big and slices are also pretty big, so I'm curious whether you really hit those limits or if it's simply to avoid a potential vulnerability.

PR #1618, opened 2026-09-06
Fix CPU-only LLVM linking while preserving the public API

jbg:
Selecting CubeCL's CPU runtime on Apple silicon currently pulls in the AMDGPU compiler too. The bundled LLVM has the AArch64 target needed for the CPU, but lacks the AMDGPU initialization functions, so CPU programs fail at the final link step. CPU users should be able to use the supplied host LLVM without installing an AMD-capable toolchain.

This adds an `amdgpu` feature for native AMDGPU code generation and linking. It stays on by default for direct LLVM users and is enabled explicitly by HIP, while the CPU runtime stops forwarding LLVM's defaults. The public modules, options fields and enum variants remain available in either configuration, so existing struct literals and matches still compile. Requests for unavailable AMDGPU compilation or linking return a clear error; the existing unsafe printf helper, whose boolean return type cannot carry an error, panics before changing the module if a rewrite is requested without the feature. CPU execution and LLVM compatibility tests pass on the AArch64-only bundle, and both feature configurations pass Clippy. Full workspace validation still stops on the existing unused `StreamId` import in `cubecl-metal`.

### DIFF
--- a/.github/workflows/cpu-jit.yml
+++ b/.github/workflows/cpu-jit.yml
@@ -1,0 +1,26 @@
+name: CPU JIT
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+
+jobs:
+  macos-cpu:
+    # The Apple silicon LLVM bundle contains AArch64 but no AMDGPU target.
+    # Executables catch native link dependencies that cargo check misses.
+    runs-on: macos-14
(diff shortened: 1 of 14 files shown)


marcantoinem:
Closed by #1597

PR #1620, opened 2026-09-07
feat: Uniformity analysis

wingertge:
Uniformity analysis and decorations for SPIR-V

Performance impact is minimal for compute-bound scenarios but it doesn't hurt and even for attention it has ~1% speedup. It's a compiler hint so highly dependent on how much the driver can use it, I think it has a bigger impact on AMD where shared scalar registers are much more useful, but I'm not an expert on that ISA.

The analysis can be used for future cost modeling and other optimizations as well, i.e. unwrapping divergent control flow with pure ops into linear ops + select.

Also removes `pliron-llvm` from `cubecl-opt` tests by adding a new intermediate branch dialect `cf`.
Fixes bug that prevented `DeadCode` from actually running in interprocedural mode. This did not affect SCCP because that ran on the functions and not the module, but should be fixed regardless (and is needed for uniformity which runs on the module).


crates/cubecl-ir/src/dialect/cf.rs
```diff
@@ -0,0 +1,627 @@
...
+        };
+        std::println!("cond: {}", cond_attr.disp(ctx));
+        let zero = cond_attr.downcast_ref::<ZeroAttr>().map(|_| false);
+        let bool = cond_attr.downcast_ref::<BoolAttr>().map(|it| it.0);
+        let Some(const_cond) = zero.or(bool) else {
+            let num_successors = self.get_operation().deref(ctx).successors().count();
+            return (0..num_successors).collect();
+        };
+        let taken = if const_cond { 0 } else { 1 };
+        std::println!("taken: {taken}");
```
ThierryCantin-Demers: leftover debug prints?

crates/cubecl-ir/src/dialect/scf.rs
```diff
@@ -891,6 +918,22 @@ impl RegionBranchOpInterface for RangeLoopOp {
     }
 }
 
+#[op_interface_impl]
+impl UniformRegionOpInterface for RangeLoopOp {
+    fn result_uniformity(&self, _ctx: &Context, operands: &[Uniformity]) -> Uniformity {
+        operands[0]
```
ThierryCantin-Demers: Should this be `min` of start/end/step like `entry_successor_region_uniformity`
below? `operands[0]` works for If/Switch where it's the whole condition, but
here it's just `start`, so `0..unit_pos()` caps at Device and the carried value
comes out marked uniform even though the trip count varies per invocation.

crates/cubecl-ir/src/dialect/plane.rs
```diff
@@ -42,10 +79,10 @@ unary_plane_op!("plane.exclusive_i_sum", ExclusiveISumOp);
 unary_plane_op!("plane.exclusive_f_sum", ExclusiveFSumOp);
```
ThierryCantin-Demers: These are prefix scans like the prod ones, so lane N gets the running total
through N. That's non-uniform even for a uniform input, which is why the prod
macro ignores the operand. Looks like the swap just missed the sum family.

ThierryCantin-Demers (approved):

PR #1628, opened 2026-09-09
test(cpu): cover the cube boundary of the grid emulation

ThierryCantin-Demers:
## What

`test_shared_memory_is_per_cube`, in `crates/cubecl-core/src/runtime_tests/synchronization.rs`:
several cubes, each unit publishing a cube-stamped value into shared memory, a `sync_cube`, then
every unit summing every slot. A slot written by another cube stands out in the sum.

## Why

The CPU backend runs one thread per unit and walks the whole cube grid inside the kernel, so every
cube of a launch reuses the same shared-memory blocks. `InsertConstantEmulationPass` appends a cube
barrier at the end of each grid iteration for that (`fence_cubes`, `crates/cubecl-llvm/src/cpu/entrypoint.rs`).

Nothing covered it: every other shared-memory test in the suite launches a single cube, so the
fence could be deleted and the suite would stay green. Forced off, the new test fails 20 runs out
of 20, units reading slots stamped by later cubes; with it on, 20 out of 20 pass.

## Verified

- `cargo test -p cubecl-cpu --lib`: 774 passed, 0 failed.
- The new test on wgpu, Vulkan on an Intel Arc B390: passes.
- CUDA, an RTX 2060: the new test and the other cube synchronization tests pass.

Since the test joins `testgen_sync_plane!`, every runtime that instantiates the macro picks it up.

### Instructions

### DIFF
--- a/crates/cubecl-core/src/runtime_tests/synchronization.rs
+++ b/crates/cubecl-core/src/runtime_tests/synchronization.rs
@@ -350,6 +350,49 @@ pub fn test_workgroup_uniform_load_atomic_synchronizes<R: Runtime>(client: Clien
     assert_eq!(u32::from_bytes(&actual), &expected);
 }
 
+/// Spaces the cubes apart in what their units publish, so a slot left by another
+/// cube stands out in the sum.
+const CUBE_STAMP: u32 = 1000;
+
+/// Shared memory belongs to one cube: what a unit reads back must have been written
+/// by its own cube, however far ahead of it the other units have run. A runtime that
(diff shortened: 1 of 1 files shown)


ThierryCantin-Demers:
Closing: this came out of chasing a cubek top-k bug on the CPU runtime that turns out not to reproduce any more, so the test is collateral from an investigation rather than something the cubek work needs. Leaving it here for the record: the cross-cube fence in the CPU grid emulation has no test, and forcing `fence_cubes` off fails this one on 20 runs out of 20.

PR #1629, opened 2026-09-09
feat(cpu): evaluate f16 arithmetic in f32 where the host has none

ThierryCantin-Demers:
x86 without AVX512-FP16 has no f16 arithmetic, so every f16 operation paid a convert pair around an
f32 one. This evaluates f16 chains in f32 and rounds once at the end.

| `compilation.f16_evaluation` | rounds | like |
| --- | --- | --- |
| `per-operation` | after every operation | a GPU, and the old behaviour |
| `chain` | at a store, a `let mut`, or a non-arithmetic read | gcc and clang on `_Float16` |
| `accumulators` | as `chain`, and holds an f16 local in f32 where that saves converts | no C compiler |

**Default:** `per-operation` on a host with native f16 (AVX512-FP16, or FP16 on aarch64), `chain`
everywhere else, detected by the CPU runtime at device creation. Set in `cubecl.toml`:

```toml
[compilation]
f16_evaluation = "chain"
```

or overridden with `CUBECL_CPU_F16_EVAL`, like the other `CUBECL_*` settings.

> [!IMPORTANT]
> - f16 results change on hosts without native f16 for callers that set nothing.
> - A chain follows SSA values, so it runs through `let`, and a fused f16 graph can round
>   differently from the same graph unfused.
> - `per-operation` restores the old behaviour exactly. The mode is in the CPU fingerprint, so
>   caches invalidate with it.

## Numbers

Fused f16 elementwise chain through burn, GOPS/s. f32, timed alongside, stays within 3%.

| | per-operation | chain | accumulators |
| --- | --- | --- | --- |
| Xeon E5-2620 v4 | 15.9 | 32.2 (**2.02x**) | 32.5 |
| Ryzen 7 5700X | 43.9 | 81.6 (**1.86x**) | 82.4 |

f16 goes from below f32 to above it. No effect on MobileNetV3-Small, where convolutions dominate.

## Why `accumulators` is opt-in

- It goes past every C compiler, and an f16 accumulator in a kernel is a defect to fix at the
  source (tracel-ai/cubek#639).
- It decides per variable by counting converts, so a total read often inside its loop stays f16.
- It costs registers. Speedup over `chain` with `W` accumulators live, each `acc = acc * k + c` for
  200,000 iterations (8 and 16 lanes, median of 5):

| W | Ryzen | Xeon |
| --- | --- | --- |
| 4 | 5.7x | 4.5x |
| 8 | 3.5x | 3.2x |
| 16 | 1.53x | 2.06x |

## Testing

The kernels are `runtime_tests::arithmetic_chains`, float-generic, with shared tests on every runtime
using operands exact in every float type. The values below depend on the CPU's mode, so they stay in
`cubecl-cpu`: one test binary per mode, set through `CubeClRuntimeConfig`, since the mode is read once
per device, plus one that checks the default against the host's feature bit.

| kernel | `per-operation` | `chain` | `accumulators` |
| --- | --- | --- | --- |
| `a * b / c` at 300, product over the f16 maximum | inf | 300 | |
| the product bound by `let` first | | 300 | |
| the product bound by `let mut` first | | inf | 300 |
| `+= 1` from 2048, a thousand times | | 2048 | 3048 |
| the same, carried through a second local | | 2048 | 3048 |
| the same, in every lane of a vector of 4 | | | 3048 |
| the product stored under an `if`, then read twice | | | inf |

2048 is where the f16 step size passes 1. The rest of the `cubecl-cpu` suite passes. Not covered:

- The `per-operation` default on native-f16 hardware. CI is x86-64 Linux without it, and the
  aarch64 detection was only compiled standalone.
- An accumulator in a local `Array<f16>`.

### Instructions

### DIFF
--- a/crates/cubecl-cpu/src/runtime.rs
+++ b/crates/cubecl-cpu/src/runtime.rs
@@ -11,17 +11,38 @@ use cubecl_core::{
     server::ServerUtilities,
     zspace::{Shape, Strides},
 };
-use cubecl_llvm::PlironCompiler;
+use cubecl_llvm::{F16Evaluation, PlironCompiler};
 use cubecl_runtime::runtime::Runtime;
 use cubecl_runtime::{allocator::ContiguousMemoryLayoutPolicy, logging::ServerLogger};
 use cubecl_std::tensor::is_contiguous;
 use std::sync::Arc;
 use sysinfo::{CpuRefreshKind, System};
 
-#[derive(Default)]
 pub struct RuntimeOptions {
     /// Configures the memory management.
     pub memory_config: MemoryConfiguration,
+    /// How wide f16 intermediates are held, from `CUBECL_CPU_F16_EVAL`.
+    pub f16_evaluation: F16Evaluation,
+}
+
+impl Default for RuntimeOptions {
+    fn default() -> Self {
+        Self {
+            memory_config: MemoryConfiguration::default(),
+            f16_evaluation: env_f16_evaluation().unwrap_or_default(),
+        }
+    }
+}
+
+/// `None` where the variable is unset or unreadable, so a value nobody recognizes leaves the
+/// default alone rather than failing a launch.
+fn env_f16_evaluation() -> Option<F16Evaluation> {
+    match std::env::var("CUBECL_CPU_F16_EVAL").ok()?.as_str() {
+        "per-operation" => Some(F16Evaluation::PerOperation),
+        "chain" => Some(F16Evaluation::Chain),
+        "accumulators" => Some(F16Evaluation::Accumulators),
+        _ => None,
+    }
 }
 
 #[derive(Debug, Clone)]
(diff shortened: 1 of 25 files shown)


crates/cubecl-cpu/src/runtime.rs
```diff
@@ -11,17 +11,40 @@ use cubecl_core::{
...
+        Self {
+            memory_config: MemoryConfiguration::default(),
+            f16_evaluation: env_f16_evaluation(),
+        }
+    }
+}
+
+/// `None` where the variable is unset or unreadable, so a value nobody recognizes leaves the
+/// host's choice alone rather than failing a launch.
+fn env_f16_evaluation() -> Option<F16Evaluation> {
```
marcantoinem: Don't put in an environment variable, put it in the cubecl toml config

crates/cubecl-cpu/tests/common/mod.rs
```diff
@@ -0,0 +1,243 @@
+//! The kernels the f16 evaluation modes are judged on.
+//!
+//! Each mode is read once, when the runtime builds its device, so each has its own test binary
+//! and they share this rather than a process.
+
+#![allow(dead_code)]
```
marcantoinem: Put this in cubecl general tests not in cubecl-cpu

crates/cubecl-llvm/src/shared/to_llvm/math.rs
```diff
@@ -339,9 +339,14 @@ impl ToLLVMDialect for BoolNotOp {
 /// Contraction only rounds once instead of twice, and a matmul inner loop is made of it.
 ///
 /// On the GPU only. The FMA is the instruction the hardware wants, and the C++ backends
-/// contract by default, so this is what the other runtimes already do. The CPU pipeline is
-/// the reference those runtimes get compared against, so its arithmetic stays exactly what
-/// the kernel wrote.
+/// contract by default, so this is what the other runtimes already do. The CPU declines it so
```
marcantoinem: Please simplify this comment

marcantoinem (commented):
Please remove f16 test that could just be shared with other runtime

marcantoinem (approved):
LGTM

PR #1633, opened 2026-09-10
fix(runtime): disambiguate kernel entrypoint names and avoid launch-path allocation (#1625)

### LINKED ISSUE #1625: Kernel names do not identify the kernel: comptime arguments are absent from both names
A kernel's name renders its generic parameters but not its comptime arguments, so
distinct kernels share one name and anything keyed on a name rather than on
`KernelId` conflates them. Observed at 5e39cce2.

cubek's reduce kernel takes its strategy and its algorithm as comptime values, so
ArgMax, ArgMin and ArgTopK at the same dtypes and vector widths are one generic
instantiation and one name,
`reduce_kernel_in_f32_in_size_16_out_u32_out_size_1_acc_f32`. Two consumers then
conflate them, through two different names.

## The entrypoint name: the IR dump overwrites kernels

`KernelSettings::kernel_name`, built by `kernel_entrypoint_name` in
`crates/cubecl-macros/src/generate/kernel.rs` from the generics alone. It is the
emitted entry symbol, so nsys shows one name for several kernels, and the pliron IR
dump writes them all to one directory:

```bash
CUBECL_DEBUG_PLIRON=/tmp/ir cargo test -p cubek-reduce --test lib \
  --features cubecl/cpu,cubecl/pliron-dump
# run twice, diff /tmp/ir: one revision, 208 vs 529 line llvm.opt.ll under one name
```

The survivor is whichever thread wrote last. `--test-threads=1` makes it
deterministic and still keeps only one kernel per name. Four call sites key on it:
`kernel_dir_name` (`cubecl-cpp/src/shared/base.rs:412`, from `base.rs:221` and
`cubecl-wgpu/src/compiler/wgsl/compiler.rs:102`), `ir_dump_path`
(`cubecl-llvm/src/cpu/jit/engine.rs:142`) and `pliron_path`
(`cubecl-llvm/src/shared/base.rs:383`).

## The debug name: profiling labels cannot tell them apart

`KernelMetadata::name()` (`cubecl-runtime/src/kernel.rs:27`), never overridden by
the macro and so always `type_name` of the generated struct. Profiling labels use it
(`cubecl-runtime/src/client.rs:1034`, `:1047`, `:1057`), which is why durations
permute across strategy slots between runs. `ProfileLevel::Full` prints the
`KernelId` too (`client.rs:1160`) and does distinguish them; `Basic` and `Medium`
do not.

## Nothing appears to depend on the current names

The compilation cache is keyed on `KernelId::stable_hash()`
(`cubecl-runtime/src/compiler.rs:74`), autotune keys are caller-supplied strings,
and `LaunchObserver` has no implementor outside cubecl's own tests. The constraint
is cost rather than compatibility: `client.rs:1034` calls `kernel.name()` on every
launch ahead of the observer's `OBSERVING` check, so a `name()` returning `String`
would allocate on every launch.

## Two directions, not exclusive

A short discriminator from `KernelId::stable_hash()` stamped on the name in
`CompiledKernel::compile`, which holds both the kernel and the definition, would
make the symbol and all four dump paths unique at once, but says two kernels differ
without saying how.

Rendering the comptime values into the name is the only thing that tells ArgMax from
ArgMin at a glance, and needs `#[cube(launch)]` to build a name from the comptime
params plus an answer for that per-launch allocation.

Sadik00789:
## Description

Resolves #1625 where kernels with identical generic signatures but differing comptime properties collide on entrypoint names, file dump directories (`CUBECL_DEBUG_PLIRON`), and profiling labels.

### Changes

1. **Upstream Entrypoint Disambiguation**: Added `KernelId::entrypoint_name(&self, base: &str)` in `cubecl-runtime`. In `cubecl-macros` (`define_body()`), suffixed the kernel name directly during `define()` via `KernelMetadata::id(self).entrypoint_name(&self.settings.kernel_name)`. This guarantees that both `KernelSettings` and the IR entry function symbol agree upon construction across all backends (LLVM, WGPU, CUDA) without mutating definitions downstream.
2. **Unified Profile Label Formatting**: Extracted a shared `profile_label` helper in `cubecl-runtime/src/client.rs` that attaches the stable discriminator suffix across all four registration sites (`ExecutionOnly` across non-profiled and profiled paths, as well as `Basic` and `Medium`).
3. **Zero-Cost Launch Hook**: Guarded `notify_launch(kernel.name())` with `is_observing()` (relaxed atomic check) to prevent unconditional dynamic name formatting and allocations when observers are inactive.
4. **Clean Workspace Dependencies & Regression Test**: Relocated the regression test to `crates/cubecl-core/src/runtime_tests/launch.rs` under the shared `testgen_launch!` suite (next to `test_kernel_with_comptime_tag`), verifying across backends that kernels differing only in comptime values generate distinct `kernel_name`s upon `.define()`. This avoids introducing a circular `cubecl-core` dev-dependency into `cubecl-runtime`.

## Validation

* `cargo test -p cubecl-runtime` (all 288 unit and integration tests pass)
* `cargo test -p cubecl-core`
* `cargo test -p cubecl-cpu -- test_launch_with_comptime_name_disambiguation`
* `cargo test -p cubecl-cpu -- test_select_true`
* `cargo test -p cubecl-wgpu --test graph`
* `cargo clippy --all-targets -- -D warnings`
* `cargo fmt --all --check`


crates/cubecl-runtime/src/kernel.rs
```diff
@@ -179,10 +179,15 @@ impl<C: Compiler> CompiledKernel<C> {
...
-        definition: KernelDefinition,
+        mut definition: KernelDefinition,
         compiler: &mut C,
         compilation_options: &C::CompilationOptions,
     ) -> Result<Self, CompilationError> {
+        let disc = (kernel.id().stable_hash() & 0xFFFF_FFFF) as u32;
+        if !definition.settings.kernel_name.is_empty() {
+            definition.settings.kernel_name =
+                alloc::format!("{}_{:08x}", definition.settings.kernel_name, disc);
+        }
```
ThierryCantin-Demers: The IR entry function was named from the settings before this runs (`cubecl-ir/src/scope.rs:318`), and every non-LLVM backend prints that name, so the runtime looks up a symbol the module doesn't declare.

Rather than renaming here, could the name be set where the kernel is built? The `define()` generated by `#[cube(launch)]` passes its settings to `KernelBuilder::new`, which names both the settings and the IR entry function, so suffixing the name there makes them agree from the start. In `define_body()` in `crates/cubecl-macros/src/generate/kernel.rs`:

```diff
     let kernel_builder = prelude_type("KernelBuilder");
+    let kernel_metadata = prelude_type("KernelMetadata");
     ...
     quote! {
-        let mut builder = #kernel_builder::new(self.settings.clone());
+        let __name = <Self as #kernel_metadata>::id(self).entrypoint_name(&self.settings.kernel_name);
+        let mut builder = #kernel_builder::new(self.settings.clone().kernel_name(__name));
```

with this on `KernelId` in `crates/cubecl-runtime/src/id.rs`:

```rust
pub fn entrypoint_name(&self, base: &str) -> String {
    format!("{base}_{:08x}", self.stable_hash() as u32)
}
```

`compile` can then go back to taking `definition` unchanged, and neither `update_entry_name` nor the two renames in `cubecl-llvm/src/shared/base.rs` are needed. I tried this locally: the wgpu graph tests and cubek-reduce on wgpu and cpu pass.

crates/cubecl-runtime/tests/kernel_discriminator.rs
```diff
@@ -0,0 +1,85 @@
...
+impl Compiler for TestCompiler {
+    type Representation = String;
+    type CompilationOptions = ();
+
+    fn compile(
+        &mut self,
+        _kernel: KernelDefinition,
+        _options: &Self::CompilationOptions,
+    ) -> Result<Self::Representation, CompilationError> {
+        Ok("void main() {}".to_string())
```
ThierryCantin-Demers: This stub ignores the definition, so the test passes while every wgpu launch fails. A test on a real `#[cube(launch)]` kernel would cover it: two instances that differ only in a comptime argument should get different `kernel_name`s from `define()`.

crates/cubecl-runtime/src/client.rs
```diff
@@ -1162,7 +1174,14 @@ impl Client {
                             ProfileLevel::Full => {
                                 format!("{name}: {kernel_id} CubeCount {count:?}")
                             }
-                            _ => type_name_format(name, TypeNameFormatLevel::Balanced),
+                            _ => {
```
ThierryCantin-Demers: The three `ExecutionOnly` registrations on the profiled path (lines 1127, 1137, 1170) still log the bare name, so those launches can't be told apart in the profile. Building the label in one helper would keep all four in sync.

ThierryCantin-Demers (changes requested):
Thanks for taking this on! One main thing to change before it can go in: it currently breaks kernel launches on every non-LLVM backend. `cargo test -p cubecl-cuda` fails with `named symbol not found`, and `cargo test -p cubecl-wgpu --test graph` with `Unable to find entry point`. The cause and a suggested fix are in the comment on `kernel.rs`, plus a couple of smaller points inline.

Sadik00789:
Thanks for the thorough review and catch! Moving the suffixing logic upstream into KernelId::entrypoint_name and generating it directly in define_body() makes total sense. It keeps the runtime, IR, and backend symbol lookups cleanly in sync from the start while avoiding backend-specific lookup panics. I've updated the implementation accordingly, unified the profiling label formatting, and verified everything with the test suite and linter.

ThierryCantin-Demers:
Thanks for the update, the move to `define()` looks good! One thing: this makes cubecl-runtime and cubecl-core depend on each other, which the workspace doesn't do anywhere else. Could the test go in `cubecl-core/src/runtime_tests/launch.rs` instead, next to `test_kernel_with_comptime_tag`? It would get the real `client` those tests receive, so it runs on every backend and needs neither this dev-dependency nor the dummy client.

Sadik00789:
Good call on keeping the workspace dependencies clean. I've removed cubecl-core from cubecl-runtime/Cargo.toml along with the standalone test file, and moved the regression test into cubecl-core/src/runtime_tests/launch.rs alongside test_kernel_with_comptime_tag. It runs under the shared testgen_launch! harness across backends without needing dummy clients or circular dev-dependencies. All tests, formatting, and clippy checks are passing.

louisfd (approved):

PR #1639, opened 2026-09-11
fix(core): warn on constant shared barrier elections

### LINKED ISSUE #1610: [cubecl 0.10 / CUDA] Miscompile: σ-shaped expression in a uniform unrolled loop miscompiles when a lane-divergent predicated tail and a cross-lane barrier reduce coexist in the kernel
## Environment

- cubecl 0.10.0 / cubecl-core 0.10.0 / cubecl-cuda 0.10.0, `#[cube(launch_unchecked)]` kernels
- GPU: NVIDIA GeForce RTX 3090 (SM86)
- Host: Linux x86_64, Rust 1.95.0

## Summary

A kernel miscompiles **only** when all three of the following coexist:

1. **a σ-shaped expression — a reciprocal whose DIVISOR CONTAINS an exp result, i.e. `1/(1+exp(-x))` — evaluated inside a comptime-unrolled, fully-uniform strided loop**, and
2. **a lane-divergent runtime-predicated block anywhere in the kernel** (our tails: `if t < d { … }`, active for a strict subset of lanes when `d % lanes != 0`), and
3. **a shared-memory + `Barrier` cross-lane reduction** over the loop accumulator.

Remove any ONE ingredient and the kernel is bit-exact. The failing outputs are deterministic and grossly wrong (max-abs up to ~436 on our data), consistent with register/predicate-state clobbering during the σ lowering (we suspect a fast-math rewrite of the `1/(1+exp(-x))` pattern).

## The 14-variant differential ladder

| Variant | exp | div | exp feeds the divisor (σ shape) | location | predicated tail | smem reduce | diff | |
|---|---|---|---|---|---|---|---|---|
| A | – | – | – | – | yes | no | 0 | ✅ |
| B | – | – | – | – | yes | yes | 0 | ✅ |
| C | – | – | – | – | yes (read+write) | yes | 0 | ✅ |
| I | – | yes | – | post-reduce | yes | yes | 0 | ✅ |
| F | yes | – | – | divergent tail | yes | no | 3e-8 | ✅ |
| G | yes | – | – | divergent tail | yes | yes | 0 | ✅ |
| K | yes | – | – | uniform loop | yes | **no** | 3e-8 | ✅ |
| L | yes | – | – | uniform loop | yes | yes | 0 | ✅ |
| M | – | yes | – | uniform loop | yes | yes | 0 | ✅ |
| N | yes | yes | **no** (independent operands `e/(1+x)`) | uniform loop | yes | yes | 0 | ✅ |
| **J** | yes | yes | **yes** (`1/(1+exp(-x))`) | uniform loop | yes | yes | **436.2** | ❌ |
| H | yes | yes | yes | uniform loop | yes (raw adds) | yes (+sqrt/div pass-2) | **2.94** | ❌ |
| D (production kernel) | yes | yes | yes | uniform loop + tails | yes (σ in tail) | yes (+sqrt/div pass-2) | **3.09** | ❌ |
| E (workaround) | yes | yes | yes | uniform loop over zero-padded width | **no** | yes (+sqrt/div) | 0 | ✅ |

Key contrasts:
- **L vs J**: identical skeleton; L's uniform loop computes plain `exp(x)`, J's computes `1/(1+exp(-x))`. Only J fails → the σ shape (exp result feeding the divisor), not exp or div individually, is required.
- **N vs J**: N has exp AND div in the same uniform loop but on independent operands (`e/(1+x)`); passes → it is not "exp+div anywhere in the loop" but specifically the divisor consuming the exp result.
- **K vs J**: K = σ-loop's exp alone without the reduce; passes → the cross-lane reduce is required.
- **E vs J**: E evaluates the same σ over a zero-padded width with the divergent tail eliminated; passes → the predicated block is required.

## Minimal failing kernel (Variant J — 3 array args + 2 comptime)

```rust
#[cube(launch_unchecked)]
fn exp_uniform_add_out_kernel<F: Float>(
    o: &Array<F>, g: &Array<F>, out: &mut Array<F>,
    #[comptime] d: usize, #[comptime] lanes: usize,
) {
    let lane = ABSOLUTE_POS % lanes;
    let row = ABSOLUTE_POS / lanes;
    let base = row * d;
    let full = d / lanes;                       // 1 at d=96, lanes=64

    let mut partial = SharedMemory::<F>::new(lanes);
    let barrier = Barrier::shared(CUBE_DIM, true);

    let mut acc = F::new(0.);
    for k in 0..full {                          // uniform, unrolled
        let idx = lane + k * lanes;
        let sig = F::new(1.) / (F::new(1.) + (g[base + idx] * F::new(-1.)).exp());  // σ
        let v = o[base + idx] * sig;
        acc += v * v;
    }
    let t = lane + full * lanes;                // divergent for lanes ≥ d % lanes
    if t < d {
        acc += o[base + t];                     // plain add — no exp, no div
    }
    partial[lane] = acc;
    barrier.arrive_and_wait();

    let mut total = F::new(0.);
    for j in 0..lanes { total += partial[j]; }

    for k in 0..full {
        let idx = lane + k * lanes;
        let sig = F::new(1.) / (F::new(1.) + (g[base + idx] * F::new(-1.)).exp());
        out[base + idx] = o[base + idx] * sig + total;
    }
    let t2 = lane + full * lanes;
    if t2 < d {
        out[base + t2] = o[base + t2] + total;
    }
}
```

Launch shape and data are in the test file; the full ladder (A–N) lives alongside it with each variant independently togglable, plus host references for every variant.

## Workaround (verified bit-exact)

Zero-pad the row width to a lane multiple (`d_pad = ceil(d/lanes)·lanes`; zero-padded inputs contribute exactly 0) and drop every runtime predicate. This is variant E — passes on all shapes we tested (`d ∈ {8, 96, 97, 128}`, `rows ∈ {8, 33, 64, 256}`), and is what our production kernel ships.

## Expectation

Variant J (and H/D) should match the host reference (diff < 1e-4) on all shapes, as the other twelve variants do.

## Notes for maintainers

- Earlier hypotheses from the same investigation — "runtime predicates alone", "any transcendental under predication", "exp+div anywhere" — were each disproven by a passing variant (A/B/C; F/G/K; L/M/N respectively); the ladder keeps those rungs so the boundaries are reproducible.
- `select(active, x, y)` on lane-divergent conditions did not rescue σ-style kernels in our earlier experiments, and clamped-index read-modify-write introduces cross-lane write races (user error, not cubecl) — zero-padding was the only fully reliable form we found.
- We could not reproduce without the `Barrier` reduce, so the barrier itself (or the shared memory it guards) appears to be part of the miscompiled interaction, not incidental.

arcusbuilds:
## Summary

`Barrier::shared(..., true)` elects every GPU unit, so they all initialize the same barrier, which is undefined behavior; `false` leaves it uninitialized.

Following the review, this is now a warning rather than a compile error:

- `Scope` gains a non-fatal `push_warning`/`pop_warnings` channel next to the existing `push_error`/`pop_errors`.
- `KernelBuilder::build` drains it once and emits each entry through `log::warn!`, prefixed with the kernel name, so every backend reports it without changes.
- `Barrier::shared` pushes the warning for a constant election and recommends `UNIT_POS == 0`.
- Unit tests assert the warning fires for a constant election and stays quiet for a runtime one.

The warning goes through `log`, so it shows up with any installed `log` subscriber at warn level and is silenced by ordinary log filtering. It never blocks compilation, so I did not add a config knob; happy to add one under `CompilationConfig` if you would rather have an explicit switch.

## Validation

- [x] `cargo +1.95.0 xtask validate` (audit, format, typos, lint, build, unit tests; the `cubecl-wgpu` unit-test binary intermittently dies with `STATUS_ACCESS_VIOLATION` on my Windows box, including on untouched `main`, so that one is a local driver issue)
- [x] `cargo +1.95.0 test -p cubecl-core --lib frontend::barrier`

Closes #1610


wingertge (commented):
I don't think this should be a compile error just because it's misuse of the API, I think there needs to be a way to emit warnings for these things of things, or at least a way to disable the check. This is closer to the `arithmetic_overflow` lint than an actual compile error.

arcusbuilds:
@wingertge I have updated the code, please review!

PR #1643, opened 2026-09-14
Refactor/async turso storage

jwric:
Changes the SQLite backend from `rusqlite` to `turso`. And unifies the store API for both web and native implementations.

## Validate your PR with burn.

It is important that you make sure that you don't introduce any bugs in burn. 

### Instructions

- [ ] Create a new branch or fork of the [cubek repo](https://github.com/Tracel-AI/cubek)
- [ ] Update the main [Cargo.toml](https://github.com/tracel-ai/cubek/blob/main/Cargo.toml#L22=L23) with this PR hash.
- [ ] Fix any broken tests or compilation errors in cubek.
- [ ] Submit a PR in cubek with your fixes and link it here.
- [ ] Create a new branch or fork of the [burn repo](https://github.com/Tracel-AI/burn)
- [ ] Update the main [Cargo.toml](https://github.com/tracel-ai/burn/blob/main/Cargo.toml#L223=L226) with this PR and the cubek PR hash.
- [ ] Fix any broken tests or compilation errors in burn.
- [ ] Submit a PR in burn with your fixes and link it here.

Issue #1608, opened 2026-09-04
Replace rusqlite with Turso in cubecl-environment's persistence layer


antimora:
### Feature description

Replace `rusqlite` with [Turso](https://github.com/tursodatabase/turso) as the engine behind `cubecl-environment`'s persistence layer, so `cubecl-runtime` stops pulling `libsqlite3-sys` into every downstream graph.

This is the implementation side of #1488, which covers the problem (the `links = "sqlite3"` conflict that stops CubeCL coexisting with `sqlx`). Filing separately so the compatibility work has somewhere to live.

### Feature motivation

`cubecl-runtime` enables `cubecl-environment/cache` unconditionally on desktop targets, so `rusqlite` and its C amalgamation reach every dependent. Turso is a pure Rust engine that reads and writes the same file format, which removes the C toolchain requirement and the `links` conflict at once.

We just did this migration in burn for `burn-dataset` (tracel-ai/burn#5546), so the following is measured against `turso 0.8.0-pre.7` rather than assumed.

### What already works

Tested against the operations `cubecl-environment` actually performs, line refs at cubecl rev `2f9b71f`:

| what the cache does | Turso |
|---|---|
| `journal_mode = WAL` (`persistence/sqlite.rs:189`) | native, and the only mode |
| `synchronous = NORMAL` (`persistence/sqlite.rs:190`) | accepted |
| `busy_timeout(5s)` (`persistence/sqlite.rs:187`) | `Connection::busy_timeout` |
| `TransactionBehavior::Immediate` | `Connection::transaction_with_behavior` |
| `ON CONFLICT ... DO UPDATE` upserts | works |
| `ATTACH DATABASE` (`bundle/export.rs:254`) | works behind `Builder::experimental_attach(true)`; verified a cross-database copy with a conflict correctly skipped |

### The two things that need design work

**`journal_mode = DELETE` in `finalize_for_shipping()`** (`persistence/sqlite.rs:335`). Turso is WAL-only and reports success while changing nothing:

```
set journal_mode=DELETE -> returned 1 row
journal_mode now        -> "wal"
```

`PRAGMA wal_checkpoint(TRUNCATE)` does fold the WAL into the main file, so a shipped bundle can still be a single file stock SQLite reads (verified with `pragma integrity_check`), but the header stays WAL-mode (`0202`) rather than legacy (`0101`). If bundles must ship as rollback-journal files, that needs another answer.

[...]

Issue #1488, opened 2026-08-08
cubecl-environment's sqlite cache forces libsqlite3-sys on all downstream users (links conflict with sqlx)


TsaoLun:
## Problem

`cubecl-runtime` enables `cubecl-environment/cache` unconditionally on desktop
targets, which pulls in `rusqlite` → `libsqlite3-sys ^0.38.1`. Since
`libsqlite3-sys` has `links = "sqlite3"`, only one version can exist per
dependency graph — so CubeCL cannot coexist with `sqlx`, whose SQLite driver
wants `^0.30.1`.

## Reproduction

```toml
[dependencies]
cubecl-runtime = { git = "https://github.com/tracel-ai/cubecl", rev = "28c69a98c3dde15012a5192c0e22b56dbd447a66" }
sqlx = { version = "0.8", default-features = false, features = ["runtime-tokio", "postgres", "migrate"] }
```

`cargo generate-lockfile`:

```
error: failed to select a version for `libsqlite3-sys`.
package `libsqlite3-sys` links to the native library `sqlite3`, but it conflicts with a previous package which links to `sqlite3` as well:
package `libsqlite3-sys v0.38.0`
    ... of package `rusqlite v0.40.0`
    ... of package `cubecl-environment v0.11.0-pre.1
    ... of package `cubecl-runtime v0.11.0-pre.1
```

Note that `sqlx`'s SQLite driver is **never enabled** here — only `postgres`.

## Why a feature flag won't fix this

The conflict happens during resolution, not compilation. Cargo still picks a
version for an optional dependency whenever a weak feature reference names it
(`sqlx-sqlite?/migrate` is what drags it in above), and the `links` check fires
regardless of activation. So as long as `rusqlite` appears in
`cubecl-environment`'s manifest, downstream graphs are pinned to
`libsqlite3-sys 0.38`.

Version bumps don't bridge it either: `sqlx-sqlite 0.9.0` allows
`>=0.30.1, <0.38.0`, while `rusqlite 0.40.1` needs `^0.38.1`.

## Impact

`burn-std` depends on `cubecl-environment` unconditionally, so **every Burn user
on a desktop target** inherits `libsqlite3-sys 0.38` — even when using only the
pure-Rust CPU backend with no autotune cache.

## Suggestion

Move the SQLite backend into its own crate (e.g. `cubecl-cache-sqlite`) so
`cubecl-environment`'s manifest no longer mentions `rusqlite`. Alternatively, a
[...]

PR #1645, opened 2026-09-14
refactor(bundle): drop the SQLite bundle format

jwric:
## What

A bundle is now always the flat layout: one blob every target reads, whether from `include_bytes!`, a fetch, or a file on disk. This removes:

- `BundleFormat` and the `format` field of `ExportOptions`
- `SqliteBundle` and `bundle/sqlite.rs`
- the `ATTACH DATABASE`-based export path (`copy_entries`/`copy_attached`) and the `-wal`/`-shm` sidecar handling around staging/publishing
- `BundleManifest::read`/`write` against a database (the manifest lives in the blob)
- `Database::finalize_for_shipping`, which rolled a bundle's journal back to `DELETE` so it could be read from a read-only install — the flat blob never had that problem
- `Bundle::save` takes just the output path

`environment::load` is unchanged: it still mounts an environment database in place, it just no longer doubles as a way to use a bundle without importing it. `bundle::open` reads the flat header only.

## Why

Groundwork for replacing rusqlite with Turso (#1608, #1488). The SQLite bundle format was the one consumer of `journal_mode = DELETE` and `ATTACH DATABASE`, the two operations #1608 flags as needing design work on a WAL-only engine. Removing the format first means the engine swap doesn't have to answer those questions, and the swap PR shrinks by ~400 lines.

The flat format already existed and already covered every target; the SQLite one was only ever the native default.

## Breaking

Public API: `BundleFormat`, `SqliteBundle`, `ExportOptions::format`, `Bundle::save(out, format)` → `Bundle::save(out)`. Existing `.bundle` files written in the SQLite format won't open; re-export them.

## Tests

`cargo test -p cubecl-environment --features cache` — the bundle integration tests are reduced to the flat path (the `both_formats_serve_the_same_entries` cross-check goes; `a_saved_bundle_can_be_loaded_in_place` becomes `an_environment_database_can_be_loaded_in_place`).

Stacked under: #1643.


jwric:
Closing: the premise was wrong. I'd assumed a WAL-only engine couldn't serve a database-format bundle from a read-only install, since SQLite needs to create the `-shm` wal-index next to the file. Turso keeps its wal-index in memory: a checkpointed (`wal_checkpoint(TRUNCATE)`) file opens with `read_only(true)` from a `0o555` directory, `0o444` file, no sidecars created. So there's no reason to give up the format — #1643 will port the database bundle to Turso instead of removing it.

PR #1656, opened 2026-09-16
feat: MemorySSA

wingertge:
Core infrastructure for handling memory in PRE/CSE/LICM. Only used in `SimpleCSE` since that's the only pass that exists right now that benefits from it.

### Instructions

- [x] Submit a PR in cubek with your fixes and link it here.


crates/cubecl-opt/src/analyses/memory_ssa.rs
```diff
@@ -0,0 +1,949 @@
...
+                        continue;
+                    }
+                    ExploredResult::Clobber => {
+                        return TraversalResult::Value(current_value);
+                    }
+                    ExploredResult::Cycle => {
+                        return TraversalResult::Cycle;
+                    }
+                }
+            }
```
marcantoinem, line 811: If two load are done on the same data and their is thread synchronization between, the two load will be considered identical by this analysis which is wrong and could to bad optimization. For example:
```
before = load(shared_data)
wait_parity(ready_barrier, phase)
after = load(shared_data)
```
would become the invalid form
```
before = load(shared_data)
wait_parity(ready_barrier, phase)
after = before
```

crates/cubecl-opt/src/analyses/memory_ssa.rs
```diff
@@ -0,0 +1,949 @@
...
+        merge_points: &mut ISet<Ptr<BasicBlock>>,
+    ) {
+        if region.is_empty(ctx) {
+            return;
+        }
+
+        let dom_tree = self.dom_info.get_dom_tree(ctx, region);
+        let frontiers = DomFrontierMap::new(ctx, &region, dom_tree);
+
+        for block in defining_blocks.iter() {
```
marcantoinem: Phi placement needs the **iterated dominance frontier**, rather than just the frontier of the original write blocks. For a lowered nested `if`:

```text
entry → outer / exit
outer → write / merge
write → merge
merge → exit
```

With a write only in `write`, this inserts a phi at `merge` but misses the required phi at `exit`. A load in `exit` consequently gets `LiveOnEntry`, ignoring the conditional write.

marcantoinem (dismissed):
There seems to be some problem

crates/cubecl-opt/src/analyses/memory_ssa.rs
wingertge: I'll double check barriers are not listed as memory effect free, they should be opaque until I can properly reason about the clobbering semantics for each barrier

wingertge: Barriers do not have the `NoMemoryEffect` trait so they're currently `Opaque` clobber-alls. So this is fine. I intentionally didn't add that trait when I first added the effects because they do have memory effects that are hard to describe (and as usual, a lack of the interface means the op is unsupported and must be treated conservatively).

crates/cubecl-opt/src/analyses/memory_ssa.rs
wingertge: I assumed `DomInfo` did that because MLIR's IDFCalculator does, but it seems I was wrong. This also affects `mem2reg` then, I'll figure out the best way to implement that. Maybe a new wrapper because I don't want to keep reimplementing it.

crates/cubecl-opt/src/analyses/memory_ssa.rs
marcantoinem: Okay that's fine

marcantoinem (approved):
LGTM

PR #1659, opened 2026-09-16
fix(throughput): probe memory outside the installed pools, and fail when it cannot be allocated

SamuelBelanger:
## Problem

1. The memory throughput probes allocate 1 GiB buffers in the dynamic pools. Under a custom `install_memory_pools` layout no pool accepts that, so the reservation fails with `BufferTooBig` on a device with memory free.
2. A failed reservation is invisible. `initialize_memory` panics, the device thread catches it, and the handle is left empty. `check`, `sync_buffers` and `sync` all answer `Ok`, and the probe launches over nothing and returns a timing (reproduced on an RTX 2060 with 5 of 6 GiB held). `measure_peak_throughput` caches any finite result, so that timing would be persisted as the device's peak.

## Changes

- **Probes allocate in a persistent window.** Exact-fit and uncapped whatever the layout, and outside the workload's budget. Buffer sizes and cache keys are unchanged. `PooledProbes` already runs the explicit cleanup that frees them.
- **A buffer whose reservation failed now fails `check`.** `ErrorGraph::reports` takes a `Claim` (`Failed` or `Unallocated`), and `ManagedMemoryDescriptor::is_allocated` is the single read of the init flag. This is on every backend's read path.
- **Probes return `ThroughputError::Allocation`** instead of timing an empty pass. Not cached, and `autotune_bound` already treats a probe error as no ceiling.
- **Memory probes confirm their kernel ran.** One untimed pass at build time, then `sync_buffers` on the output: a launch failure lives on the buffers it never wrote, and `sync` answers `Ok` regardless. Returns `ThroughputError::Launch`. First pass only, and the compute probes are not covered.

The first two commits (`max_servable`) are superseded by the third and their code is gone.

## Known limits

- The new error names the buffer, not the cause, which is only in the device log. Carrying it needs a placeholder slice for failed reservations: follow-up PR.

## Testing

RTX 2060, CUDA:

- 64 MiB capped layout: all three probes run, dynamic page peak stays 0.
- Full card: probes return `allocation failed`, then run again once memory is released.
- `verify` on the `bf16` fma kernel, which does not compile on this card: `Err(Launch)`. The f32 one passes.
- Reserved memory around a probe: 0, 2048, 2048 after drop, 0 MiB after `memory_cleanup`.
- `cargo test -p cubecl-cuda`: 950 passed, 23 failed (`bf16` kernels that do not compile, `complex_math` unexpanded functions). None involves allocation. Not compared against `main`.

New tests: `persistent_window_serves_what_a_capped_layout_refuses`, `a_handle_whose_reservation_failed_is_refused`, `a_shape_that_cannot_be_built_fails_the_sweep`. Clippy is clean on `cubecl-std`, `cubecl-server`, `cubecl-runtime`, `cubecl-cuda`. HIP, Metal and wgpu not run.


crates/cubecl-std/src/throughput/runners/memory_probe.rs
```diff
@@ -95,7 +122,7 @@ impl MemoryProbe {
     pub fn new(client: &Client, config: LaunchConfig, line_bytes: usize, spec: MemorySpec) -> Self {
         let blocked = config.plane_size == 1;
         let shape = DeviceShape {
-            max_alloc: client.properties().memory.max_page_size as usize,
+            max_alloc: servable_alloc(client) as usize,
```
ThierryCantin-Demers: `servable_alloc` only caps the size of one buffer. The copy probe keeps two
`buffer_bytes` allocations live, so a hard-capped layout with one 512 MiB pool
still accepts the first buffer but rejects the second with
`PoolCapacityExceeded`, even if the device has free memory. Please account for
the total capacity required by all buffers in this probe, not only the largest
single accepted allocation.

ThierryCantin-Demers (approved):

PR #1667, opened 2026-09-18
fix(llvm): build the bitcode linker shim for NVPTX too

ThierryCantin-Demers:
Since #1609, a build that links the CUDA runtime without the ROCm one fails at link time:

```
rust-lld: error: undefined symbol: cubecl_link_device_bitcode
>>> referenced by libdevice.rs:68 (crates/cubecl-llvm/src/nvptx/libdevice.rs:68)
rust-lld: error: undefined symbol: cubecl_free_message
>>> referenced by libdevice.rs:73 (crates/cubecl-llvm/src/nvptx/libdevice.rs:73)
```

The NVPTX backend links libdevice through two functions of the C++ shim in `src/amdgpu/cpp_shims/device_libs.cpp`, but `build.rs` only compiles that shim with the `amdgpu` feature, and `cubecl-cuda` enables `cubecl-llvm` with `nvptx` alone. A workspace build hides it, since `cubecl-hip` turns `amdgpu` on for every crate in the build. Downstream, burn's `cuda` feature without `rocm` would hit it on the next cubecl update. This PR builds the bitcode linker for NVPTX too, then makes it shared code rather than AMDGPU code.

## Changes

- **`build.rs` compiles the bitcode linker whenever `amdgpu` or `nvptx` is on** (first commit, the fix on its own). `lld.cpp`, `printf.cpp` and the `lldELF`/`lldCommon` link lines stay with `amdgpu`.
- **The bitcode linker moves to `src/shared/cpp_shims/bitcode.cpp`, unchanged, behind one Rust wrapper** (second commit). `shared::bitcode::link_bitcode` declares the two shim functions and turns the returned message into a `Result`; `amdgpu/device_libs.rs` and `nvptx/libdevice.rs` each drop their own copy. It can be dropped without touching the fix.

## Testing

- Before, on main: `cargo test -p cubecl-cuda --no-run` fails to link the lib and `graph` test binaries with the two undefined symbols above.
- After, at each commit: both link. `cargo clippy -p cubecl-llvm --all-targets` is clean with no backend, `nvptx`, `amdgpu`, and both; the `amdgpu` device library tests pass.
- `cargo test -p cubecl-cuda` with only CUDA, on an RTX 2060: 624 tests passed, including the f16 and f32 `sin`, `cos`, `tanh`, `exp`, `sinh`, `cosh` and `expm1` tests that go through libdevice. Then the sixteen running tests (among them `test_all_reduce_sync_collective`, the cmma tests and `stream_errors`) stopped together and nothing moved for 13 minutes, so the run was killed. Main cannot run this suite without ROCm to compare against, so whether the hang predates this PR is not known.


marcantoinem (approved):
LGTM

PR #1671, opened 2026-09-18
fix(std): keep a reshape out of the perpendicular copy

### LINKED ISSUE #1670: cpu: copy_into sends a reshape to a kernel that assumes the shapes match
**Crate:** `cubecl-std` (`tensor::contiguous`), on `cubecl-cpu`
**Rev:** cubecl `fc902af`, via burn `0.22.0-pre.3` / `burn-cpu`. `main` has the same code.

## Summary

`copy_perpendicular` derives the input offset from the output's coordinates:

```rust
// perpendicular.rs
for axis in 0..input.rank() {
    let coordinate = output.coordinate(offset_output * vector_size, axis);
    batch_offset += coordinate * input.stride(axis);
}
```

That holds only when the two shapes are equal, which is the `into_contiguous` case it was written for. `copy_into` guards on the rank instead:

```rust
// launch.rs:35
let same_rank = input.strides.len() == output.strides.len();
if input.strides[rank - 1] != 1 && is_cpu && same_rank {
    launch_copy_perpendicular_ref(client, input, output, dtype);
} else {
    copy_gpu_ref(client, input, output, dtype);
};
```

So a reshape that keeps the rank and changes the shape reaches a kernel that cannot express it, and reads out of bounds. For input `[3,2,1]` strides `(1,3,6)` into output `[1,3,2]`, output index 3 decomposes to `(0, 1, 1)` against the output shape and lands at `0*1 + 1*3 + 1*6 = 9` in a six element buffer.

`is_cpu` is why no GPU backend sees this.

## Reproduce

Through burn, on `burn-cpu`:

```rust
// no unit axis
let t = Tensor::<B, 1, Int>::arange(0..12, &device).float().reshape([2, 3, 2]);
t.permute([2, 1, 0])                   // [2,3,2], strides (1,2,6)
 .reshape([3, 2, 2]);
// expected [[[0,6],[2,8]], [[4,10],[1,7]], [[3,9],[5,11]]]
// got      positions 4..8 hold 1, 7, 3, 9, 2

// trailing unit axis
let t = Tensor::<B, 3>::from([[[1., 3., 5.], [2., 4., 6.]]]);   // [1,2,3]
t.permute([2, 1, 0])                   // [3,2,1], strides (1,3,6)
 .reshape([1, 3, 2]);
// expected [[[1,2],[3,4],[5,6]]], positions 1 to 5 all wrong
```

Both pass on flex and ndarray. Reshapes that change the rank are correct, since `same_rank` sends them to `copy_gpu_ref`.

## Fix

Guard on the shape rather than the rank:

```rust
let same_shape = input.shape == output.shape;
```

## Impact

`copy_into` and `into_contiguous` are what every op calls to normalise a layout, so on the cpu runtime any permute followed by a same-rank reshape can return wrong data.

burn's tensor suite on `burn-cpu` had 7 einsum tests returning wrong values; einsum reaches matmul through `permute -> reshape -> matmul -> reshape -> permute`. Five were a burn-side contiguity check that did not ignore unit axes and sent contiguous tensors down the copy path; fixing that took the suite from 1478 passed / 16 failed to 1483 passed / 11 failed. `einsum_contraction_places_right_output_before_left` and `einsum_noncontiguous_operands` still fail, and the two repros above are reduced from them.

Liberxue:
`copy_perpendicular` reads the input at the output's coordinates:

```rust
let coordinate = output.coordinate(offset_output * vector_size, axis);
batch_offset += coordinate * input.stride(axis);
```

That only expresses a change of layout, which is the `into_contiguous` case it was written for. `copy_into` guarded on the rank, so a reshape keeping the rank reached it and indexed the input out of bounds. For input `[3,2,1]` strides `(1,3,6)` into output `[1,3,2]`, output index 3 decomposes to `(0,1,1)` against the output shape and lands at `9` in a six element buffer.

Only the cpu runtime takes this path.

## Tests

`test_into_contiguous_shape_mismatch` is the sibling of the existing `test_rank_mismatch`: same rank, different shape.

| | `test_shape_mismatch` | other 7 |
|---|---|---|
| without | FAILED | ok |
| with | ok | ok |

`test_permuted_sweep` passes either way, so narrowing the guard does not lose a layout the kernel handled correctly.

Verified through burn on `burn-cpu`: the seven einsum tests that returned wrong values all pass, taking the suite from 1483 passed / 11 failed to 1485 / 9.

Fixes #1670


wingertge (commented):
I didn't write this one so gotta leave the final assessment to @nathanielsimard without fully understanding the kernel logic, looks safe at least though

PR #1678, opened 2026-09-21
fix(throughput): release the probe pool lock before the device call

### LINKED ISSUE #1668: Autotune deadlocks: `PooledProbes::cleanup_unless_held` holds a global lock across a blocking device call
**Crate:** `cubecl-std` (`throughput::pooling`), reached through `cubecl-cpu`
**Rev:** cubecl `707610b`, via burn `0.22.0-pre.3` / `burn-cpu`. `main` has the same code.
**Environment:** Linux x86_64, 36 cores

## Summary

`cubecl-std/src/throughput/pooling.rs:35` takes the global `POOLED_PROBES` mutex and then makes a blocking device call while still holding it:

```rust
static POOLED_PROBES: Mutex<Option<HashMap<ServiceId, usize>>> = Mutex::new(None);

pub(super) fn cleanup_unless_held(client: &Client) {
    let pooled = POOLED_PROBES.lock();

    if !Self::held_by(&pooled, client.service_id()) {
        client.memory_cleanup();   // blocks, with the lock still held
    }
}
```

Every autotune bounds generation reaches this through `roofline_bounds`. With several threads autotuning at once, one thread blocks inside `memory_cleanup()` while holding the lock and never returns, so every other autotuning thread piles up behind it. The process does not recover and spins at about 85% CPU.

## Stacks

The lock holder, blocked on its own device submit:

```
#9 autotune_bounds::with_bounds<InterpolateAutotuneKey, ...>
#8 cubecl_std::throughput::base::roofline_bounds
#7 cubecl_std::throughput::base::measure_launch_overhead
#6 PooledProbes::cleanup_unless_held      <- holds POOLED_PROBES
#5 Client::memory_cleanup
#4 ChannelDeviceHandle::submit_inner
#3 std::thread::sleep
```

Five more threads are blocked on that same address, arriving from four autotune keys (Matmul, Interpolate, ReduceDim) and both `roofline_bounds` entry points:

```
#1 <std::sys::sync::mutex::futex::Mutex>::lock_contended
#2 <cubecl_std::throughput::pooling::PooledProbes>::cleanup_unless_held
#3 cubecl_std::throughput::base::compute_bound        (or measure_launch_overhead)
#4 cubecl_std::throughput::base::roofline_bounds
```

All 75 threads at that point:

| Count | State |
|---|---|
| 37 | `DispatcherWorker` parked on `mpmc::recv`, idle |
| 20 | `submit_inner` -> `thread::sleep`, retry spin |
| 6 | `submit_blocking` -> oneshot recv |
| 5 | `lock_contended` on `PooledProbes` |
| 5 | oneshot recv waiting for an autotune `BatchOutcome` |
| 1 | the lock holder above |

One dispatcher thread (`DSD-0-0`) is itself running autotune and blocked on the same mutex, which is why the 37 idle workers never receive the task that would let `memory_cleanup()` return.

## Reproduce

```
BURN_DEVICE=cpu cargo test --release -p burn-backend-tests \
  --no-default-features --features cpu,std --test tensor
```

Wedges after about 3.5 minutes with 313 of 1466 tests done, and stays wedged (observed for 13 hours before being killed). A second run wedged at 312.

Two data points that isolate it to concurrent autotuning:

- `--test-threads=1` runs the same 1466 tests in 68 seconds.
- `CUBECL_AUTOTUNE_LEVEL=minimal` does not help; it wedges in the same place.

## Suggested fix

Decide under the lock, release it, then clean up:

```rust
pub(super) fn cleanup_unless_held(client: &Client) {
    let held = {
        let pooled = POOLED_PROBES.lock();
        Self::held_by(&pooled, client.service_id())
    };

    if !held {
        client.memory_cleanup();
    }
}
```

This widens the race the existing comment warns about, so the count may need to be atomic or the cleanup may need to re-check after acquiring. Either way the blocking call should not run under the global lock.

## Related

- #1566 changes the client wait on the CPU runtime from spinning to parking on the completion counter, which matches the 20 spinning threads above. It does not touch `pooling.rs`, so the lock inversion would remain.
- #1663 defers autotune, which would change how often several threads reach `roofline_bounds` at once.

ThierryCantin-Demers:
Closes #1668.

`cleanup_unless_held` held the global `POOLED_PROBES` mutex across `client.memory_cleanup()`, which spins until the device runner thread drains its queue. That thread takes the same mutex when it autotunes, so the two wait on each other and never recover.

- The decision stays under the lock, the release is issued after it is dropped. A sweep entering in between pays one fault for pools it had not measured against.
- The cleanup is skipped when the cache answered and no probe ran. `roofline_bounds` asked for three per call on every autotuning thread, which is the traffic that fills the queue.

burn's CPU tensor suite at default parallelism, 16 thread desktop:

| cubecl | test threads | result |
| --- | --- | --- |
| merge base | 16 | wedged at 321 of 1494 tests, killed after 9 minutes of silence |
| merge base | 1 | 1481 passed, 13 failed, 48.05s |
| this branch | 16 | 1481 passed, 13 failed, 58.84s |

The 13 are pre-existing burn CPU failures (einsum, mask broadcast, reduce dims, aggregation) that the parallel suite never reached before. The added test takes the lock from a second thread during the release: against the old code it fails in 5s instead of hanging.

No API change, so no cubek or burn companion PR.


ThierryCantin-Demers:
Hi @Liberxue! This should fix the issue for the deadlock, if you want to test it fixes it properly on your side too it would be great, otherwise it's fine too and I will merge it as is.

Liberxue:
@ThierryCantin-Demers Thanks for picking this up. The change breaks the cycle in the issue: `held_by` is decided under the lock and `memory_cleanup()` is issued after it is dropped, so the dispatcher thread that was blocked on `POOLED_PROBES` can take it, finish its autotune and drain the queue, which is what lets the release return. `enter_service` and `Drop` only touch the map, so that was the only place the lock crossed a device call.

Verified on the 36 core box from the issue. Same burn commit (`faec3983`) and same cubecl rev (`4675224`, what burn main pins) in both rows, patched in over a path override so the only difference is this PR's 87 lines. Default test parallelism, `BURN_DEVICE=cpu cargo test --release -p burn-backend-tests --no-default-features --features cpu,std --test tensor`:

| cubecl | result |
| --- | --- |
| `4675224` | wedged at 371 of 1510 tests, killed at 25 minutes |
| `4675224` + this PR | 1493 passed, 11 failed, finished in 92.53s |

Same shape as your table, one row further out: 371 of 1510 at 36 threads against your 321 of 1494 at 16. The wedged run goes quiet right after a batch of conv1d/conv2d tests report `has been running for over 60 seconds` and prints nothing for the remaining 24 minutes.

The 11 that remain are pre-existing and unrelated to this: nine einsum, which are #1670 here and tracel-ai/burn#5738, and two zero-sized tensor cases under tracel-ai/burn#5310. The `mask_where`/`mask_fill` broadcast failures your table counted are gone, fixed by tracel-ai/burn#5767.

Two things I would like your read on.

**The released pool can still be measured against.** The comment this replaces said the decision had to be made under the lock, so a sweep entering between the read and the release would not lose the pool it is about to measure. That race is now reachable, and the cost looks larger than one fault: `PooledProbes` exists because faulting the next probe back in costs more than the measurement, so the entering sweep measures a cold pool and `client.measure_throughput` caches what it measured. A pessimistic peak then sits in the device's throughput store and feeds `roofline_bounds` from there on, with nothing to mark it apart from a real one. Narrow window, one measurement, but silent. Is a release-pending marker kept under the lock, or a per-service lock instead of the global one, worth the extra state?

**`measure_memory_curve` still releases unconditionally.** The `probed` gate went into `measure_peak_throughput` only, so `base.rs:47` issues the blocking release even when every point of the sweep came from the cache. It is off the `roofline_bounds` path, so it is not the traffic you measured, but it is the same call in the same shape.

ThierryCantin-Demers:
@Liberxue Thanks for running it.

I do not think the race reaches the cache: allocation and first touch happen before anything is timed, since a probe does `reserve`, then `prime` writing every line, then `verify` as one untimed pass, and only then keeps the fastest of many samples. A low peak would loosen `time_limit` anyway, not tighten it.

You are pointing at something real next to it, though. `Schedule::run_batch` wraps the round robin in `client.exclusive` so unrelated work cannot land in the middle of a measurement, and the peak probes take nothing of the sort. So candidates are timed with the device to themselves and the ceiling they are judged against is not. That deserves its own issue, and it covers the pool case: a sweep that can lose its pool was sharing the device regardless. I will do it as a follow-up though as this PR does already fix what it was supposed to and is verified.

A per-service lock would not break the cycle, being inside one device. `exclusive` is also how the runner thread ends up in autotune in your stack, so it gets there on every tune batch rather than by chance.

`measure_memory_curve` you are right about, and I have fixed it.

